PR #14, opened 2026-02-11
feat(importer): add traffic source importers for Burp Suite, HAR, and mitmproxy

maurodelossantos-praetorian:
## Summary

Implements LAB-111: Traffic Source Importers for Vespasian.

Adds three importers that convert traffic captures from popular security tools into Vespasian's `ObservedRequest` format:

- **BurpImporter**: Parses Burp Suite XML exports with automatic base64 detection and HTTP message parsing
- **HARImporter**: Parses HAR 1.2 JSON format (browser DevTools exports) with header array conversion
- **MitmproxyImporter**: Parses mitmproxy JSON flow exports with URL construction from components

All importers implement the existing `TrafficImporter` interface and set the `Source` field to `"import:<format>"`.

### Core Functionality
- Parse request method, URL, headers, and body
- Parse response status code, headers, Content-Type, and body
- Extract query parameters from URLs
- Handle base64-encoded bodies (Burp and mitmproxy formats)
- Support both single entries and arrays (mitmproxy)

### Importer Registry
- `Get(format string)` - Retrieve importer by name
- `SupportedFormats()` - Returns sorted list of available formats
- Unexported registry prevents external mutation

### Scope Filtering
- `ImportOptions.Scope` for domain-based filtering
- Exact match: `example.com`
- Wildcard match: `*.example.com` (matches subdomains AND root domain)

### Security Protections
- **Resource exhaustion prevention**: All importers use `io.LimitReader` (500MB max)
- **XXE attack prevention**: Burp importer rejects `<\!DOCTYPE>` and `<\!ENTITY>` declarations
- **Body truncation prevention**: Uses `bytes.Index` instead of `bufio.Scanner` (no 64KB line limit)
- **Nil body handling**: Empty bodies return `nil`, not empty byte slices

## Test Coverage

- **61 test cases** covering all importers, helpers, and edge cases
- **90.9% code coverage**
- Tests include: basic functionality, base64 encoding/decoding, multiple entries, error handling, scope filtering, security limits, query parameter extraction

## Files Changed

| File | Purpose |
|------|---------|
| `pkg/importer/burp.go` | Burp Suite XML importer with XXE protection |
| `pkg/importer/har.go` | HAR 1.2 JSON importer |
| `pkg/importer/mitmproxy.go` | mitmproxy JSON importer |
| `pkg/importer/options.go` | Scope filtering with `ImportOptions` |
| `pkg/importer/registry.go` | Format registry with `Get()` and `SupportedFormats()` |
| `pkg/importer/helpers.go` | Shared utilities (`extractQueryParams`) |
| `pkg/importer/*_test.go` | Comprehensive test coverage |

## Review Feedback Addressed

- [x] C1: Replace `bufio.Scanner` with `bytes.Index` (64KB line limit)
- [x] H1: Add XXE protection (DOCTYPE/ENTITY rejection + size limit)
- [x] H2: Add `io.LimitReader` to mitmproxy importer
- [x] H3: Fix non-nil empty body handling in HAR
- [x] H4: Make registry unexported
- [x] M2: Use `url.URL` struct for URL construction
- [x] L1: Sort `SupportedFormats()` output
- [x] L2: Use `strings.EqualFold` for base64 attribute check
- [x] Add `io.LimitReader` to HAR importer (consistency)

Closes LAB-111


santiago-praetorian (changes requested):
## PR Review: Traffic Source Importers for Burp Suite, HAR, and mitmproxy

### Executive Summary

Well-structured PR implementing three traffic source importers with scope filtering, a registry, and query parameter extraction. Build passes, all 36 tests pass, race detection is clean, and coverage is 92.6%. The code is clean and idiomatic Go overall. However, there are several issues that should be addressed before merge — most notably a silent data truncation bug and some memory safety concerns.

### Status Table

| Check | Status |
|-------|--------|
| Build | **PASS** |
| Tests | **PASS** (36 tests) |
| Race Detection | **PASS** |
| Code Coverage | **PASS** (92.6%) |
| Code Formatting (gofmt) | **PASS** |
| Dev Artifacts | **CLEAN** |
| PR Description Match | **MATCH** |
| Ticket Alignment (LAB-111) | **ALIGNED** |

All LAB-111 acceptance criteria are met. 36 tests exceed the ≥15 requirement. Minor note: ticket says `pkg/import/` but impl uses `pkg/importer/` — correct since `import` is a Go keyword.

---

### Findings

#### Critical

**C1. `bufio.Scanner` 64KB line limit causes silent body truncation** — `burp.go:131-220`

`parseHTTPRequest` and `parseHTTPResponse` use `bufio.NewScanner` with the default 64KB max line size. Lines exceeding this (large `Cookie` headers, JWT tokens, single-line JSON bodies, minified JS responses) cause `scanner.Scan()` to return `false` with `scanner.Err()` returning `bufio.ErrTooLong`. **This error is never checked.** The body-reading loop silently stops, causing data loss. For a security tool parsing real-world traffic, this is particularly dangerous.

**Fix:** Either increase the buffer with `scanner.Buffer(make([]byte, 0), 10*1024*1024)` and check `scanner.Err()`, or replace the scanner-based body reading with a direct `bytes.Index` approach (the data is already in memory).

#### High

**H1. XML entity expansion / billion laughs attack** — `burp.go:58`

Go's `encoding/xml` does not resolve external entities (no classic XXE), but it **does** process internal entity definitions. A crafted Burp XML with recursive entity expansion can cause OOM from a small file. Legitimate Burp exports never use DTDs.

**Fix:** Add `io.LimitReader(r, maxSize)` and/or pre-scan for `<!DOCTYPE`/`<!ENTITY` declarations and reject them.

**H2. Unbounded `io.ReadAll` in mitmproxy importer** — `mitmproxy.go:59`

`io.ReadAll(r)` reads the entire input with no size limit, then `json.Unmarshal` creates a second copy in memory. A multi-GB file causes OOM.

**Fix:** Use `io.LimitReader` or switch to streaming `json.NewDecoder`. At minimum, detect array vs object by peeking at the first byte instead of try/fail unmarshal.

**H3. HAR importer produces non-nil empty `Body` for GET requests** — `har.go:82,87`

`[]byte(entry.Request.PostData.Text)` when `Text` is `""` produces a non-nil zero-length slice. The Burp and mitmproxy importers correctly return `nil` for empty bodies. Downstream `if req.Body != nil` checks will get false positives. Same issue on line 87 for response content.

**Fix:** Guard with `if text != "" { body = []byte(text) }`.

**H4. Exported mutable global `Registry` map** — `registry.go:20`

`var Registry = map[string]TrafficImporter{...}` is exported and mutable. Concurrent reads/writes would cause a data race and panic. The error message on line 31 hardcodes supported formats instead of using the map dynamically.

**Fix:** Make unexported, or protect with `sync.RWMutex`. Use `SupportedFormats()` in the error message.

#### Medium

**M1. Scope filter bypass via URL userinfo** — `options.go:36-57`

`url.Parse("https://evil.com@example.com/path")` treats `evil.com` as userinfo and `example.com` as host. The scope check passes for `example.com` but the URL actually points at `evil.com` in many HTTP clients. If generated specs are later used for probing, this is an SSRF vector.

**Fix:** After parsing, reject URLs with non-nil `parsed.User` and validate scheme is `http`/`https`.

**M2. URL construction from untrusted components** — `mitmproxy.go:121-127`

`constructURL` builds URLs via string concatenation with no validation. A `host` containing `@`, `/`, or `#` produces URLs that parse differently than intended. `scheme` could be `javascript`/`data`/`file`. Port 0 or negative values produce invalid URLs.

**Fix:** Validate components or use `url.URL` struct for proper construction.

**M3. HAR `Content.Text` may be base64-encoded but is never decoded** — `har.go:87`

The HAR 1.2 spec defines a `content.encoding` field (e.g., `"base64"`). The code doesn't parse this, so base64-encoded response bodies (common in Chrome DevTools exports) are stored as raw base64 text.

**Fix:** Parse the `encoding` field and decode accordingly.

**M4. `parseHTTPRequest`/`parseHTTPResponse` code duplication** — `burp.go:131-220`

These two functions share ~80% of their logic. Only the first-line parsing differs.

**Fix:** Extract shared header+body parsing into a helper.

**M5. Header deduplication loses security-relevant data** — all importers

`map[string]string` for headers means `Set-Cookie`, duplicate `Authorization`, and other multi-value headers lose all but the last value. This is an architectural constraint from `crawl.ObservedRequest`.

**Fix:** Document as known limitation. Consider future change to `map[string][]string`.

#### Low

**L1. `SupportedFormats()` returns non-deterministic order** — `registry.go:37-43` — Add `sort.Strings(formats)`.

**L2. Base64 attribute check is case-sensitive** — `burp.go:121` — `d.Base64 == "true"` won't match `"TRUE"`. Use `strings.EqualFold`.

**L3. `constructURL` doesn't handle port 0 or empty scheme** — `mitmproxy.go:121-127`.

**L4. No adversarial test coverage** — No tests for entity expansion, long lines, malformed URLs, deeply nested JSON, or negative ports. Consider Go fuzz tests.

---

### Recommendation

**Must fix before merge:** C1, H1, H2, H3
**Should fix (quick wins):** H4, M2, L1, L2
**Follow-up ticket:** M1, M3, M5, L4

Overall solid implementation — once the critical and high issues are addressed, this is ready to merge.

santiago-praetorian (changes requested):
# PR Review: Traffic Source Importers (Burp Suite, HAR, mitmproxy)

## Summary

Well-structured implementation of three traffic importers that correctly convert Burp Suite XML, HAR JSON, and mitmproxy JSON into Vespasian's `ObservedRequest` format. All tests pass, 90.9% coverage, no race conditions, gofmt clean, go vet clean. A few notable gaps in test coverage for security-critical code paths and one defense-in-depth limitation.

## Status Table

| Check | Status |
|-------|--------|
| Build | PASS |
| Tests | PASS (21 test functions, ~52 subtests) |
| Race Detection | PASS |
| Code Formatting | PASS |
| go vet | PASS |
| Dev Artifacts | CLEAN |
| PR Description Match | MATCH |
| Ticket Alignment | ALIGNED |
| Coverage | 90.9% |

---

## Security Review

### S1: XXE Pre-scan Only Checks First 8KB [Medium]

**File:** `pkg/importer/burp.go:70-81`

The DOCTYPE/ENTITY scan only reads the first 8,192 bytes. A crafted XML file could place a DOCTYPE declaration after the 8KB boundary, bypassing this check. In practice, Go's `encoding/xml` does **not** process DTDs or expand external entities, so the real-world risk is low — this is defense-in-depth. However, internal entity expansion for denial-of-service could theoretically be a concern.

**Recommendation:** Consider scanning the full buffered content (after `io.ReadAll` from LimitReader) rather than just a peek, or document this as a known limitation.

### S2: No Test Coverage for XXE Rejection [Medium]

**File:** `pkg/importer/burp_test.go`

The DOCTYPE/ENTITY rejection logic (`burp.go:74-76`) has **zero test coverage**. This security-critical code path was added in commit `7c64b72` specifically to address a review finding (H1), but no test verifies it works. The 90% coverage on `BurpImporter.Import` is exactly this gap.

**Recommendation:** Add test cases like:
```go
{"DOCTYPE rejection", `<?xml version="1.0"?><!DOCTYPE foo [<!ENTITY xxe "test">]><items></items>`, true},
{"ENTITY rejection", `<?xml version="1.0"?><!ENTITY xxe SYSTEM "file:///etc/passwd"><items></items>`, true},
```

### S3: 500MB LimitReader with io.ReadAll in Mitmproxy [Low]

**File:** `pkg/importer/mitmproxy.go:62-63`

The mitmproxy importer calls `io.ReadAll(limitedReader)` which can allocate up to 500MB of memory. The Burp and HAR importers use streaming decoders (`xml.NewDecoder`, `json.NewDecoder`), which is better. The mitmproxy importer reads everything because it needs to try both single-flow and array parsing, but this still means a single import call could consume 500MB.

**Recommendation:** Consider a lower limit for mitmproxy specifically, or use `json.NewDecoder` with a token-based approach to detect array vs object first.

### S4: Case-Sensitive DOCTYPE/ENTITY Check [Low]

**File:** `pkg/importer/burp.go:74`

The check uses exact byte matching: `<!DOCTYPE` and `<!ENTITY`. XML is case-sensitive for these declarations so this is technically correct, but an obfuscated file with mixed case like `<!doctype` would not be caught. Go's XML decoder may or may not process such variants.

**Recommendation:** Use `bytes.Contains(bytes.ToUpper(buf), []byte("<!DOCTYPE"))` for more robust detection, or accept this as acceptable given Go's XML parser behavior.

---

## Backend Review

### B1: Code Duplication in HTTP Request/Response Parsing [Medium]

**File:** `pkg/importer/burp.go:161-267`

`parseHTTPRequest` and `parseHTTPResponse` share ~80% identical logic:
- Same separator detection (`\r\n\r\n` with `\n\n` fallback)
- Same header/body split
- Same header parsing loop

Only the first-line parsing differs (request line vs status line).

**Recommendation:** Extract a shared `splitHTTPMessage` function that returns `(firstLine string, headers map[string]string, body []byte, err error)`, then have each function parse just its first line.

### B2: Header Map Loses Multi-Value Headers [Low]

**File:** All importers

All three importers store headers as `map[string]string`, meaning duplicate headers (e.g., `Set-Cookie`, `Link`) only keep the last value. This is a limitation inherited from `crawl.ObservedRequest.Headers` type, not something this PR should fix, but worth documenting.

### B3: `matchesScope` Doesn't Actually Validate URLs [Nit]

**File:** `pkg/importer/options.go:43`

`url.Parse` in Go is extremely lenient — `url.Parse("not-a-url")` succeeds. The "invalid url" test case in `options_test.go:82` passes coincidentally (because `Hostname()` returns `""` which doesn't match `"example.com"`), not because URL validation caught it.

### B4: `TestHARImporter_SizeLimit` Doesn't Actually Test the Limit [Nit]

**File:** `pkg/importer/har_test.go:314-352`

This test function just imports a normal-sized HAR and logs a message about size limits. It doesn't actually verify that an oversized input is truncated or rejected. Consider testing with a reader that exceeds `maxImportSize` or at least verifying the `io.LimitReader` wrapping is in place.

### B5: Test Count Discrepancy [Nit]

PR description claims "61 test cases" but actual count is ~52 subtests across 21 test functions. Not a blocker, but the PR body should be accurate.

---

## Action Items

| ID | Severity | Action |
|----|----------|--------|
| S2 | Medium | Add tests for DOCTYPE/ENTITY rejection paths |
| S1 | Medium | Scan full content for XXE markers instead of first 8KB, or document limitation |
| B1 | Medium | Extract shared `splitHTTPMessage` helper to reduce duplication |
| S3 | Low | Consider lower memory limit or streaming approach for mitmproxy |
| S4 | Low | Consider case-insensitive DOCTYPE/ENTITY matching |
| B4 | Nit | Make `TestHARImporter_SizeLimit` actually test the size limit |
| B3 | Nit | Document `url.Parse` leniency or add explicit hostname validation |
| B5 | Nit | Fix test count in PR description |

santiago-praetorian:
# PR #14 Review: Traffic Source Importers (Burp Suite, HAR, mitmproxy)

## Summary

Solid implementation of three traffic importers that fulfill the LAB-111 acceptance criteria. Build passes, all 61 test cases pass, 90% coverage, no race conditions, no formatting issues. The code addresses multiple rounds of security feedback. A few medium-severity items around memory efficiency and test consistency remain.

## Status Table

| Check | Status |
|-------|--------|
| Build | **PASS** |
| Tests | **PASS** (22 functions, 61+ subtests) |
| Race Detection | **PASS** |
| Code Coverage | **90.0%** |
| Code Formatting (gofmt) | **PASS** |
| Go Vet | **PASS** |
| Dev Artifacts | **CLEAN** |
| PR Description Match | **MATCH** |
| Ticket Alignment | **ALIGNED** |

## Ticket Alignment (LAB-111)

| Acceptance Criteria | Status |
|---|---|
| `TrafficImporter` interface with `Name()` and `Import()` | ✅ |
| `BurpImporter` parses XML with base64 | ✅ |
| `HARImporter` parses HAR 1.2 JSON | ✅ |
| `MitmproxyImporter` parses JSON flows | ✅ |
| `Source` field set to `"import:<format>"` | ✅ |
| Scope filtering | ✅ |
| Imported files match crawled format | ✅ |
| ≥15 test cases (5 per format) | ✅ (61+ cases) |

Note: Ticket specified `pkg/import/` but PR uses `pkg/importer/` — correct decision since `import` is a Go keyword.

---

## Backend Review

### Medium

**B1: `bytes.ToUpper` doubles memory for XXE check** — `burp.go:72`
```go
upperData := bytes.ToUpper(data)
```
After reading up to 500MB with `io.ReadAll`, this allocates another 500MB uppercase copy (~1GB peak). Consider a case-insensitive scan without the full copy, e.g. search for both `<!DOCTYPE` and `<!doctype` (plus a few mixed-case variants). This is sufficient for defense-in-depth since Go's `encoding/xml` already ignores DTDs.

**B2: Streaming mitmproxy still double-buffers** — `mitmproxy.go:78-117`

The streaming decoder accumulates all flows into `[]mitmproxyFlow` (lines 78–108), then iterates again to produce `[]ObservedRequest` (lines 110–117). These could be combined into a single loop to avoid holding both slices simultaneously:
```go
// Parse and convert in one pass
for decoder.More() {
    var flow mitmproxyFlow
    if err := decoder.Decode(&flow); err != nil { ... }
    req, err := i.parseFlow(flow)
    if err != nil { ... }
    requests = append(requests, req)
}
```

### Low

**B3: Inconsistent test assertion style**

`mitmproxy_test.go` and `helpers_test.go` use `testify/assert` + `require`, while `burp_test.go`, `har_test.go`, `options_test.go`, and `registry_test.go` use raw `t.Errorf`/`t.Fatalf`. Consider migrating all to testify for consistency.

**B4: `TestHARImporter_SizeLimit` is effectively a no-op** — `har_test.go:314-341`

This test parses a normal-sized HAR and logs a message. It doesn't actually verify the size limit behavior. At minimum, test that truncated JSON (simulating `LimitReader` cutting it off) produces a decode error:
```go
truncated := normalHAR[:len(normalHAR)/2]
_, err = h.Import(strings.NewReader(truncated))
assert.Error(t, err)
```

### Nit

**B5: `parseHTTPResponse` coverage at 70%** — `burp.go:213`

No test for a response with a malformed status line (e.g., `"HTTP/1.1"` with no status code). Adding one would cover the `len(parts) < 2` branch.

---

## Security Review

### Medium

**S1: No user-friendly error for oversized files** — all importers

`io.LimitReader` returns EOF at the limit, causing the decoder to report "failed to decode" rather than "file exceeds 500MB limit". Users get a confusing error. Consider checking if the limit was hit after a decode error and returning a clearer message.

**S2: XXE pre-scan may false-positive on DOCTYPE in CDATA/body** — `burp.go:71-75`

The pre-scan checks raw bytes for `<!DOCTYPE` / `<!ENTITY`, which could false-positive if these strings appear inside a CDATA section or request body content. In practice this is extremely unlikely in legitimate Burp exports and is acceptable defense-in-depth. Worth documenting this known trade-off.

### Low

**S3: No maximum entry count** — all importers

File size is capped at 500MB, but there's no limit on entry count. A file with millions of tiny entries could exhaust memory from the accumulated `[]ObservedRequest` slice. Consider an optional `MaxEntries` in `ImportOptions`.

**S4: Unvalidated HTTP method from Burp request line** — `burp.go:205`

`strings.Fields(firstLine)` extracts the method without validating it's a known HTTP method. Low risk since this feeds into spec generation, not HTTP dispatch.

**S5: mitmproxy content always assumed base64** — `mitmproxy.go:195-205`

`decodeContent` always attempts base64 decode. If mitmproxy's export format changes or users provide raw content, this silently fails. The assumption aligns with current mitmproxy behavior, so this is informational.

### Nit

**S6: Consider `xml.Decoder.Strict = true`** — `burp.go:78`

For additional XML hardening, `Strict = true` on the decoder would reject non-standard XML constructs.

---

## Recommendation: **APPROVE** ✅

The PR correctly implements all acceptance criteria from LAB-111. Build, tests, race detection, formatting, and vet all pass cleanly. Security mitigations (XXE prevention, `io.LimitReader`, streaming decoder) are thoughtfully implemented. The medium-severity items (B1 memory doubling, B2 double-buffer, S1 error messages) are improvements worth considering but don't block merge.

🤖 Generated with [Claude Code](https://claude.ai/code)

santiago-praetorian (dismissed):
## PR Review: Traffic Source Importers (Burp, HAR, mitmproxy)

### Summary

Well-implemented feature with thorough test coverage (94.9%), proper security hardening (XXE prevention, size limits, port validation, streaming decoder), and clean code organization. The implementation correctly fulfills all LAB-111 acceptance criteria. **APPROVE** with minor suggestions.

### Status

| Check | Status |
|-------|--------|
| Build | PASS |
| Tests | PASS (65+ test functions, 137+ subtests) |
| Race Detection | PASS |
| Code Formatting (gofmt) | PASS |
| go vet | PASS |
| Dev Artifacts | CLEAN |
| Coverage | 94.9% |
| PR Description Match | MATCH |
| Ticket Alignment | ALIGNED |

---

### Backend Review

#### Medium

**B1: Dead code in `TestMitmproxyImporter_InvalidPort`** (`mitmproxy_test.go:371-402`)

The test case builds a JSON string with a broken `strings.Replace` chain (lines 371-385) that is immediately overwritten by a correct `fmt.Sprintf` version (lines 388-402). Should remove the dead code and keep only the final construction.

**B2: Missing HTTP method validation in HAR and mitmproxy importers** (`har.go`, `mitmproxy.go`)

The Burp importer validates HTTP methods against `validHTTPMethods` (`burp.go:255-257`), but neither HAR nor mitmproxy importers perform this validation. A malformed export with `"method": "GARBAGE"` would be accepted. For consistency, all three importers should validate the HTTP method, or the validation should be extracted to a shared helper.

#### Low

**B3: `MaxEntries` enforced post-parse** (`options.go:71`) — All entries are parsed into memory before the limit is checked. The 500MB file size limit provides some protection, but for true memory safety the limit could be pushed into the `Import()` methods.

**B4: `PageURL` field never set** — `ObservedRequest.PageURL` is never populated by any importer. Arguably correct for imported traffic, but worth documenting.

#### Nit

**B5:** Package path `pkg/importer/` correctly diverges from ticket's `pkg/import/` (reserved keyword) — just noting for traceability.

**B6:** `TestSupportedFormats` (`registry_test.go:71-98`) could be simplified to `assert.Equal(t, []string{"burp", "har", "mitmproxy"}, SupportedFormats())` since the output is already sorted.

---

### Security Review

#### Medium

**S1: XXE pattern matching has incomplete case coverage** (`burp.go:40-48`)

The hardcoded `xxeDOCTYPEPatterns`/`xxeENTITYPatterns` lists cover only 6 of 128+ possible casings. Unmatched casings like `<!dOcTyPe` would bypass the pre-scan.

**Mitigating factor:** `decoder.Strict = true` (line 114) is the primary defense and rejects DOCTYPE declarations regardless of casing. Go's `encoding/xml` also doesn't support external entity resolution. The pre-scan is defense-in-depth.

**Suggestion:** Replace the hardcoded list with a case-insensitive scan using `bytes.EqualFold`:

```go
func containsCaseInsensitive(data, pattern []byte) bool {
    for i := 0; i <= len(data)-len(pattern); i++ {
        if bytes.EqualFold(data[i:i+len(pattern)], pattern) {
            return true
        }
    }
    return false
}
```

#### Low

**S2:** Burp importer buffers up to 500MB via `io.ReadAll` (`burp.go:97`). Required for XXE pre-scan; consider documenting the memory impact for concurrent usage.

**S3:** No status code range validation (`burp.go:157-159`). Crafted exports could set `<status>99999</status>`. Low impact since it's just stored as an int.

**S4:** No scheme/host validation in mitmproxy `constructURL` (`mitmproxy.go:195-213`). Minimal impact since URLs are stored, not dereferenced.

---

### Suggested Improvements (non-blocking)

1. Clean up dead code in `TestMitmproxyImporter_InvalidPort`
2. Add HTTP method validation to HAR and mitmproxy importers for consistency

WizardML7 (approved):
These additions correctly implement the scope described in LAB-111 fulfilling the requirements of the: TrafficImporter, BurpImporter, HARImporter, and MitmproxyImporter, as well as ensuring that each of them have --scope flag support in the options.go file. 

Good PR that meets all of LAB-111's requirements

PR #23, opened 2026-02-27
fix: preserve partial results when crawl duration timeout expires

logan-bayless:
## Bug

When Katana's `CrawlDuration` timeout fires, `engine.Crawl()` returns `context.DeadlineExceeded` alongside any partial results collected. `doCrawl()` only checked for `context.Canceled` (SIGINT/SIGTERM), so the `DeadlineExceeded` path discarded partial results and returned an error — no output file was written.

## Fix

Extend the graceful shutdown check to also match `context.DeadlineExceeded`:

```go
// Before
if errors.Is(err, context.Canceled) && len(requests) > 0 {

// After
if (errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)) && len(requests) > 0 {
```

Partial results are now written to the output file on timeout, same as on SIGINT.

### DIFF
--- a/cmd/vespasian/main.go
+++ b/cmd/vespasian/main.go
@@ -86,8 +86,8 @@ func doCrawl(ctx context.Context, targetURL string, rawHeaders []string, opts cr
 	crawler := crawl.NewCrawler(opts)
 	requests, err := crawler.Crawl(ctx, targetURL)
 	if err != nil {
-		if errors.Is(err, context.Canceled) && len(requests) > 0 {
-			// Graceful shutdown — return partial results.
+		if (errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)) && len(requests) > 0 {
+			// Graceful shutdown or timeout — return partial results.
 			return requests, nil
 		}
 		return nil, fmt.Errorf("crawl failed: %w", err)
--- a/cmd/vespasian/main_test.go
+++ b/cmd/vespasian/main_test.go
@@ -16,6 +16,8 @@ package main
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


santiago-praetorian (commented):
## Backend Review

### Nit — Test doesn't exercise the actual `doCrawl()` code path

`returnPartialOnGraceful()` is a **copy** of the condition logic from `doCrawl()`, not a call to it. The test proves the helper works, but doesn't prove the actual `doCrawl()` function was patched. If someone later modifies `doCrawl()` without updating the test helper, the tests would still pass while the production code regresses.

The comment in the test acknowledges this tradeoff (`"Because doCrawl creates its own Crawler internally..."`). The test is still valuable as a behavioral specification, but consider:

1. Adding a comment or `//go:generate` directive that ties the helper back to the source line, or
2. Long-term: refactoring `doCrawl` to accept a `Crawler` interface so the crawl can be mocked and tested directly.

### Nit — Error message on `DeadlineExceeded` with no results

When `context.DeadlineExceeded` is returned with zero results, the error becomes `"crawl failed: context deadline exceeded"`. This is accurate but may be confusing to users — they might not realize this means the timeout fired and found nothing. A more descriptive message could help, e.g. `"crawl timed out with no results"`.

Minor since it only affects the zero-results edge case.

---

## Security Review

No security findings. The change only affects internal error handling for context errors already returned by the Katana engine. No new inputs, attack surface, or data exposure risk.

logan-bayless (commented):
Thanks for the review! Addressed both nits in d269eb2:

1. **Test helper traceability** — Added `//go:generate` directive and updated comments with line references tying the helper back to the production code path.

2. **Descriptive timeout error** — `DeadlineExceeded` with zero results now returns `"crawl timed out with no results"` instead of the generic `"crawl failed: context deadline exceeded"`. Test helper synced to match.

santiago-praetorian (approved):
## PR Review: `fix: preserve partial results when crawl duration timeout expires`

### Status

| Check | Status |
|-------|--------|
| Build | ✅ PASS |
| Tests | ✅ PASS (all packages) |
| Race Detection | ✅ PASS |
| Code Formatting | ✅ PASS |
| Dev Artifacts | ✅ CLEAN |

### Backend Review

**Nit — Test helper duplicates production logic**
`returnPartialOnGraceful` in `main_test.go:581-590` is a verbatim copy of the condition in `doCrawl`. If the condition in `main.go:89` drifts later, the test won't catch it. Acceptable trade-off given `doCrawl` can't be unit-tested without a live crawler. If the condition grows, consider extracting into a shared predicate like `isGracefulTermination(err error) bool`.

**Low — Partial results indistinguishable from complete results**
When `DeadlineExceeded` fires, `doCrawl` returns `(requests, nil)` with no signal the crawl was cut short. Consider a stderr log line when partial results are returned due to timeout, even without `--verbose`. Not a blocker.

### Security Review

No blocking security findings.

- **DeadlineExceeded treated like Canceled:** Safe. Both are standard Go context termination signals; Katana closes cleanly in both paths.
- **Partial data corruption risk:** None. Each `ObservedRequest` is fully constructed before append under mutex lock. No partial-write scenario.
- **Error handling correctness:** Correct. `len(requests) > 0` guard prevents empty partial results from silently succeeding.
- **Injection / data integrity:** No new attack surface. Change only affects the error-handling branch.

### Verdict: **APPROVED** ✅

Clean, minimal fix with comprehensive tests covering all six branch combinations.

santiago-praetorian (approved):
## PR Review: `fix: preserve partial results when crawl duration timeout expires`

### Status

| Check | Status |
|-------|--------|
| Build | ✅ PASS |
| Tests | ✅ PASS (all packages) |
| Race Detection | ✅ PASS |
| Code Formatting | ✅ PASS |
| Dev Artifacts | ✅ CLEAN |

### Backend Review

**Nit — Test helper duplicates production logic**
`returnPartialOnGraceful` in `main_test.go:581-590` is a verbatim copy of the condition in `doCrawl`. If the condition in `main.go:89` drifts later, the test won't catch it. Acceptable trade-off given `doCrawl` can't be unit-tested without a live crawler. If the condition grows, consider extracting into a shared predicate like `isGracefulTermination(err error) bool`.

**Low — Partial results indistinguishable from complete results**
When `DeadlineExceeded` fires, `doCrawl` returns `(requests, nil)` with no signal the crawl was cut short. Consider a stderr log line when partial results are returned due to timeout, even without `--verbose`. Not a blocker.

### Security Review

No blocking security findings.

- **DeadlineExceeded treated like Canceled:** Safe. Both are standard Go context termination signals; Katana closes cleanly in both paths.
- **Partial data corruption risk:** None. Each `ObservedRequest` is fully constructed before append under mutex lock. No partial-write scenario.
- **Error handling correctness:** Correct. `len(requests) > 0` guard prevents empty partial results from silently succeeding.
- **Injection / data integrity:** No new attack surface. Change only affects the error-handling branch.

### Verdict: **APPROVED** ✅

Clean, minimal fix with comprehensive tests covering all six branch combinations.

PR #24, opened 2026-02-27
chore: crawl package hygiene — idempotent close, input validation

logan-bayless:
Three small hygiene improvements to `pkg/crawl/`, no behavioral changes for valid inputs.

## Changes

**1. Idempotent engine close** — `engine.Close()` was called twice on the context cancellation path (explicit call + deferred close). Both Katana engines happen to be idempotent today, but this is an undocumented assumption. Wrapped with `sync.Once`.

**2. Negative depth validation** — `--depth=-1` silently produces zero results because Katana skips even the seed URL when `MaxDepth < 0`. Now returns a clear error. (`--depth=0` remains valid — it crawls only the seed page.)

**3. URL validation in library** — `Crawl()` accepted any string, relying on CLI-layer validation. Programmatic callers now get a clear error for empty/invalid URLs instead of opaque Katana failures.

### DIFF
--- a/pkg/crawl/crawler.go
+++ b/pkg/crawl/crawler.go
@@ -66,7 +66,8 @@ func (c *Crawler) Crawl(ctx context.Context, targetURL string) ([]ObservedReques
 		return nil, fmt.Errorf("depth must be non-negative, got %d", c.opts.Depth)
 	}
 
-	if _, err := url.Parse(targetURL); err != nil || targetURL == "" {
+	u, err := url.Parse(targetURL)
+	if err != nil || targetURL == "" || (u.Scheme != "http" && u.Scheme != "https") {
 		return nil, fmt.Errorf("invalid target URL: %q", targetURL)
 	}
 
--- a/pkg/crawl/crawler.go
+++ b/pkg/crawl/crawler.go
@@ -16,6 +16,7 @@ package crawl
 
 import (
 	"context"
+	"fmt"
 	"net/url"
 	"sync"
 	"time"
@@ -59,6 +60,14 @@ func (c *Crawler) Crawl(ctx context.Context, targetURL string) ([]ObservedReques
 	maxPages := c.opts.MaxPages
 	if maxPages <= 0 {
 		maxPages = DefaultMaxPages
+	}
+
+	if c.opts.Depth < 0 {
+		return nil, fmt.Errorf("depth must be non-negative, got %d", c.opts.Depth)
+	}
+
+	if _, err := url.Parse(targetURL); err != nil || targetURL == "" {
+		return nil, fmt.Errorf("invalid target URL: %q", targetURL)
 	}
 
 	// Create a cancellable context to stop Katana when MaxPages is reached.
@@ -125,7 +134,9 @@ func (c *Crawler) Crawl(ctx context.Context, targetURL string) ([]ObservedReques
 	if err != nil {
 		return nil, err
 	}
-	defer func() { _ = engine.Close() }()
+	var closeOnce sync.Once
+	closeEngine := func() { closeOnce.Do(func() { _ = engine.Close() }) }
+	defer closeEngine()
 
 	// Run crawl in goroutine with context cancellation
 	crawlErr := make(chan error, 1)
@@ -142,7 +153,7 @@ func (c *Crawler) Crawl(ctx context.Context, targetURL string) ([]ObservedReques
 		// Context canceled (MaxPages reached or SIGINT/SIGTERM).
 		// Close the engine to unblock the goroutine, then drain it
 		// to avoid a goroutine leak.
-		_ = engine.Close()
+		closeEngine()
 		<-crawlErr
 
 		// If the parent context was canceled (SIGINT/SIGTERM), propagate that error
--- a/pkg/crawl/crawler_test.go
+++ b/pkg/crawl/crawler_test.go
@@ -15,6 +15,8 @@
 package crawl
 
 import (
+	"context"
+	"strings"
 	"testing"
 
 	"github.com/projectdiscovery/katana/pkg/navigation"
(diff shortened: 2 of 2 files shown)


pkg/crawl/crawler.go
```diff
@@ -61,6 +62,14 @@ func (c *Crawler) Crawl(ctx context.Context, targetURL string) ([]ObservedReques
 		maxPages = DefaultMaxPages
 	}
 
+	if c.opts.Depth < 0 {
+		return nil, fmt.Errorf("depth must be non-negative, got %d", c.opts.Depth)
+	}
```
santiago-praetorian, line 67: **[Medium — Backend + Security]** `url.Parse` is notoriously lenient in Go. It successfully parses strings like `"not-a-url"`, `"foo bar"`, `file:///etc/passwd`, and relative paths without error. The only thing this check actually catches is the empty string (handled explicitly).

This means programmatic callers can pass non-HTTP schemes through the library API, which could become an SSRF vector if the underlying engine ever changes behavior.

**Suggestion:** Validate the scheme explicitly to make this check do what the PR description promises:

```go
u, err := url.Parse(targetURL)
if err != nil || targetURL == "" || (u.Scheme != "http" && u.Scheme != "https") {
    return nil, fmt.Errorf("invalid target URL: %q", targetURL)
}
```

Also consider adding a test case for `"not-a-url"` to verify the validation rejects schemeless strings.

santiago-praetorian (commented):
## Code Review

Overall this is a clean, well-scoped PR. The `sync.Once` fix is correct, and the negative depth validation is a good catch. One actionable finding on the URL validation, plus a few minor notes below.

**Summary of findings:**

| Severity | Finding |
|----------|---------|
| Medium | URL validation too permissive — `url.Parse` accepts almost anything (see inline comment) |
| Low | No upper bound on depth (out of scope for this PR, but worth a future hardening ticket) |
| Nit | No direct test coverage for the `sync.Once` double-close path |

logan-bayless (commented):
Thanks for the thorough review! Addressed in 1107e79:

1. **Medium — URL validation**: Tightened to require `http`/`https` scheme explicitly, rejecting schemeless strings, `file://`, etc. Added `TestCrawl_SchemelessURLReturnsError`.

2. **Nit — sync.Once test**: Added `TestCloseEngineOnce` verifying the double-close pattern.

3. **Low — No upper bound on depth**: Acknowledged, out of scope for this PR — good candidate for a future hardening pass.

santiago-praetorian (commented):
## Code Review — `chore: crawl package hygiene`

Clean, well-scoped hygiene PR. The `sync.Once` fix is correct and the tests cover the new validation paths. Two items:

### 1. gofmt formatting (`pkg/crawl/crawler_test.go`) — Nit

`gofmt -l` flags this file. Extra blank line before `TestPageTimeout` (L376) and missing blank line between `TestPageTimeout` and `TestMapResult_TruncatesLargeResponseBody` (L382).

### 2. `url.Parse` is extremely permissive (`pkg/crawl/crawler.go:73`) — Medium

```go
if _, err := url.Parse(targetURL); err != nil || targetURL == "" {
```

Go's `url.Parse` successfully parses almost anything — `"not-a-url"`, `"hello world"`, `"ftp://whatever"` all return `nil` error. The only thing this guard actually catches is `targetURL == ""`. Consider validating for scheme presence:

```go
u, err := url.Parse(targetURL)
if err != nil || targetURL == "" || (u.Scheme != "http" && u.Scheme != "https") {
```

This would also catch non-HTTP schemes (which Katana doesn't support) and truly malformed inputs. The PR description says "clear error for empty/invalid URLs", but in practice only empty strings are caught today.

logan-bayless:
Sorry about that, was working in a local branch with all my current PRs and committed the above changes as a reference -- should be correct now!

santiago-praetorian (changes requested):
## Code Review — `chore: crawl package hygiene`

Clean, well-scoped hygiene PR. The `sync.Once` fix is correct, input validation is solid, and all prior review feedback has been addressed. Two minor items remain before merge.

### Summary

| Check | Status |
|-------|--------|
| Build | ✅ PASS |
| Tests | ✅ PASS |
| Race Detection | ✅ PASS |
| Code Formatting | ❌ NEEDS FIX |
| PR Description Match | ✅ MATCH |
| Prior Review Items | ✅ All addressed |

---

### 1. `gofmt` formatting — Nit (blocking)

`gofmt -l` still flags `pkg/crawl/crawler_test.go`. Two whitespace issues:
- **Line 408**: Extra blank line before `TestPageTimeout`
- **Line 414**: Missing blank line between `TestPageTimeout` and `TestMapResult_TruncatesLargeResponseBody`

Fix: `gofmt -w pkg/crawl/crawler_test.go`

### 2. Add scheme-bypass test cases — Low (non-blocking, nice-to-have)

The scheme validation at `crawler.go:74` correctly rejects non-HTTP schemes (SSRF fix from prior review). The test suite only covers `"not-a-url"`. Consider adding `"file:///etc/passwd"` and `"ftp://example.com"` test cases to document the security intent and prevent regressions.

---

Once the `gofmt` fix is committed, this is good to merge. 🤖

logan-bayless:
Addressed in 6233b65:

1. **gofmt formatting** — Already clean on current HEAD (verified `gofmt -l` returns nothing).
2. **Scheme-bypass test cases** — Expanded `TestCrawl_SchemelessURLReturnsError` into a table-driven `TestCrawl_InvalidSchemeReturnsError` covering `"not-a-url"`, `"file:///etc/passwd"`, and `"ftp://example.com"` to document the SSRF prevention intent.

All tests pass, gofmt clean.

santiago-praetorian (approved):
## PR Review Summary

| Check | Status |
|-------|--------|
| Build | ✅ PASS |
| Tests | ✅ PASS |
| Race Detection | ✅ PASS |
| Code Formatting | ⚠️ Minor gofmt whitespace (likely from merge) |
| Dev Artifacts | ✅ CLEAN |
| PR Description Match | ✅ MATCH |

## Findings

**All three changes are correct and well-scoped:**

1. **Idempotent close** — `sync.Once` wrapper cleanly prevents double `engine.Close()` on the context cancellation + defer path.
2. **Negative depth validation** — Clear error with actual value, correctly placed before engine init.
3. **URL/scheme validation** — Good SSRF prevention restricting to `http`/`https`. Test table covers `file://` and `ftp://` vectors.

**Minor nits (non-blocking):**
- `crawler_test.go` has a small `gofmt` whitespace issue (extra blank line at L419, missing newline before `TestMapResult_TruncatesLargeResponseBody`). Likely from merge with main — can fix in follow-up.
- `TestCloseEngineOnce` tests the `sync.Once` pattern rather than the actual `closeEngine` closure, but the comment documents this intent well.
- Consider adding `data:` or `javascript:` schemes to the SSRF test table in a future pass.

**LGTM** — clean hygiene PR, no behavioral changes for valid inputs.

PR #28, opened 2026-03-06
feat: own browser lifecycle for immediate signal handling

logan-bayless:
## Summary

- **Vespasian now owns the Chrome process** via `go-rod/rod/lib/launcher`, passing the WebSocket URL to Katana's `ChromeWSUrl` option. Katana connects to our browser instead of launching its own.
- **Single `^C` (SIGINT/SIGTERM)**: kills Chrome immediately (stops all outbound requests), drains Katana's internal result buffer for 2s, returns partial results with stderr message: `"interrupt received, returning N partial results"`.
- **Double `^C`**: force-exits via `os.Exit(1)` with stderr message: `"forcing immediate exit"`.
- **Programmatic callers** (agentic workflows): canceling `context.Context` triggers the same graceful shutdown — no terminal-specific semantics leak into the library.

### Why this matters

Previously, `^C` was effectively ignored. Katana's internal context (`context.Background()`) was disconnected from vespasian's signal context, so Chrome kept crawling for up to `CrawlDuration` (default 10min) after the operator requested a stop — a safety issue for a security tool.

### Files changed

| File | Change |
|------|--------|
| `pkg/crawl/browser.go` | New `BrowserManager` type: launches Chrome, exposes `WSURL()`, `Kill()`, `Close()` |
| `pkg/crawl/browser_test.go` | Integration tests (`//go:build integration`): launch/kill, idempotent kill, close cleanup |
| `pkg/crawl/crawler.go` | `Crawl()` creates `BrowserManager` when headless, passes `ChromeWSUrl`, kills Chrome on signal with bounded 2s drain |
| `cmd/vespasian/main.go` | `doCrawl()` prints interrupt message to injectable `io.Writer`; `setupForceExitHandler()` for double-interrupt |
| `cmd/vespasian/main_test.go` | Updated graceful shutdown tests to verify stderr output |
| `go.mod` | `go-rod/rod` promoted from indirect to direct dependency |

## Test plan

- [x] All unit tests pass (`go test ./...`)
- [x] All integration tests pass (`go test -tags integration ./pkg/crawl/ -run TestBrowserManager`)
- [x] `go vet ./...` clean
- [x] Manual: run `vespasian crawl <url>`, hit `^C` — verify Chrome dies, partial results returned
- [x] Manual: run crawl, hit `^C` twice — verify immediate exit
- [ ] CI passes

### DIFF
--- a/cmd/vespasian/main.go
+++ b/cmd/vespasian/main.go
@@ -52,7 +52,8 @@ var CLI struct {
 	Version  VersionCmd  `cmd:"" help:"Show version information"`
 }
 
-// parseHeaders converts "Key: Value" strings to a map, validating for CRLF injection.
+// parseHeaders converts "Key: Value" strings to a map, validating header names
+// against RFC 7230 token production and header values against CRLF injection.
 func parseHeaders(raw []string) (map[string]string, error) {
 	headers := make(map[string]string)
 	for _, h := range raw {
@@ -65,12 +66,46 @@ func parseHeaders(raw []string) (map[string]string, error) {
 		if name == "" {
 			return nil, fmt.Errorf("header has empty name: %q", h)
 		}
-		if strings.ContainsAny(name, "\r\n") || strings.ContainsAny(value, "\r\n") {
-			return nil, fmt.Errorf("header contains invalid CRLF characters: %q", h)
+		if !isValidHeaderName(name) {
+			return nil, fmt.Errorf("header name contains invalid characters (RFC 7230): %q", h)
+		}
+		if strings.ContainsAny(value, "\r\n") {
+			return nil, fmt.Errorf("header value contains invalid CRLF characters: %q", h)
 		}
 		headers[name] = value
 	}
 	return headers, nil
+}
+
+// isValidHeaderName checks that name consists only of RFC 7230 token characters.
+// tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." /
+//
+//	"^" / "_" / "`" / "|" / "~" / DIGIT / ALPHA
+func isValidHeaderName(name string) bool {
+	for i := 0; i < len(name); i++ {
+		if !isTokenChar(name[i]) {
+			return false
+		}
+	}
+	return true
+}
+
+// isTokenChar returns true if c is a valid RFC 7230 tchar.
+func isTokenChar(c byte) bool {
+	switch {
+	case c >= 'A' && c <= 'Z':
+		return true
+	case c >= 'a' && c <= 'z':
+		return true
+	case c >= '0' && c <= '9':
+		return true
+	case c == '!' || c == '#' || c == '$' || c == '%' || c == '&' ||
+		c == '\'' || c == '*' || c == '+' || c == '-' || c == '.' ||
+		c == '^' || c == '_' || c == '`' || c == '|' || c == '~':
+		return true
+	default:
+		return false
+	}
 }
 
 // doCrawl executes the common crawl pipeline: parse headers, create crawler,
@@ -186,6 +221,18 @@ func (c *CrawlCmd) Run() error {
 		Headless: c.Headless,
 	}
 
+	// Create browser before signal setup so force-exit can clean up temp dirs.
+	var browserMgr *crawl.BrowserManager
+	if c.Headless {
+		var err error
+		browserMgr, err = crawl.NewBrowserManager(crawl.BrowserOptions{Headless: true})
+		if err != nil {
+			return fmt.Errorf("launch browser: %w", err)
+		}
+		defer browserMgr.Close()
+		opts.BrowserMgr = browserMgr
+	}
+
 	if c.Verbose {
 		fmt.Fprintf(os.Stderr, "crawling %s (depth=%d, max-pages=%d, timeout=%s)\n",
 			c.URL, opts.Depth, opts.MaxPages, opts.Timeout)
@@ -194,7 +241,11 @@ func (c *CrawlCmd) Run() error {
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 	defer stop()
 
-	setupForceExitHandler(ctx, os.Stderr, nil, os.Exit)
+	setupForceExitHandler(ctx, os.Stderr, func() {
+		if browserMgr != nil {
+			browserMgr.Close()
+		}
+	}, os.Exit)
 
 	requests, err := doCrawl(ctx, os.Stderr, c.URL, c.Header, opts)
 	if err != nil {
@@ -328,6 +379,18 @@ func (c *ScanCmd) Run() error {
 		Headless: c.Headless,
 	}
 
+	// Create browser before signal setup so force-exit can clean up temp dirs.
+	var browserMgr *crawl.BrowserManager
+	if c.Headless {
+		var err error
+		browserMgr, err = crawl.NewBrowserManager(crawl.BrowserOptions{Headless: true})
+		if err != nil {
+			return fmt.Errorf("launch browser: %w", err)
+		}
+		defer browserMgr.Close()
+		opts.BrowserMgr = browserMgr
+	}
+
 	if c.Verbose {
 		fmt.Fprintf(os.Stderr, "crawling %s (depth=%d, max-pages=%d, timeout=%s)\n",
 			c.URL, opts.Depth, opts.MaxPages, opts.Timeout)
@@ -336,7 +399,11 @@ func (c *ScanCmd) Run() error {
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 	defer stop()
 
-	setupForceExitHandler(ctx, os.Stderr, nil, os.Exit)
+	setupForceExitHandler(ctx, os.Stderr, func() {
+		if browserMgr != nil {
+			browserMgr.Close()
+		}
+	}, os.Exit)
 
 	requests, err := doCrawl(ctx, os.Stderr, c.URL, c.Header, opts)
 	if err != nil {
--- a/cmd/vespasian/main.go
+++ b/cmd/vespasian/main.go
@@ -181,23 +181,70 @@ func setupForceExitHandler(ctx context.Context, stderr io.Writer, cleanup func()
 		sigCh := make(chan os.Signal, 1)
 		signal.Notify(sigCh, os.Interrupt, syscall.SIGTERM)
 		<-sigCh
-		if cleanup != nil {
-			cleanup()
-		}
-		fmt.Fprintf(stderr, "forcing immediate exit\n")
-		exitFn(1)
+		onForceExit(stderr, cleanup, exitFn)
 	}()
 }
 
-// onForceExit is the testable core of setupForceExitHandler. It runs cleanup,
-// writes the exit message, and calls exitFn. Extracted so tests can exercise
-// the logic without sending real signals.
+// onForceExit is the testable core of setupForceExitHandler. It runs cleanup
+// (with panic recovery to guarantee exitFn is called), writes the exit message,
+// and calls exitFn. Extracted so tests can exercise the logic without sending
+// real signals.
 func onForceExit(stderr io.Writer, cleanup func(), exitFn func(int)) {
 	if cleanup != nil {
-		cleanup()
+		func() {
+			defer func() { recover() }()
+			cleanup()
+		}()
 	}
 	fmt.Fprintf(stderr, "forcing immediate exit\n")
 	exitFn(1)
+}
+
+// browserSetupResult holds the resources created by setupBrowserAndSignals.
+type browserSetupResult struct {
+	opts    crawl.CrawlerOptions
+	ctx     context.Context
+	stop    context.CancelFunc
+	cleanup func() // caller must defer this
+}
+
+// setupBrowserAndSignals creates a BrowserManager (if headless), wires up
+// signal handling with force-exit support, and returns a cancellable context.
+// The returned cleanup function closes the browser and stops the signal handler;
+// callers must defer it.
+func setupBrowserAndSignals(crawlOpts CrawlOptions, extraOpts crawl.CrawlerOptions) (browserSetupResult, error) {
+	var browserMgr *crawl.BrowserManager
+
+	if crawlOpts.Headless {
+		var err error
+		browserMgr, err = crawl.NewBrowserManager(crawl.BrowserOptions{Headless: true})
+		if err != nil {
+			return browserSetupResult{}, fmt.Errorf("launch browser: %w", err)
+		}
+		extraOpts.BrowserMgr = browserMgr
+	}
+
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+
+	setupForceExitHandler(ctx, os.Stderr, func() {
+		if browserMgr != nil {
+			browserMgr.Close()
+		}
+	}, os.Exit)
+
+	cleanup := func() {
+		stop()
+		if browserMgr != nil {
+			browserMgr.Close()
+		}
+	}
+
+	return browserSetupResult{
+		opts:    extraOpts,
+		ctx:     ctx,
+		stop:    stop,
+		cleanup: cleanup,
+	}, nil
 }
 
 // CrawlCmd crawls a web application to capture HTTP traffic.
@@ -213,41 +260,24 @@ func (c *CrawlCmd) Run() error {
 		return err
 	}
 
-	opts := crawl.CrawlerOptions{
+	bs, err := setupBrowserAndSignals(c.CrawlOptions, crawl.CrawlerOptions{
 		Depth:    c.Depth,
 		MaxPages: c.MaxPages,
 		Timeout:  c.Timeout,
 		Scope:    c.Scope,
 		Headless: c.Headless,
+	})
+	if err != nil {
+		return err
 	}
-
-	// Create browser before signal setup so force-exit can clean up temp dirs.
-	var browserMgr *crawl.BrowserManager
-	if c.Headless {
-		var err error
-		browserMgr, err = crawl.NewBrowserManager(crawl.BrowserOptions{Headless: true})
-		if err != nil {
-			return fmt.Errorf("launch browser: %w", err)
-		}
-		defer browserMgr.Close()
-		opts.BrowserMgr = browserMgr
-	}
+	defer bs.cleanup()
 
 	if c.Verbose {
 		fmt.Fprintf(os.Stderr, "crawling %s (depth=%d, max-pages=%d, timeout=%s)\n",
-			c.URL, opts.Depth, opts.MaxPages, opts.Timeout)
+			c.URL, bs.opts.Depth, bs.opts.MaxPages, bs.opts.Timeout)
 	}
 
-	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
-	defer stop()
-
-	setupForceExitHandler(ctx, os.Stderr, func() {
-		if browserMgr != nil {
-			browserMgr.Close()
-		}
-	}, os.Exit)
-
-	requests, err := doCrawl(ctx, os.Stderr, c.URL, c.Header, opts)
+	requests, err := doCrawl(bs.ctx, os.Stderr, c.URL, c.Header, bs.opts)
 	if err != nil {
 		return err
 	}
@@ -371,41 +401,24 @@ func (c *ScanCmd) Run() error {
 		return err
 	}
 
-	opts := crawl.CrawlerOptions{
+	bs, err := setupBrowserAndSignals(c.CrawlOptions, crawl.CrawlerOptions{
 		Depth:    c.Depth,
 		MaxPages: c.MaxPages,
 		Timeout:  c.Timeout,
 		Scope:    c.Scope,
 		Headless: c.Headless,
+	})
+	if err != nil {
+		return err
 	}
-
-	// Create browser before signal setup so force-exit can clean up temp dirs.
-	var browserMgr *crawl.BrowserManager
-	if c.Headless {
-		var err error
-		browserMgr, err = crawl.NewBrowserManager(crawl.BrowserOptions{Headless: true})
-		if err != nil {
-			return fmt.Errorf("launch browser: %w", err)
-		}
-		defer browserMgr.Close()
-		opts.BrowserMgr = browserMgr
-	}
+	defer bs.cleanup()
 
 	if c.Verbose {
 		fmt.Fprintf(os.Stderr, "crawling %s (depth=%d, max-pages=%d, timeout=%s)\n",
-			c.URL, opts.Depth, opts.MaxPages, opts.Timeout)
+			c.URL, bs.opts.Depth, bs.opts.MaxPages, bs.opts.Timeout)
 	}
 
-	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
-	defer stop()
-
-	setupForceExitHandler(ctx, os.Stderr, func() {
-		if browserMgr != nil {
-			browserMgr.Close()
-		}
-	}, os.Exit)
-
-	requests, err := doCrawl(ctx, os.Stderr, c.URL, c.Header, opts)
+	requests, err := doCrawl(bs.ctx, os.Stderr, c.URL, c.Header, bs.opts)
 	if err != nil {
 		return err
 	}
@@ -415,7 +428,7 @@ func (c *ScanCmd) Run() error {
 		fmt.Fprintf(os.Stderr, "generating REST spec\n")
 	}
 
-	spec, err := generateSpec(ctx, requests, "rest", c.Confidence, c.Probe, c.Verbose)
+	spec, err := generateSpec(bs.ctx, requests, "rest", c.Confidence, c.Probe, c.Verbose)
 	if err != nil {
 		return err
 	}
--- a/cmd/vespasian/main.go
+++ b/cmd/vespasian/main.go
@@ -108,6 +108,13 @@ func isTokenChar(c byte) bool {
 	}
 }
 
+// shutdownBackstop is the maximum time doCrawl waits for Crawl() to return
+// after context cancellation. Crawl() has internal bounded shutdown (~2.5s),
+// so this is a defense-in-depth timeout that avoids blocking indefinitely if
+// the engine is stuck. The force-exit handler (second signal) is the final
+// safety net beyond this.
+const shutdownBackstop = 10 * time.Second
+
 // doCrawl executes the common crawl pipeline: parse headers, create crawler,
 // run the crawl with the provided context, and return the results. On graceful
 // shutdown (SIGINT/SIGTERM) partial results are returned instead of an error.
@@ -122,7 +129,42 @@ func doCrawl(ctx context.Context, stderr io.Writer, targetURL string, rawHeaders
 	opts.Stderr = stderr
 
 	crawler := crawl.NewCrawler(opts)
-	requests, err := crawler.Crawl(ctx, targetURL)
+
+	// Run Crawl in a goroutine so we can apply a backstop timeout after
+	// context cancellation. Without this, a stuck Crawl() blocks forever.
+	//
+	// Goroutine lifecycle: if the backstop fires before Crawl() returns,
+	// the goroutine remains alive until Crawl() eventually completes.
+	// The buffered channel (capacity 1) ensures it won't block on send.
+	// In a CLI context the process exits shortly after, and the force-exit
+	// handler (second SIGINT) is the final safety net.
+	type crawlResult struct {
+		requests []crawl.ObservedRequest
+		err      error
+	}
+	ch := make(chan crawlResult, 1)
+	go func() {
+		reqs, crawlErr := crawler.Crawl(ctx, targetURL)
+		ch <- crawlResult{reqs, crawlErr}
+	}()
+
+	var requests []crawl.ObservedRequest
+	select {
+	case r := <-ch:
+		requests, err = r.requests, r.err
+	case <-ctx.Done():
+		// Context cancelled (signal or deadline). Give Crawl() up to
+		// shutdownBackstop to finish its bounded internal shutdown.
+		backstop := time.NewTimer(shutdownBackstop)
+		select {
+		case r := <-ch:
+			requests, err = r.requests, r.err
+		case <-backstop.C:
+			return nil, fmt.Errorf("crawl failed: shutdown timed out after %s", shutdownBackstop)
+		}
+		backstop.Stop()
+	}
+
 	if err != nil {
 		if (errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)) && len(requests) > 0 {
 			fmt.Fprintf(stderr, "returning %d partial results\n", len(requests))
@@ -226,17 +268,21 @@ func setupBrowserAndSignals(crawlOpts CrawlOptions, extraOpts crawl.CrawlerOptio
 
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 
-	setupForceExitHandler(ctx, os.Stderr, func() {
+	// Single closure for browser teardown, shared between the force-exit
+	// handler and the deferred cleanup. BrowserManager.Close() is already
+	// idempotent (sync.Once), but a shared closure makes the intent explicit
+	// and avoids duplicating the nil-check.
+	closeBrowser := func() {
 		if browserMgr != nil {
 			browserMgr.Close()
 		}
-	}, os.Exit)
+	}
+
+	setupForceExitHandler(ctx, os.Stderr, closeBrowser, os.Exit)
 
 	cleanup := func() {
 		stop()
-		if browserMgr != nil {
-			browserMgr.Close()
-		}
+		closeBrowser()
 	}
 
 	return browserSetupResult{
--- a/cmd/vespasian/main.go
+++ b/cmd/vespasian/main.go
@@ -69,8 +69,8 @@ func parseHeaders(raw []string) (map[string]string, error) {
 		if !isValidHeaderName(name) {
 			return nil, fmt.Errorf("header name contains invalid characters (RFC 7230): %q", h)
 		}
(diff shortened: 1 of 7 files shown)


logan-bayless:
## Security Review — Backend Security Agent

**Verdict: APPROVED** — No critical or high-severity vulnerabilities found.

### Findings Summary

| ID | Severity | Finding | Justification |
|---|---|---|---|
| FINDING-01 | LOW | Goroutine leak possible when 2s grace period expires before crawl goroutine exits | Acceptable for CLI — Chrome is dead at this point, no network activity, process is exiting. If vespasian becomes a library, this should be revisited. |
| FINDING-02 | LOW | `os.Exit(1)` in double-interrupt handler bypasses deferred cleanup (temp dirs) | Intentional design — user explicitly requested forced termination via second `^C`. Chrome is already dead from first signal's `Kill()`. Temp dir leak is minor disk hygiene, not a security issue. |
| FINDING-03 | INFO | `NoSandbox` option exists in `BrowserOptions` | Never enabled in production code — defaults to `false`. Exists for future CI/container environments where Chrome sandboxing is unavailable. |
| FINDING-04 | INFO | `ChromePath` accepts arbitrary binary paths | Not exposed via CLI flags — internal API only. No user-controlled input reaches this field. |
| FINDING-05–11 | PASS | Race conditions (`sync.Mutex`/`sync.Once`), URL validation (defense-in-depth at two layers), CRLF injection prevention, error message safety, resource cleanup ordering, body size limits | All correctly implemented with test coverage. |

### Verification

| Check | Result |
|-------|--------|
| `go test ./...` | PASS (all 8 packages) |
| `go test -race ./...` | PASS (no data races detected) |
| `go test -tags integration` (BrowserManager) | PASS (4/4 — real Chrome launch/kill/cleanup) |
| `go vet ./...` | PASS |
| OWASP Top 10 assessment | All applicable categories PASS |
| Go-specific patterns (race conditions, context propagation, resource cleanup) | All PASS |

Full report: `.claude/.output/agents/2026-03-06-browser-lifecycle-security-review/backend-security-security-review.md`

santiago-praetorian (changes requested):
## Review Summary

Solid PR that fixes a real safety issue (LAB-1378). The approach — owning the Chrome lifecycle via `go-rod/rod/lib/launcher` and killing it immediately on signal — is architecturally sound. Code is clean, well-commented, and the multi-commit history shows good iterative refinement.

**All checks pass:** build ✅, unit tests ✅, race detection ✅, go vet ✅, gofmt ✅

---

### Backend Findings

#### Medium: Goroutine leak on signal timeout (`pkg/crawl/crawler.go:200-204`)

When the 2s timer fires before `crawlErr` returns, the goroutine running `engine.Crawl()` is leaked. The comment acknowledges this, and it's acceptable for a CLI that exits shortly after — but it prevents clean use as a library where `Crawl()` might be called multiple times in a long-lived process.

```go
select {
case <-crawlErr:
case <-timer.C:
    // Goroutine leaked but Chrome is dead — no network activity.
}
```

The PR description explicitly mentions "programmatic callers (agentic workflows)" as a use case. If that's a real scenario, this needs a cleanup mechanism or at minimum a documented limitation that callers should exit the process after a signal-interrupted crawl.

#### Low: `Cleanup()` not idempotent (`pkg/crawl/browser.go:80-82`)

`Kill()` is protected by `sync.Once`, but `Cleanup()` is not. In the signal path, `Kill()` is called at `crawler.go:192`, then `defer browserMgr.Close()` (line 93) calls `Kill()` (no-op) + `Cleanup()` again. Verify go-rod's `Cleanup()` handles double calls, or add `sync.Once` protection.

#### Nit: Consider making `Cleanup()` unexported (`pkg/crawl/browser.go:78-89`)

Having both `Cleanup()` and `Close()` as public methods is confusing. `Close()` is the intended public API — `Cleanup()` could be unexported.

### Security Findings

All clear. Chrome sandbox enabled by default ✅, no user input flows into launch params ✅, process handle (not PID) used for Kill ✅, signal race properly fixed ✅.

#### Info: `os.Exit(1)` in force-exit bypasses deferred cleanup (`cmd/vespasian/main.go:145`)

go-rod temp user data directories could persist on disk after double-^C. Not a security risk, but worth noting for operational awareness.

---

**Requesting changes** for the goroutine leak concern given the stated library/agentic use case. The other items are low/nit.

logan-bayless:
**Addressing review feedback:**

**Cleanup idempotency (Low) + Unexport Cleanup (Nit)** — Both fixed in e0097c5. `Cleanup()` is now `cleanup()` (unexported), protected by `sync.Once`, and only callable through `Close()`.

**Goroutine leak (Medium)** — Left as a clarified comment rather than a code change. Here's the reasoning:

The goroutine running `engine.Crawl()` is **not permanently leaked**. After the 2s timer fires, `engine.Close()` is called immediately (crawler.go:207), which causes `engine.Crawl()` to return, and the goroutine exits. The 2s timer is just how long we wait *synchronously* — not a deadline on the goroutine's lifetime.

**Concrete risk:** In a long-lived process that calls `Crawl()` repeatedly and interrupts each call, there's a brief window (milliseconds to low seconds) where the previous goroutine hasn't exited yet before the next `Crawl()` starts. This is not a resource accumulation problem — each goroutine exits on its own shortly after. There's no network activity (Chrome is dead), no memory growth, and no file descriptor leak.

For the current CLI use case, the process exits after one crawl. For the agentic/library use case, callers cancel the context and get results back — the brief goroutine overlap is benign. If we ever see evidence of a problem here (e.g., Katana's `engine.Close()` hanging), we'd address it at that layer. Adding a cleanup goroutine now would be complexity without a concrete benefit.

santiago-praetorian (dismissed):
## Review: APPROVE

All automated checks pass (build, tests, race detection, go vet, gofmt). The PR correctly fixes LAB-1378 by having vespasian own Chrome's lifecycle via go-rod's launcher and killing the browser immediately on signal.

### Key strengths
- Clean `BrowserManager` abstraction with `sync.Once` guards for idempotent Kill/Cleanup
- Bounded 2s shutdown drain — no unbounded waits
- Good separation: signal path in `crawler.go`, force-exit in `main.go`
- Solid test coverage for graceful shutdown logic (`returnPartialOnGraceful` extraction pattern)

### Minor items for future consideration (non-blocking)
- **Goroutine lifecycle on timer expiry** (`crawler.go:200-208`): if `engine.Crawl()` never returns after `engine.Close()`, the goroutine leaks. Acceptable for CLI usage but worth a second bounded wait if this becomes library code.
- **`engine.Close()` double-call on signal path** (explicit at line 211 + defer at line 167): verify Katana's `Close()` is idempotent or guard with `sync.Once`.
- **`NoSandbox` not wired through**: `BrowserOptions.NoSandbox` exists but is never set. May need plumbing for container/CI environments.
- **`Kill()` swallows errors**: consider storing the error for observability.

No security issues found. Good security patterns throughout (CRLF validation, URL scheme checks, body size limits, bounded drains, `0600` perms).

santiago-praetorian (changes requested):
## PR Review: feat: own browser lifecycle for immediate signal handling

### Build Status: FAIL

The PR does not compile. 5 compilation errors in `pkg/crawl/crawler.go`:

```
pkg/crawl/crawler.go:54:11: undefined: io
pkg/crawl/crawler.go:129:5: undefined: browserMgr
pkg/crawl/crawler.go:130:28: undefined: browserMgr
pkg/crawl/crawler.go:186:7: undefined: browserMgr
pkg/crawl/crawler.go:187:5: undefined: browserMgr
```

---

### Critical Findings

#### 1. `browserMgr` never declared in `Crawl()` (build failure)
**`pkg/crawl/crawler.go:129-130, 186-187`**

The variable `browserMgr` is referenced 4 times inside `Crawl()` but is never declared or initialized. The PR description states *"Crawl() creates BrowserManager when headless"* but the code that launches Chrome and assigns `browserMgr` is missing entirely.

#### 2. Missing `io` import (build failure)
**`pkg/crawl/crawler.go:54`**

`CrawlerOptions.Stderr` is typed `io.Writer` but `"io"` is not in the import block.

#### 3. Deadlock — double drain of `crawlErr` channel
**`pkg/crawl/crawler.go:169-212`**

The `<-crawlCtx.Done()` case drains `crawlErr` unconditionally at line 174. Then:

- **Signal path (line 196):** `<-crawlErr` is read again inside the `select`. Since the channel (cap 1) was already consumed, this blocks forever. The 2s timer saves it from a permanent hang, but the "drain buffered results" grace period becomes a no-op — the goroutine already exited.
- **MaxPages path (line 212):** `<-crawlErr` blocks forever with **no timer** — this is a real deadlock.

The shutdown sequence needs restructuring. Currently:
```
closeEngine() → drain crawlErr → kill browser → wait on crawlErr (already drained!)
```

Should be:
```
// Signal path: kill browser first, then bounded-drain, then close engine
// MaxPages path: close engine, then drain crawlErr (once)
```

#### 4. Redundant `engine.Close()` bypassing `closeOnce`
**`pkg/crawl/crawler.go:206, 211`**

Lines 206 and 211 call `_ = engine.Close()` directly, bypassing the `closeOnce` guard established at lines 154-155. Since `closeEngine()` already called `engine.Close()` at line 173, these are double-closes.

---

### Medium Findings

#### 5. `setupForceExitHandler` skips deferred cleanup
**`cmd/vespasian/main.go:144-146`**

`os.Exit(1)` on double-interrupt bypasses all deferred functions, including `browserMgr.Close()` and temp directory cleanup. Could leave Chrome temp dirs on disk containing target browsing data.

#### 6. `setupForceExitHandler` not testable
**`cmd/vespasian/main.go:138-147`**

Hard-codes `os.Stderr` and `os.Exit(1)`. Consider making exit function and writer injectable (same pattern used for `doCrawl`'s `stderr` param).

#### 7. No unit tests for the `Crawl()` shutdown orchestration
The new signal-handling logic (lines 178-208) has no test coverage. `TestDoCrawl_gracefulShutdownCondition` only validates the `doCrawl` wrapper, not the actual shutdown path.

---

### Summary

The core approach — owning Chrome via `go-rod/rod/lib/launcher` and killing it on signal — is sound. The `BrowserManager` abstraction is clean with proper `sync.Once` guards. The issues are:

1. Missing browser launch code in `Crawl()` (the PR appears incomplete)
2. Channel drain logic needs restructuring to avoid deadlocks
3. Redundant engine close calls should use the existing `closeOnce` wrapper

| Check | Status |
|-------|--------|
| Build | ❌ FAIL |
| Tests | ⏸️ Blocked by build |
| Race Detection | ⏸️ Blocked by build |
| PR ↔ Implementation | ❌ Mismatch (browser launch code missing) |
| Ticket Alignment | ⚠️ Correct approach, incomplete impl |

santiago-praetorian (changes requested):
## PR Review: feat: own browser lifecycle for immediate signal handling

### Build Status: FAIL

The PR does not compile. 5 compilation errors in `pkg/crawl/crawler.go`:

```
pkg/crawl/crawler.go:54:11: undefined: io
pkg/crawl/crawler.go:129:5: undefined: browserMgr
pkg/crawl/crawler.go:130:28: undefined: browserMgr
pkg/crawl/crawler.go:186:7: undefined: browserMgr
pkg/crawl/crawler.go:187:5: undefined: browserMgr
```

---

### Critical Findings

#### 1. `browserMgr` never declared in `Crawl()` (build failure)
**`pkg/crawl/crawler.go:129-130, 186-187`**

The variable `browserMgr` is referenced 4 times inside `Crawl()` but is never declared or initialized. The PR description states *"Crawl() creates BrowserManager when headless"* but the code that launches Chrome and assigns `browserMgr` is missing entirely.

#### 2. Missing `io` import (build failure)
**`pkg/crawl/crawler.go:54`**

`CrawlerOptions.Stderr` is typed `io.Writer` but `"io"` is not in the import block.

#### 3. Deadlock — double drain of `crawlErr` channel
**`pkg/crawl/crawler.go:169-212`**

The `<-crawlCtx.Done()` case drains `crawlErr` unconditionally at line 174. Then:

- **Signal path (line 196):** `<-crawlErr` is read again inside the `select`. Since the channel (cap 1) was already consumed, this blocks forever. The 2s timer saves it from a permanent hang, but the "drain buffered results" grace period becomes a no-op — the goroutine already exited.
- **MaxPages path (line 212):** `<-crawlErr` blocks forever with **no timer** — this is a real deadlock.

The shutdown sequence needs restructuring. Currently:
```
closeEngine() → drain crawlErr → kill browser → wait on crawlErr (already drained!)
```

Should be:
```
// Signal path: kill browser first, then bounded-drain, then close engine
// MaxPages path: close engine, then drain crawlErr (once)
```

#### 4. Redundant `engine.Close()` bypassing `closeOnce`
**`pkg/crawl/crawler.go:206, 211`**

Lines 206 and 211 call `_ = engine.Close()` directly, bypassing the `closeOnce` guard established at lines 154-155. Since `closeEngine()` already called `engine.Close()` at line 173, these are double-closes.

---

### Medium Findings

#### 5. `setupForceExitHandler` skips deferred cleanup
**`cmd/vespasian/main.go:144-146`**

`os.Exit(1)` on double-interrupt bypasses all deferred functions, including `browserMgr.Close()` and temp directory cleanup. Could leave Chrome temp dirs on disk containing target browsing data.

#### 6. `setupForceExitHandler` not testable
**`cmd/vespasian/main.go:138-147`**

Hard-codes `os.Stderr` and `os.Exit(1)`. Consider making exit function and writer injectable (same pattern used for `doCrawl`'s `stderr` param).

#### 7. No unit tests for the `Crawl()` shutdown orchestration
The new signal-handling logic (lines 178-208) has no test coverage. `TestDoCrawl_gracefulShutdownCondition` only validates the `doCrawl` wrapper, not the actual shutdown path.

---

### Summary

The core approach — owning Chrome via `go-rod/rod/lib/launcher` and killing it on signal — is sound. The `BrowserManager` abstraction is clean with proper `sync.Once` guards. The issues are:

1. Missing browser launch code in `Crawl()` (the PR appears incomplete)
2. Channel drain logic needs restructuring to avoid deadlocks
3. Redundant engine close calls should use the existing `closeOnce` wrapper

| Check | Status |
|-------|--------|
| Build | ❌ FAIL |
| Tests | ⏸️ Blocked by build |
| Race Detection | ⏸️ Blocked by build |
| PR ↔ Implementation | ❌ Mismatch (browser launch code missing) |
| Ticket Alignment | ⚠️ Correct approach, incomplete impl |

logan-bayless:
## Fixes Applied

A merge conflict when rebasing PR #27 (`fix/crawl-hygiene`) into this branch dropped critical code from `crawler.go`. The following commits resolve the regression, address all requested changes, and fix additional issues found during a post-fix security review.

### Merge Regression (64b4339)
- Restored `"io"` import dropped during merge conflict resolution.
- Restored `browserMgr` creation block (12 lines) that launches Chrome via `NewBrowserManager`.
- Restored correct shutdown logic that was replaced by main's simpler pattern, introducing a `crawlErr` double-drain deadlock.
- Integrated `closeOnce`/`closeEngine` pattern from main for idempotent engine close.

### Requested Changes (443540c)
- Made `setupForceExitHandler` injectable (`stderr`, `cleanup`, `exitFn`) for testability.
- Added `cleanup` callback called before `os.Exit` on double-interrupt for best-effort resource cleanup.
- Added shutdown orchestration tests: signal path, nil stderr, and MaxPages path.

### Security Review Findings (c8b5ec8)
- **Goroutine leak**: added bounded 500ms wait after `closeEngine()` when the 2s shutdown timer expires.
- **Force-exit temp dir leak**: lifted browser creation to commands and wired `browserMgr.Close()` into the force-exit cleanup callback.
- **Weak header validation**: header names now validated against RFC 7230 token production, not just CRLF.
- **WS URL sensitivity**: added security comment noting the DevTools URL grants full browser control.
- **Empty-host SSRF**: added `u.Host == ""` check to `crawler.go` URL validation for defense-in-depth.
- **BodyReadSize gap**: documented the intentional 10MB read / 1MB retain gap for Katana link extraction.
- **Data race**: all `Crawl()` return paths now snapshot the results slice under the mutex.

### Test Quality (57e880a)
- Replaced shadow-copy test (`returnPartialOnGraceful`) with real `doCrawl` integration test against a live httptest server.
- Removed boilerplate constant-value tests that only assert literals equal themselves.
- Removed `TestCloseEngineOnce` that tested a local `sync.Once` recreation rather than production code.

santiago-praetorian (changes requested):
## PR Review: feat: own browser lifecycle for immediate signal handling

**Requesting changes** — no critical issues, but several medium findings that should be addressed before merge.

---

### Backend Findings

#### HIGH: Duplicated browser setup in CrawlCmd and ScanCmd (`cmd/vespasian/main.go`)

The browser creation + force-exit wiring block is copy-pasted verbatim between `CrawlCmd.Run()` and `ScanCmd.Run()` (~18 lines each). Any future change must be made in both places.

**Suggestion:** Extract a shared helper like `setupBrowserAndSignals(ctx, headless, opts)` that encapsulates the lifecycle.

#### MEDIUM: `onForceExit` is dead production code

`onForceExit` duplicates the inline logic of `setupForceExitHandler`'s goroutine, but the goroutine does NOT actually call `onForceExit`. Tests validate `onForceExit` while production runs different code — a false-confidence pattern.

**Suggestion:** Have `setupForceExitHandler`'s goroutine call `onForceExit(stderr, cleanup, exitFn)` so the extracted function is both tested AND used in production.

#### MEDIUM: Graceful shutdown test coverage reduced

The old 6-case table-driven `TestDoCrawl_gracefulShutdownCondition` was removed. The replacement `TestDoCrawl_GracefulShutdownReturnsPartialResults` only covers the `context.Canceled` path.

Missing coverage for:
- `context.DeadlineExceeded` path
- Signal path with zero results
- Non-context errors with partial results

**Suggestion:** Add at least a `context.DeadlineExceeded` test case (e.g., `context.WithTimeout` with a very short deadline).

#### MEDIUM: Dual browser-ownership path loses force-exit benefit

`Crawl()` has a fallback that creates its own `BrowserManager` internally when `opts.BrowserMgr` is nil. This internal browser lacks force-exit wiring, so signals won't immediately kill Chrome for callers who don't provide their own `BrowserManager`.

**Suggestion:** Add a doc comment on `CrawlerOptions.BrowserMgr` explaining that callers who want immediate signal-based browser termination MUST provide their own `BrowserManager`.

---

### Security Findings

#### MEDIUM: `NoSandbox` and `ChromePath` lack guardrails

`BrowserOptions.NoSandbox` disables Chrome's OS-level sandbox (primary exploit barrier). `ChromePath` passes directly to `launcher.Bin()` → `exec.Command`. Neither is currently wired to CLI flags, but both are exported fields that future callers could misuse.

**Suggestion:** Add doc comments warning:
- `NoSandbox` should only be used in containerized/CI environments
- `ChromePath` must be a trusted value, never user-controlled input

#### LOW: Force-exit cleanup panic could leave process stuck

If `cleanup()` panics inside `setupForceExitHandler`, `exitFn(1)` never runs. The graceful signal is already consumed, so the user would need SIGKILL.

**Suggestion:** Wrap cleanup in a recover:
```go
if cleanup != nil {
    func() {
        defer func() { recover() }()
        cleanup()
    }()
}
```

---

### Positive Observations

- `sync.Once` on `killOnce` and `cleanupOnce` — idempotent lifecycle management
- Bounded shutdown drain (2s + 500ms fallback) — no unbounded waits
- Result snapshot under mutex on all return paths — fixes data race with OnResult callback
- RFC 7230 header name validation — strictly more secure than previous CRLF-only check
- `u.Host == ""` defense-in-depth — catches `http:///path` edge case
- Injectable `stderr`/`exitFn` in `setupForceExitHandler` — fully testable

---

### Summary

| Severity | Count |
|----------|-------|
| High | 1 (DRY violation) |
| Medium | 5 (dead code, test gap, dual ownership, NoSandbox/ChromePath docs) |
| Low | 1 (recover wrapper) |

**Recommended pre-merge fixes:** B1 (extract shared helper), B2 (wire `onForceExit` into production), B3 (add DeadlineExceeded test), S1/S2 (doc comments on NoSandbox/ChromePath).

logan-bayless:
## Summary

Addresses all 7 findings from santiago-praetorian's review plus 3 LOW findings from follow-up security review.

### Review Round 1 — Original 7 Findings

**HIGH**
1. **B1 — DRY violation**: Extracted `setupBrowserAndSignals()` helper, eliminating ~18 lines of copy-pasted browser+signal setup between `CrawlCmd.Run()` and `ScanCmd.Run()`

**MEDIUM**
2. **B2 — Dead code**: `setupForceExitHandler` goroutine now calls `onForceExit()` instead of inlining duplicate logic
3. **B3 — Test gap**: Added `TestDoCrawl_DeadlineExceededReturnsPartialResults` for the `context.DeadlineExceeded` path
4. **B4 — Missing docs**: Added doc comment on `CrawlerOptions.BrowserMgr` explaining that callers must provide their own `BrowserManager` to get force-exit integration
5. **S1 — NoSandbox guardrail**: Added doc warning — containerized/CI-only
6. **S2 — ChromePath guardrail**: Added doc warning — trusted path only, never user-controlled
7. **S3 — Cleanup panic safety**: Wrapped `cleanup()` in `onForceExit` with `recover()` to guarantee `exitFn(1)` always runs; added `TestOnForceExit_CleanupPanicRecovery`

### Review Round 2 — LOW Findings

8. **FINDING-07 — Double Close**: Shared single `closeBrowser` closure between force-exit handler and deferred cleanup
9. **FINDING-10 — No shutdown timeout**: Added 10s backstop timer in `doCrawl` so a stuck `Crawl()` can't block indefinitely after context cancellation
10. **FINDING-12 — Lock contention**: Moved `MapResult()` outside mutex in `OnResult` callback, reducing contention from Katana's 10 concurrent workers

### Test Plan

- [x] All unit tests pass (`go test ./cmd/vespasian/ ./pkg/crawl/`)
- [x] Self-review passed (dead code, DRY, coverage, cleanup safety, e2e path tracing)
- [x] Backend review: APPROVE, no CRITICAL/HIGH findings
- [x] Security review: APPROVE, no CRITICAL/HIGH findings

santiago-praetorian (changes requested):
## PR Review: `feat: own browser lifecycle for immediate signal handling`

### Summary

This PR correctly solves LAB-1378 — Chrome is now owned by vespasian and killed immediately on first signal. The architecture (BrowserManager with `sync.Once` idempotency, two-signal escalation, bounded shutdown timers) is well-designed. A few items to address before merge.

### Status

| Check | Status |
|-------|--------|
| Build | ✅ PASS |
| Tests (`go test ./...`) | ✅ PASS |
| Race Detection | ⚠️ FAIL — upstream Katana/goleveldb race, not vespasian code |
| Code Formatting (`gofmt`) | ❌ NEEDS FIX — `pkg/crawl/crawler.go` |
| `go vet` | ✅ PASS |
| Ticket Alignment | ✅ ALIGNED |

---

### Must Fix

**1. `gofmt` needed on `crawler.go`**
`pkg/crawl/crawler.go` fails `gofmt -l`. Please run `gofmt -w pkg/crawl/crawler.go`.

### Should Fix

**2. No timeout on `<-crawlErr` in MaxPages path (crawler.go:257)**
When MaxPages is reached, `closeEngine()` is called then an unbounded `<-crawlErr` blocks. If `engine.Close()` hangs, this blocks forever. The signal path has bounded waits (2s + 500ms) but the MaxPages path does not — should be consistent.

**3. Chrome launched before header validation (main.go:309 → doCrawl:124)**
`setupBrowserAndSignals` launches Chrome *before* `doCrawl` validates headers. Invalid headers cause Chrome to launch and immediately close, wasting ~1-2s of startup. Suggest validating headers before browser launch.

**4. `onForceExit` silently swallows cleanup panic (main.go:237-239)**
The `recover()` catches panics from `cleanup()` without logging the panic value. In a force-exit scenario, knowing *why* cleanup failed is valuable. Suggest `fmt.Fprintf(stderr, ...)` the recovered value.

### Should Consider

**5. Goroutine leak documentation for `doCrawl` backstop (main.go:146-149)**
When the 10s backstop fires, the goroutine running `Crawl()` leaks. The inline comments acknowledge this for CLI use, but `doCrawl`'s reusable signature (accepts `ctx`, `stderr`) invites library use. Consider adding a godoc warning.

**6. `setupBrowserAndSignals` hardcodes `context.Background()` (main.go:269)**
PR summary claims "Programmatic callers: canceling context.Context triggers same graceful shutdown," but the command layer creates its context from `context.Background()` — no way to inject an external context. The claim holds for `doCrawl` but not for `Run()` methods.

**7. Signal handler goroutines accumulate in tests (main.go:220-228)**
`setupForceExitHandler` calls `signal.Notify` but never `signal.Stop`. Tests calling `setupBrowserAndSignals` register stale signal handler goroutines.

### Low / Nit

- Unused `stop` field in `browserSetupResult` (main.go:249) — already called inside `cleanup`
- Magic number 500ms drain timer (crawler.go:238) — not a named constant unlike `ShutdownGracePeriod`
- `PID()` on `BrowserManager` (browser.go:107) only used in tests — could move to test file
- `ChromePath` (browser.go:58) lacks runtime validation — comment-only protection against command injection. Not exposed via CLI today, but consider validation or removing until needed.

### Positive Observations

- SSRF protection is thorough (pre-flight DNS + connect-time re-validation)
- Header injection prevention is correct (RFC 7230 tchar + CRLF rejection)
- `sync.Once` on Kill/cleanup prevents concurrent-call races
- `onForceExit` has panic recovery ensuring `exitFn` always runs
- URL validation consistent across CLI and crawler layers
- Output file permissions restrictive (`0600`)

logan-bayless:
## Addressing All Review Findings

All items from every review round have been addressed across commits `0ef1dda` through `5a05f9b`. Here is the complete resolution matrix:

### Latest Review (Round 6)

| # | Severity | Finding | Resolution |
|---|----------|---------|------------|
| 1 | Must Fix | `gofmt` needed on `crawler.go` | Fixed — applied `gofmt -w` alignment fixes in `398fe07` |
| 2 | Should Fix | No timeout on `<-crawlErr` in MaxPages path (`crawler.go:257`) — unbounded block if `engine.Close()` hangs | Fixed — added `ShutdownGracePeriod` bounded select, matching signal path's timeout discipline (`398fe07`) |
| 3 | Should Fix | Chrome launched before header validation — invalid headers waste ~1-2s browser startup | Fixed — moved `parseHeaders()` into `setupBrowserAndSignals()` before `NewBrowserManager()` so invalid headers fail fast (`398fe07`) |
| 4 | Should Fix | `onForceExit` silently swallows cleanup panic — no visibility into why cleanup failed | Fixed — `recover()` now logs panic value via `fmt.Fprintf(stderr, "cleanup panicked: %v\n", r)` (`398fe07`) |
| 5 | Consider | Goroutine leak undocumented for `doCrawl` backstop — reusable signature invites library use | Fixed — added godoc warning on `doCrawl` documenting goroutine lifecycle and library caller implications (`398fe07`) |
| 6 | Consider | `setupBrowserAndSignals` hardcodes `context.Background()` — no way to inject external context | Acknowledged — `doCrawl` and `Crawl()` both accept `ctx` for programmatic callers. The command layer's `context.Background()` is correct for CLI usage where `signal.NotifyContext` is the cancellation source |
| 7 | Consider | Signal handler goroutines accumulate in tests — `signal.Notify` never `signal.Stop`'d | Acknowledged — the goroutine blocks on channel receive (no CPU), and test processes exit anyway. Adding a done channel would add complexity for minimal benefit |
| — | Nit | Unused `stop` field in `browserSetupResult` | Fixed — removed field (`398fe07`) |
| — | Nit | Magic number 500ms drain timer | Fixed — extracted `DrainTimeout` named constant (`398fe07`) |
| — | Nit | `PID()` only used in tests | Kept — it's 3 lines and useful for integration test assertions. Moving to test file would require `//go:linkname` or exporting the launcher field |
| — | Nit | `ChromePath` lacks runtime validation | Acknowledged — not exposed via CLI. Doc comment warns it must be a trusted, hardcoded path (addressed in round 5) |

### Round 5 Findings (all resolved before round 6)

| # | Severity | Finding | Resolution |
|---|----------|---------|------------|
| B1 | High | Duplicated browser setup in CrawlCmd and ScanCmd | Fixed — extracted `setupBrowserAndSignals()` shared helper (`5742b86`) |
| B2 | Medium | `onForceExit` is dead production code — goroutine doesn't call it | Fixed — `setupForceExitHandler` goroutine now calls `onForceExit(stderr, cleanup, exitFn)` (`5742b86`) |
| B3 | Medium | Graceful shutdown test coverage reduced — missing `DeadlineExceeded` path | Fixed — added `TestDoCrawl_DeadlineExceededReturnsPartialResults` (`57e880a`) |
| B4 | Medium | Dual browser-ownership path undocumented | Fixed — added doc comment on `CrawlerOptions.BrowserMgr` explaining force-exit wiring requirement (`5742b86`) |
| S1 | Medium | `NoSandbox` lacks guardrails | Fixed — doc comment warns: "only set in containerized or CI environments" (`c8b5ec8`) |
| S2 | Medium | `ChromePath` lacks guardrails | Fixed — doc comment warns: "must be a trusted, hardcoded path. Never populate from user-controlled input" (`c8b5ec8`) |
| S3 | Low | Force-exit cleanup panic could leave process stuck | Fixed — `onForceExit` wraps cleanup in `recover()` with panic logging (`5742b86`, enhanced in `398fe07`) |

### Additional Fix Found During Self-Review

| # | Severity | Finding | Resolution |
|---|----------|---------|------------|
| — | Medium | `ScanCmd.Run()` passes already-cancelled `bs.ctx` to `generateSpec` after signal-interrupted crawl — probing bails out immediately on partial results | Fixed — create fresh `signal.NotifyContext` for the generate phase (`5a05f9b`) |

### Current Status

| Check | Status |
|-------|--------|
| Build | ✅ PASS |
| Tests (`go test ./...`) | ✅ PASS (all 8 packages) |
| `gofmt -l` | ✅ PASS (no files) |
| `go vet` | ✅ PASS |

santiago-praetorian (changes requested):
## PR Review — feat: own browser lifecycle for immediate signal handling

Overall this is a well-designed fix for a critical safety bug. The three-tier signal handling (graceful → backstop → force-exit) is solid, and the browser ownership model correctly addresses Katana's disconnected context. A few items to address before merge:

### Must-fix

**B1 (MEDIUM): Signal-path snapshot timing loses late results**
`pkg/crawl/crawler.go:250-256`

The signal shutdown path takes a snapshot of results *before* `closeEngine()` and the drain wait complete. Late-arriving results from Katana callbacks that fire between the snapshot and engine close are lost. The MaxPages path (lines 259-273) correctly snapshots *after* engine close and drain — the signal path should follow the same ordering.

### Recommended improvements

**S1 (MEDIUM): Unexport `WSURL()`**
`pkg/crawl/browser.go:79`

`WSURL()` grants full browser control and is only consumed within the `crawl` package (`crawler.go:166`). Making it unexported (`wsURL()`) reduces attack surface.

**S2 (LOW): Add null byte rejection to header value validation**
`cmd/vespasian/main.go:72`

The CRLF check doesn't reject `\x00`, which some HTTP implementations treat as string terminators:
```go
strings.ContainsAny(value, "\r\n\x00")
```

**B2 (LOW): Katana dependency data race**
Race detector found a data race in Katana v1.4.0's goleveldb/custom_field.go during `NewCrawlerOptions`. Not vespasian code, but worth tracking as a known issue.

### Positive observations

- Force-exit handler correctly waits for `ctx.Done()` before registering for second signal — avoids the race where both handlers consume the same first signal
- `sync.Once` guards on Kill/cleanup make `BrowserManager` concurrency-safe
- Panic recovery in `onForceExit` guarantees `exitFn` runs even if cleanup panics
- Fresh signal context for generate phase after interrupted crawl is correct
- RFC 7230 header name validation is thorough
- Good test coverage of shutdown paths (signal, deadline, MaxPages)

logan-bayless (commented):
Thanks for the thorough review! Here's how each item was addressed in 3b319da:

**B1 (MEDIUM) — Signal-path snapshot timing loses late results**
No change — false positive. The signal path sequences: Kill (line 221) → ShutdownGracePeriod drain (227-234) → closeEngine (237) → DrainTimeout drain if needed (239-248) → snapshot (252-255). The snapshot is the last operation, after all shutdown and draining. The signal path actually has two drain windows (pre-close and post-close) compared to MaxPages' single post-close drain.

**S1 (MEDIUM) — Unexport `WSURL()`**
Implemented:
1. Renamed struct field `wsURL` → `wsEndpoint` to avoid Go field/method name collision
2. Renamed method `WSURL()` → `wsURL()` (now package-private)
3. Updated constructor assignment in `browser.go:69`
4. Updated call site in `crawler.go:166`
5. Updated test references in `browser_test.go:48,50,53`

**S2 (LOW) — Add null byte rejection to header validation**
Implemented:
1. Added `\x00` to `strings.ContainsAny` check in `main.go:72`
2. Updated error message from "invalid CRLF characters" to "invalid characters"
3. Added test case for null byte in header value to `TestParseHeaders_CRLFInjection`

**B2 (LOW) — Katana dependency data race**
No change — upstream issue in Katana v1.4.0's `goleveldb/custom_field.go`. Acknowledged as known; will revisit when Katana ships a fix.

santiago-praetorian (approved):
## Review: APPROVE

Build, tests, `gofmt`, and `go vet` all pass. The PR correctly fixes LAB-1378 by owning Chrome's lifecycle via `go-rod/rod/lib/launcher` and killing it immediately on signal. All findings from prior review rounds have been addressed.

### Remaining observations (non-blocking)

**MEDIUM: `backstop.Stop()` ordering in `doCrawl` (`cmd/vespasian/main.go:167`)**

When the `ctx.Done()` branch fires and the inner select reads from `<-ch` (line 162), `backstop.Stop()` at line 167 correctly releases the timer. However, when `<-backstop.C` fires instead, `backstop.Stop()` is a no-op since the timer already expired. The current code is functionally correct — just noting the ordering is slightly asymmetric. No action needed.

**LOW: `setupForceExitHandler` goroutine outlives tests (`cmd/vespasian/main.go:222-229`)**

Each call to `setupBrowserAndSignals` in tests leaks a goroutine blocked on `<-sigCh` and a `signal.Notify` registration. Benign for the current test suite size. Previously discussed and acknowledged.

### Positive observations

- `sync.Once` guards on Kill/cleanup make `BrowserManager` concurrency-safe
- Three-tier signal handling (graceful → backstop → force-exit) is well-designed
- Bounded shutdown drains on both signal and MaxPages paths
- Panic recovery in `onForceExit` guarantees `exitFn` runs
- RFC 7230 header validation + null byte rejection is thorough
- Fresh signal context for generate phase after interrupted crawl is correct
- Good test coverage across shutdown paths (signal, deadline, MaxPages)

PR #31, opened 2026-03-09
feat: enable jsluice-based JS endpoint extraction

eli-wald:
## Summary
- Enables `ScrapeJSLuiceResponses` in the Katana crawler configuration alongside the existing `ScrapeJSResponses` option.
- This leverages jsluice's tree-sitter-based JavaScript parser to extract endpoints from JS response bodies, improving API path discovery from JavaScript bundles.

## Caveat
jsluice integration is currently provided transitively through Katana's `ScrapeJSLuiceResponses` option. If we move away from Katana as our crawling engine in the future, we will need to re-introduce jsluice (or an equivalent JS endpoint extraction library) as a direct dependency.

## Test plan
- [x] Run both `vespasian-main` and `vespasian-jsluice` binaries against the same target and compare captured endpoints
- [x] Verify jsluice-discovered paths are valid API endpoints (not false positives from string literals)
- [x] Confirm no performance regression on large JS-heavy targets

### DIFF
--- a/pkg/crawl/crawler.go
+++ b/pkg/crawl/crawler.go
@@ -132,15 +132,15 @@ func (c *Crawler) Crawl(ctx context.Context, targetURL string) ([]ObservedReques
 		// Katana needs the full body for link extraction and JS parsing to maximize
 		// crawl coverage; we only retain MaxResponseBodySize for classification.
 		// Peak memory: up to Concurrency × BodyReadSize (100 MB with 10 workers).
-		BodyReadSize:      10 * 1024 * 1024,
-		Concurrency:       10,
-		Parallelism:       10,
-		RateLimit:         150,
-		TimeStable:        3, // seconds to wait for DOM stability; 0 causes go-rod panic in time.NewTicker
-		ScrapeJSResponses: true,
+		BodyReadSize:           10 * 1024 * 1024,
+		Concurrency:            10,
+		Parallelism:            10,
+		RateLimit:              150,
+		TimeStable:             3, // seconds to wait for DOM stability; 0 causes go-rod panic in time.NewTicker
+		ScrapeJSResponses:      true,
 		ScrapeJSLuiceResponses: true,
-		XhrExtraction:     true,
-		Silent:            true,
+		XhrExtraction:          true,
+		Silent:                 true,
 		OnResult: func(result output.Result) {
 			// Map result outside the lock — MapResult may do URL parsing
 			// and body truncation, which is wasted work under contention.
--- a/pkg/crawl/crawler.go
+++ b/pkg/crawl/crawler.go
@@ -76,22 +76,22 @@ func (c *Crawler) Crawl(ctx context.Context, targetURL string) ([]ObservedReques
 
 	// Build Katana options
 	katanaOpts := &types.Options{
-		MaxDepth:          c.opts.Depth,
-		Timeout:           PageTimeout,
-		CrawlDuration:     c.opts.Timeout,
-		FieldScope:        MapScope(c.opts.Scope),
-		Headless:          c.opts.Headless,
-		CustomHeaders:     ToStringSlice(c.opts.Headers),
-		Strategy:          "depth-first",
-		BodyReadSize:      10 * 1024 * 1024, // 10 MB limit to prevent memory exhaustion from hostile targets
-		Concurrency:       10,
-		Parallelism:       10,
-		RateLimit:         150,
-		TimeStable:        3, // seconds to wait for DOM stability; 0 causes go-rod panic in time.NewTicker
+		MaxDepth:               c.opts.Depth,
+		Timeout:                PageTimeout,
+		CrawlDuration:          c.opts.Timeout,
+		FieldScope:             MapScope(c.opts.Scope),
+		Headless:               c.opts.Headless,
+		CustomHeaders:          ToStringSlice(c.opts.Headers),
+		Strategy:               "depth-first",
+		BodyReadSize:           10 * 1024 * 1024, // 10 MB limit to prevent memory exhaustion from hostile targets
+		Concurrency:            10,
+		Parallelism:            10,
+		RateLimit:              150,
+		TimeStable:             3, // seconds to wait for DOM stability; 0 causes go-rod panic in time.NewTicker
 		ScrapeJSResponses:      true,
 		ScrapeJSLuiceResponses: true,
-		XhrExtraction:     true,
-		Silent:            true,
+		XhrExtraction:          true,
+		Silent:                 true,
 		OnResult: func(result output.Result) {
 			mu.Lock()
 			defer mu.Unlock()
--- a/pkg/crawl/crawler.go
+++ b/pkg/crawl/crawler.go
@@ -137,7 +137,7 @@ func (c *Crawler) Crawl(ctx context.Context, targetURL string) ([]ObservedReques
 		RateLimit:         150,
 		TimeStable:        3, // seconds to wait for DOM stability; 0 causes go-rod panic in time.NewTicker
 		ScrapeJSResponses: true,
-    ScrapeJSLuiceResponses: true,
+		ScrapeJSLuiceResponses: true,
 		XhrExtraction:     true,
 		Silent:            true,
 		OnResult: func(result output.Result) {
--- a/pkg/crawl/crawler.go
+++ b/pkg/crawl/crawler.go
@@ -181,6 +181,8 @@ func MapResult(r output.Result) ObservedRequest {
 			req.Body = req.Body[:MaxResponseBodySize]
 		}
 		req.Source = r.Request.Source
+		req.Tag = r.Request.Tag
+		req.Attribute = r.Request.Attribute
 	}
 
 	// Parse query params from URL
--- a/pkg/crawl/crawler.go
+++ b/pkg/crawl/crawler.go
@@ -120,7 +120,6 @@ func (c *Crawler) Crawl(ctx context.Context, targetURL string) ([]ObservedReques
 
 	// Build Katana options
 	katanaOpts := &types.Options{
-
 		MaxDepth:      c.opts.Depth,
 		Timeout:       PageTimeout,
 		CrawlDuration: c.opts.Timeout,
--- a/pkg/crawl/crawler.go
+++ b/pkg/crawl/crawler.go
@@ -88,7 +88,8 @@ func (c *Crawler) Crawl(ctx context.Context, targetURL string) ([]ObservedReques
 		Parallelism:       10,
 		RateLimit:         150,
 		TimeStable:        3, // seconds to wait for DOM stability; 0 causes go-rod panic in time.NewTicker
-		ScrapeJSResponses: true,
+		ScrapeJSResponses:      true,
+		ScrapeJSLuiceResponses: true,
 		XhrExtraction:     true,
 		Silent:            true,
 		OnResult: func(result output.Result) {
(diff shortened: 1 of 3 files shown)


eli-wald:
Manual testing: ran this against a [BugForge](https://app.bugforge.io/dashboard) SPA with and without JSluice to confirm further coverage. Didn't notice significant performance regressions. 

- [x] Run both vespasian-main and vespasian-jsluice binaries against the same target and compare captured endpoints
- [ ] Verify jsluice-discovered paths are valid API endpoints (not false positives from string literals)
- [x]  Confirm no performance regression on large JS-heavy targets

One interesting case: scans/crawls will pick up path parameters formatted with a colon, for instance `/orders/:orderId`. Not sure if we should/can add steps to scan with other values in the path parameter here, or if that's out of scope for a tool like Vespasian.

santiago-praetorian:
## LAB-1375

## Code Review: `feat: enable jsluice-based JS endpoint extraction`

### Status Table

| Check | Status |
|-------|--------|
| Build | **PASS** |
| Tests | **PASS** (all packages) |
| Race Detection | **PASS** |
| Code Formatting | **NEEDS FIX** (`gofmt -l` flags `crawler.go`) |
| Dev Artifacts | **CLEAN** |
| PR Description Match | **MATCH** |
| Ticket Alignment | **PARTIAL** |
| Coverage (`pkg/crawl`) | 52.3% |

---

### Backend Review Findings

**[MEDIUM] Ticket overclaims — jsluice does NOT extract method/headers/body in Katana v1.4.0**
LAB-1375 says jsluice will "add additional context for each path (method, headers, body params)" and help with "HTTP methods on endpoints not being found." However, Katana's jsluice integration (`parser_generic.go:50`) hardcodes `Method: http.MethodGet` with no headers or body. The method/headers/body context claim is **not fulfilled** by this one-line change alone. Please clarify whether that's expected follow-up work or a gap in this PR.

**[LOW] jsluice type hint (`Attribute`) lost in MapResult**
When jsluice finds endpoints, Katana sets `Attribute` to values like `jsluice-fetch` or `jsluice-xhr`. But `MapResult` doesn't propagate these fields, and `ObservedRequest` has no such fields. This metadata could help infer HTTP methods downstream — which is exactly what the ticket wants. Consider adding an `Attribute` or `Tag` field to `ObservedRequest` and mapping it in `MapResult`.

**[LOW] No jsluice-specific test cases**
`crawler_test.go` has thorough `MapResult` tests but none with jsluice-realistic source values (e.g., `Source: "jsluice-fetch"`). Low risk since jsluice results use the same `output.Result` → `MapResult` pipeline, but would improve confidence.

**[NIT] Column alignment formatting**
The whitespace alignment (`ScrapeJSResponses:      true,`) is flagged by `gofmt -l`. Run `gofmt -w pkg/crawl/crawler.go` to fix.

**[LOW] String concatenation claim IS valid** ✅
jsluice uses tree-sitter AST parsing rather than regex, correctly resolving JS string concatenation that `ScrapeJSResponses` misses. This is the primary value of the change and works as described.

---

### Security Review Findings

**[LOW] Resource consumption from tree-sitter parsing**
Tree-sitter AST parsing is more expensive than regex, but existing bounds (`BodyReadSize: 10MB`, `Concurrency: 10`, `MaxPages` cap, `CrawlDuration` timeout) adequately limit additional cost. No action required.

**[NIT] No new attack surface** — Extracted URLs flow into `ObservedRequest` as data only. No SSRF vector.

**[NIT] Dependency security** — jsluice published by BishopFox, pinned in `go.mod`, zero advisories. Already a compiled transitive dep via Katana — this flag only activates it at runtime.

---

### Recommendation: **REQUEST CHANGES**

The code change itself is correct, minimal, and low-risk. However:

1. **`gofmt`** — Fix the alignment on lines 101-102 (`gofmt -w pkg/crawl/crawler.go`).
2. **Ticket gap** — Please clarify the scope: does this PR intend to fully address LAB-1375, or is method/headers/body param extraction planned as follow-up? If follow-up, note that in the PR description.
3. **Suggested improvement** — Propagating jsluice's `Attribute` field through `MapResult` → `ObservedRequest` would enable method inference downstream and better fulfill LAB-1375's goals.

santiago-praetorian (changes requested):
See detailed review comment. Key items: (1) fix gofmt formatting, (2) clarify ticket scope — method/headers/body extraction is not fulfilled by this change alone, (3) consider propagating jsluice Attribute field through MapResult.

eli-wald:
## Updates addressing review feedback

### Changes in latest commits:
1. **Propagated `Tag` and `Attribute` fields** — Added `Tag` and `Attribute` to `ObservedRequest` and mapped them from Katana's `navigation.Request` in `MapResult`. This preserves jsluice's type hints (`jsluice-fetch`, `jsluice-xhr`) and HTMX attributes (`hx-get`, `hx-post`) in capture files, enabling downstream method inference.

2. **`gofmt` fix** — Resolved the column alignment formatting issue flagged in the review.

### Scope clarification
This PR enables jsluice and propagates its metadata. **Full method/headers/body param extraction** beyond what jsluice and Katana currently provide is **planned as follow-up work**. Specifically:
- Inferring HTTP methods from `Attribute` values (e.g., `jsluice-fetch` → POST, `hx-post` → POST)
- Extracting request headers/body from JS source (not supported by Katana's jsluice integration in v1.4.0, which hardcodes `Method: http.MethodGet`)

These are out of scope for this PR but are natural next steps once the `Attribute` field is available in capture files.

santiago-praetorian (changes requested):
## Review: Request Changes

Overall this is a clean, well-scoped PR. One actionable item before merge:

### Missing test coverage for Tag/Attribute propagation (Medium)

`TestMapResult_Normal` sets up a `navigation.Request` but does not populate `Tag` or `Attribute`. No test verifies these new fields are correctly mapped through `MapResult`. Please add `Tag` and `Attribute` values to the test fixture and assert they appear on the output `ObservedRequest`.

### Minor observations (non-blocking)

- **Tag/Attribute fields captured but not yet consumed downstream** — `pkg/classify`, `pkg/generate`, and `pkg/probe` have zero references to these fields. The PR description mentions "enabling downstream method inference" — consider documenting the intended consumer or noting it as future work.

- **No length validation on Tag/Attribute** — Unlike `Body` (truncated at `MaxResponseBodySize`), these string fields have no length cap. Low risk since values come from jsluice/Katana internals (typically short strings like `"script"`, `"jsluice-fetch"`, `"hx-get"`), but a defensive cap (e.g., 256 chars) in `MapResult` would be consistent with the existing truncation pattern.

### What looks good

- Build passes, all tests pass, no race conditions detected
- `gofmt` clean
- Existing security controls (MaxPages, BodyReadSize, scope, timeouts) adequately mitigate the increased crawl surface from jsluice
- `encoding/json` serialization handles escaping — no injection vectors from the new fields

santiago-praetorian:
[LAB-1375: Feature: Enable JSluice](https://linear.app/praetorianlabs/issue/LAB-1375/feature-enable-jsluice)

eli-wald:
Merge conflict addressed, should be ready for review.

santiago-praetorian (changes requested):
## Review: Request Changes

One actionable item before merge:

### [Medium] `gofmt` formatting violations in `crawler.go` and `crawler_test.go`

`gofmt -l` flags both files. Three issues:

1. **`pkg/crawl/crawler.go:141`** — `ScrapeJSLuiceResponses` is indented with spaces instead of tabs, causing column misalignment across the entire struct literal (lines 135–143).
2. **`pkg/crawl/crawler_test.go:443`** — Closing brace of `TestPageTimeout` uses 2-space indent instead of tab.
3. **`pkg/crawl/crawler_test.go:621`** — Closing brace of `TestMapResult_JsluiceTagAndAttribute` uses 2-space indent instead of tab.

Fix: `gofmt -w pkg/crawl/crawler.go pkg/crawl/crawler_test.go`

### [Low] Stray blank line in Katana options struct literal

**`pkg/crawl/crawler.go:123`** — An empty line between the opening `&types.Options{` and the first field `MaxDepth` appears unintentional and is inconsistent with the rest of the struct literal style.

eli-wald:
Manual inspection regarding these comments: Running `gofmt -w pkg/crawl/crawler.go pkg/crawl/crawler_test.go` did not produce any changes. I did not see an empty line at `pkg/crawl/crawler.go:123` or between `&types.Options{` and `MaxDepth` elsewhere in the file. I think these format issues were already addressed.

pkg/crawl/crawler.go
```diff
@@ -137,6 +138,7 @@ func (c *Crawler) Crawl(ctx context.Context, targetURL string) ([]ObservedReques
 		RateLimit:         150,
 		TimeStable:        3, // seconds to wait for DOM stability; 0 causes go-rod panic in time.NewTicker
 		ScrapeJSResponses: true,
+    ScrapeJSLuiceResponses: true,
```
santiago-praetorian: It seems this line uses spaces instead of tabs like the rest of the lines.

pkg/crawl/crawler_test.go
```diff
@@ -436,6 +436,13 @@ func TestCrawl_EmptyURLReturnsError(t *testing.T) {
 	}
 }
 
+// TestPageTimeout verifies the PageTimeout constant value
+func TestPageTimeout(t *testing.T) {
+	if PageTimeout != 30 {
+		t.Errorf("PageTimeout = %d, want 30", PageTimeout)
+  }
```
santiago-praetorian: This uses two spaces instead of a tab.

pkg/crawl/crawler_test.go
```diff
@@ -553,6 +560,67 @@ func TestMapResult_SmallBodyNotTruncated(t *testing.T) {
...
+			observed := MapResult(result)
+
+			if observed.Tag != tt.tag {
+				t.Errorf("Tag = %q, want %q", observed.Tag, tt.tag)
+			}
+			if observed.Attribute != tt.attribute {
+				t.Errorf("Attribute = %q, want %q", observed.Attribute, tt.attribute)
+			}
+		})
+  }
```
santiago-praetorian: This line uses two spaces instead of a tab.

pkg/crawl/crawler.go
```diff
@@ -120,6 +120,7 @@ func (c *Crawler) Crawl(ctx context.Context, targetURL string) ([]ObservedReques
 
 	// Build Katana options
 	katanaOpts := &types.Options{
+
```
santiago-praetorian: This is the empty line.

santiago-praetorian (commented):
It is odd that the formatter did not pick this up. I did a manual review and confirmed the format issues.

eli-wald:
Weird, thanks for pointing that out. Just updated this manually in commit [7b0db46](https://github.com/praetorian-inc/vespasian/pull/31/commits/7b0db4680e662ebda8077021306b327b9896c093)[7b0db46](https://github.com/praetorian-inc/vespasian/pull/31/commits/7b0db4680e662ebda8077021306b327b9896c093), should be addressed now.

santiago-praetorian:
It seems the issue with `pkg/crawl/crawler.go:140` (4 spaces instead of tab) remains.

santiago-praetorian (approved):
## Approve

The code change is correct, minimal, and well-tested. One new finding to document:

### [Low] Upstream data race in Katana v1.4.0 exposed by `ScrapeJSLuiceResponses`

`go test -race` consistently fails on `TestCrawl_MaxPagesPath_ReturnsNoError` when `ScrapeJSLuiceResponses: true` is enabled. The race is between `katana/pkg/output.loadCustomFields()` (map write at `custom_field.go:117`) and `katana/pkg/engine/parser.customFieldRegexParser()` (map read at `parser.go:720`) during Katana initialization. This does not reproduce on `main` where `ScrapeJSLuiceResponses` is disabled.

This is an upstream bug in Katana's internals, not in this PR's code. The race occurs during initialization (not steady-state crawling), so practical risk is low. Consider filing an issue with `projectdiscovery/katana` to track.

Reproduction: `go test -race -count=3 -run TestCrawl_MaxPagesPath ./pkg/crawl/...`

PR #33, opened 2026-03-10
fix: improve Burp XML import robustness

eli-wald:
## Summary

- **Allow DOCTYPE declarations** in Burp XML imports. Only ENTITY declarations (the actual XXE vector) are blocked. Burp Suite legitimately includes DOCTYPE with ELEMENT definitions in its exports.
- **Handle empty response data** gracefully. Burp exports can contain items with empty `<response>` elements (timeouts, connection resets). Previously this aborted the entire import with "no header/body separator found". Now these items are imported with the XML status code and empty response fields.

## Changes

- Removed `xxeDOCTYPEPatterns` list; renamed `containsXXEPatterns` to `containsXXEEntity`
- Updated error message from "DOCTYPE or ENTITY declarations not allowed" to "ENTITY declarations not allowed"
- Added early return in `parseItem()` when response bytes are empty
- Updated and added test cases covering both fixes

## Test plan

- [x] All 77 importer tests pass (`go test ./pkg/importer/... -v`)
- [x] `gofmt` clean on both modified files
- [x] Successfully imports real Burp XML export (`~/Documents/test-dvwa-burp`) that previously failed
- [x] ENTITY declarations still blocked (7 XXE prevention test cases)
- [x] DOCTYPE-only XML now parses successfully (4 test cases)
- [x] Empty response items imported with status code from XML (1 test case)

### DIFF
--- a/pkg/importer/burp.go
+++ b/pkg/importer/burp.go
@@ -63,26 +63,11 @@ func (BurpImporter) Name() string {
 	return "burp"
 }
 
-// containsXXEEntity checks for ENTITY declarations in XML data (XXE attack vector).
-// DOCTYPE declarations are allowed as they are standard in Burp Suite XML exports.
-func containsXXEEntity(data []byte) bool {
-	target := []byte("<!ENTITY")
-	for i := 0; i <= len(data)-len(target); i++ {
-		// Fast-path: skip positions that don't start with '<' to avoid
-		// calling EqualFold on every byte offset.
-		if data[i] == '<' && bytes.EqualFold(data[i:i+len(target)], target) {
-			return true
-		}
-	}
-	return false
-}
-
 // Import reads Burp Suite XML and converts to ObservedRequest format.
 func (i *BurpImporter) Import(r io.Reader) ([]crawl.ObservedRequest, error) {
 	// Limit reader to prevent resource exhaustion
 	limitedReader := newLimitedReader(r, maxImportSize)
 
-	// Read all content to check for entity declarations
 	data, err := io.ReadAll(limitedReader)
 	if err != nil {
 		return nil, fmt.Errorf("burp importer: failed to read input: %w", err)
@@ -91,11 +76,6 @@ func (i *BurpImporter) Import(r io.Reader) ([]crawl.ObservedRequest, error) {
 	// Check if file was too large
 	if limitedReader.hitLimit {
 		return nil, ErrFileTooLarge
-	}
-
-	// Check for ENTITY declarations (XXE attack vectors)
-	if containsXXEEntity(data) {
-		return nil, fmt.Errorf("burp importer: XML validation: ENTITY declarations not allowed")
 	}
 
 	var items burpItems
--- a/pkg/importer/burp.go
+++ b/pkg/importer/burp.go
@@ -138,6 +138,21 @@ func (i *BurpImporter) parseItem(item burpItem) (crawl.ObservedRequest, error) {
 		return crawl.ObservedRequest{}, fmt.Errorf("failed to parse request: %w", err)
 	}
 
+	// Handle empty response data (e.g., timeouts, connection resets)
+	if len(respBytes) == 0 {
+		return crawl.ObservedRequest{
+			Method:      method,
+			URL:         item.URL,
+			Headers:     headers,
+			QueryParams: extractQueryParams(item.URL),
+			Body:        body,
+			Response: crawl.ObservedResponse{
+				StatusCode: item.Status,
+			},
+			Source: "import:burp",
+		}, nil
+	}
+
 	// Parse HTTP response
 	statusCode, respHeaders, respBody, err := parseHTTPResponse(respBytes)
 	if err != nil {
--- a/pkg/importer/burp.go
+++ b/pkg/importer/burp.go
@@ -66,7 +66,13 @@ func (BurpImporter) Name() string {
 // containsXXEEntity checks for ENTITY declarations in XML data (XXE attack vector).
 // DOCTYPE declarations are allowed as they are standard in Burp Suite XML exports.
 func containsXXEEntity(data []byte) bool {
-	return bytes.Contains(bytes.ToUpper(data), []byte("<!ENTITY"))
+	target := []byte("<!ENTITY")
+	for i := 0; i <= len(data)-len(target); i++ {
+		if data[i] == '<' && bytes.EqualFold(data[i:i+len(target)], target) {
+			return true
+		}
+	}
+	return false
 }
 
 // Import reads Burp Suite XML and converts to ObservedRequest format.
--- a/pkg/importer/burp.go
+++ b/pkg/importer/burp.go
@@ -68,6 +68,8 @@ func (BurpImporter) Name() string {
 func containsXXEEntity(data []byte) bool {
 	target := []byte("<!ENTITY")
 	for i := 0; i <= len(data)-len(target); i++ {
+		// Fast-path: skip positions that don't start with '<' to avoid
+		// calling EqualFold on every byte offset.
 		if data[i] == '<' && bytes.EqualFold(data[i:i+len(target)], target) {
 			return true
 		}
@@ -144,6 +146,7 @@ func (i *BurpImporter) parseItem(item burpItem) (crawl.ObservedRequest, error) {
 			Body:        body,
 			Response: crawl.ObservedResponse{
 				StatusCode: item.Status,
+				Headers:    map[string]string{},
 			},
 			Source: "import:burp",
 		}, nil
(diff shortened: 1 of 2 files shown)


santiago-praetorian (changes requested):
## Review: Request Changes

### Status

| Check | Status |
|-------|--------|
| Build | ✅ PASS |
| Tests | ✅ PASS (all packages) |
| Race Detection | ✅ PASS |
| Code Formatting | ✅ PASS |
| Coverage | ✅ 94.9% overall; 100% on modified functions |

### Security — Medium: Incomplete case coverage in ENTITY pattern matching

`burp.go:40-43` — The `xxeENTITYPatterns` list has only 6 specific case combinations out of 64 possible casings for "ENTITY". A payload like `<!eNtItY xxe SYSTEM "file:///etc/passwd">` would bypass this check.

**Mitigating factors:** Go's `encoding/xml` does not process external entities regardless, so this is defense-in-depth. However, a `bytes.Contains(bytes.ToUpper(data), []byte("<!ENTITY"))` approach would be more robust and simpler than maintaining a hand-rolled case list.

Suggest replacing:
```go
xxeENTITYPatterns = [][]byte{
    []byte("<!ENTITY"), []byte("<!entity"), ...
}
```

With a single case-insensitive check:
```go
func containsXXEEntity(data []byte) bool {
    return bytes.Contains(bytes.ToUpper(data), []byte("<!ENTITY"))
}
```

### Security — Low: No status code validation on empty response path

When a response is empty, `item.Status` from XML is used directly without range validation. Acceptable since the input is a user-provided Burp export, but worth noting.

### Backend — Nit: Empty response test coverage

Consider adding a sub-test for the non-base64 empty response case (`<response></response>` without `base64="true"`) to cover both decode paths hitting the `len(respBytes) == 0` check.

---

Overall the changes are correct, well-tested, and address real Burp import failures. Requesting changes only for the ENTITY case-coverage hardening — the rest are minor suggestions.

santiago-praetorian:
[LAB-1414: Bug: Burp import rejects XML with DOCTYPE declaration](https://linear.app/praetorianlabs/issue/LAB-1414/bug-burp-import-rejects-xml-with-doctype-declaration)

santiago-praetorian (approved):
## Review: Approve

### Ticket Alignment

The PR correctly addresses [LAB-1414](https://linear.app/praetorianlabs/issue/LAB-1414/bug-burp-import-rejects-xml-with-doctype-declaration):
1. DOCTYPE declarations are now allowed — `xxeDOCTYPEPatterns` removed; only ENTITY is blocked.
2. ENTITY declarations remain blocked via case-insensitive `bytes.ToUpper` check.
3. Empty response handling added for timeouts/connection resets encountered in real Burp exports.

All prior review findings have been addressed in commit `4761fa6`.

### Backend

#### Low

1. **`burp.go:69` — `bytes.ToUpper` allocates a full copy of the input on every call.** The input can be up to 500MB (`maxImportSize`). This creates a transient ~500MB allocation for the ENTITY check. A streaming scan or case-folding search would avoid this. Functionally correct, but worth noting for memory pressure on large imports.

#### Nit

1. **`burp.go:132-143` — The empty-response path returns `Headers: nil` in `ObservedResponse`.** The non-empty path always returns a non-nil `Headers` map from `splitHTTPMessage`. Callers iterating over `Response.Headers` without a nil check could panic. Initializing to `map[string]string{}` would make behavior consistent between the two paths.

2. **`burp_test.go:725` — `base64.StdEncoding.EncodeToString([]byte(""))` in `TestBurpImporter_EmptyResponse`.** A comment clarifying this simulates Burp's `<response base64="true"></response>` for timed-out requests would aid readability.

### Security

#### Low

1. **`burp.go:69` — The `<!ENTITY` string scan is defense-in-depth, not a primary XXE mitigation.** Go's `encoding/xml` does not resolve external entities, so actual XXE risk is already mitigated by the standard library. The scan could produce false positives if a legitimate response body contains the literal string `<!ENTITY`. Acceptable trade-off, but the doc comment could note this limitation.

2. **`burp.go:94-95` — `decoder.Strict = true` but no custom `CharsetReader`.** Burp exports declaring non-UTF-8 encoding (e.g., `encoding="ISO-8859-1"`) would fail. Likely acceptable since Burp defaults to UTF-8, but could cause unexpected failures with non-standard configurations.

### Summary

All tests pass (77/77). Coverage is 100% on modified functions. Prior review findings resolved. No critical, high, or medium findings. Approving.

santiago-praetorian (changes requested):
## Review: Request Changes

### Backend — Low

**`burp.go:69` — `bytes.ToUpper` allocates a full copy of the input on every call.** The input can be up to 500MB (`maxImportSize`). This creates a transient ~500MB allocation for the ENTITY check. A case-folding search using `bytes.Index` with manual case comparison, or a streaming approach, would avoid this allocation.

eli-wald:
Zero-allocation implementation of containsXXEEntity should be in place now. Ready for review.

santiago-praetorian (changes requested):
## Review

### Backend

#### Low

1. **`burp.go:138-149` — Empty-response path returns `Headers: nil` and `Body: nil` in `ObservedResponse`, while the non-empty path always returns initialized values.** The non-empty path sets `Headers` to a non-nil `map[string]string` (from `splitHTTPMessage`) and `Body` to a `[]byte`. Callers that iterate over `Response.Headers` or check `Response.Body` length without nil guards could behave differently on these two paths. Initializing `Headers: map[string]string{}` would make the contract consistent.

#### Nit

1. **`burp.go:70` — The `<` guard before `bytes.EqualFold` is redundant.** `bytes.EqualFold(data[i:i+len(target)], target)` already checks whether `data[i]` matches `<` (fold-insensitively). The extra comparison adds a negligible fast-path short-circuit but also a subtle implication that it's doing something meaningful. Consider removing it or adding a brief comment noting it's a performance optimization.

2. **`burp_test.go:725-729` — `base64.StdEncoding.EncodeToString([]byte(""))` produces an empty string.** This is functionally correct but non-obvious. A short comment explaining this simulates Burp's `<response base64="true"></response>` for timed-out requests would aid readability.

### Security

#### Low

1. **`burp.go:68-76` — The `<!ENTITY` string scan is defense-in-depth and may produce false positives.** If a legitimate HTTP response body captured by Burp contains the literal string `<!ENTITY` (e.g., an XML tutorial page), the entire import will be rejected. Go's `encoding/xml` does not resolve external entities, so the actual XXE risk is already mitigated by the standard library. The doc comment on `containsXXEEntity` could note this false-positive possibility.

2. **`burp.go:100-101` — `decoder.Strict = true` with no custom `CharsetReader`.** Burp exports declaring a non-UTF-8 encoding in the XML declaration (e.g., `encoding="ISO-8859-1"`) would fail with an unsupported charset error. Burp defaults to UTF-8, so this is unlikely in practice.

eli-wald:
Checking for <!ENTITY> strings in Burp responses would produce false positives here, blocking Vespasian from importing burp files that happen to include <!ENTITY> strings. Given that `encoding/xml` does not resolve external entities, this check has been removed.

santiago-praetorian (approved):
## Review: Approve

### Ticket Alignment

The PR correctly addresses [LAB-1414](https://linear.app/praetorianlabs/issue/LAB-1414/bug-burp-import-rejects-xml-with-doctype-declaration):
1. DOCTYPE declarations now allowed — `xxeDOCTYPEPatterns` and `containsXXEPatterns` removed entirely.
2. ENTITY pre-scan removed to avoid false positives on legitimate response bodies. Go's `encoding/xml` does not resolve external entities, providing inherent XXE protection.
3. Empty response handling added — Burp items with empty `<response>` elements (timeouts, connection resets) are imported with the XML status code and empty response fields.

### Prior Review Findings — All Addressed

- Case-insensitive ENTITY matching → resolved by removing the pre-scan
- `bytes.ToUpper` 500MB allocation → resolved by removing the pre-scan
- `nil` Headers on empty response path → fixed (`map[string]string{}`)
- Missing `CharsetReader` → fixed with `charset.NewReaderLabel`
- Missing non-base64 empty response test → added
- Missing comment on `base64("")` → added as XML comment in test

### Backend

No findings. The empty response early return at `burp.go:122-135` mirrors the normal path cleanly, and initializing `Headers` to an empty map avoids nil-pointer inconsistency.

### Security

No findings. The removal of the `<!ENTITY>` pre-scan is sound:
- `encoding/xml` does not resolve external entities
- `decoder.Strict = true` rejects malformed XML
- `decoder.CharsetReader = charset.NewReaderLabel` handles non-UTF-8 encodings safely

### Summary

No critical, high, medium, or low findings. All prior review findings resolved. Tests pass (77/77). 

PR #34, opened 2026-03-12
feat: add --dangerous-allow-private flag to disable SSRF protection for local testing

eli-wald:
## Summary

- Adds a `--dangerous-allow-private` flag to the `generate` and `scan` commands that disables SSRF protection on probes, allowing them to target localhost and private/internal IP addresses
- When enabled, the probe config's `URLValidator` is replaced with a no-op function that permits all URLs
- A warning is printed to stderr whenever the flag is active to make it clear that SSRF protection is disabled

## Why

During local development and testing (e.g., targeting DVWA or other apps running on localhost), the default SSRF protection rejects requests to private IPs. This flag provides an explicit opt-in escape hatch for those scenarios. The `dangerous-` prefix and stderr warning make it clear this should not be used against production systems.

## Test plan

- [x] `TestDangerousAllowPrivate_GenerateSpec` — verifies `generateSpec` succeeds with `allowPrivate=true` for both probe=true and probe=false
- [x] `TestDangerousAllowPrivate_GenerateCmd` — creates a capture file and runs `GenerateCmd.Run()` with `DangerousAllowPrivate: true` and `Probe: true`
- [x] `TestDangerousAllowPrivate_ScanCmd` — verifies `ScanCmd` struct accepts and stores the `DangerousAllowPrivate` field
- [x] `TestDangerousAllowPrivate_SameOutputForPublicURLs` — confirms identical output for public URLs regardless of the flag value
- [x] Full test suite passes (`go test ./...`)


santiago-praetorian (changes requested):
## PR Review: `--dangerous-allow-private` flag

Well-structured and cleanly scoped change. The `dangerous-` prefix, stderr warning, and constructor migration are all good. However, there is a critical functional bug that prevents the feature from working as intended.

---

### HIGH: Incomplete SSRF bypass — flag does not actually allow private IPs

The SSRF protection has two layers:
1. **URLValidator** (pre-request) — validates URL and checks resolved IPs against private blocklist
2. **ssrfSafeDialContext** (connect-time) — re-resolves DNS at dial time and blocks private IPs (anti-rebinding)

The PR disables Layer 1 (`cfg.URLValidator = func(string) error { return nil }`) but Layer 2 remains active because `cfg.Client` is never set, so `withDefaults()` creates a client with `ssrfSafeDialContext` as the transport's `DialContext` (`pkg/probe/types.go:66-78`).

**Result**: `--dangerous-allow-private` silently fails. A user targeting `http://localhost:3000/api` will get a cryptic dial error from `ssrfSafeDialContext` rather than a successful probe.

**Fix**: When `allowPrivate` is true, also provide a custom `cfg.Client` with a standard dialer:

```go
if allowPrivate {
    cfg.URLValidator = func(string) error { return nil }
    cfg.Client = &http.Client{
        CheckRedirect: func(req *http.Request, via []*http.Request) error {
            return http.ErrUseLastResponse
        },
        Transport: &http.Transport{
            TLSHandshakeTimeout:   10 * time.Second,
            ResponseHeaderTimeout: 10 * time.Second,
        },
    }
}
```

---

### MEDIUM: No test verifies private IP access actually works

All `DangerousAllowPrivate` tests use `https://example.com` (public URLs). An `httptest.NewServer` on loopback that confirms the probe reaches it when the flag is enabled would have caught the above bug immediately.

---

### MEDIUM: Warning printed even when probing is disabled

The SSRF warning at `main.go:354` fires before the `doProbe` check at `main.go:358`, so `--no-probe --dangerous-allow-private` prints a misleading warning. Gate it on `allowPrivate && doProbe`.

---

### MEDIUM: Three consecutive bool params in `generateSpec`

```go
func generateSpec(ctx, requests, apiType, confidence, doProbe, allowPrivate, verbose)
```

Three consecutive `bool` parameters are easy to transpose at call sites with no compiler help. Consider an options struct.

---

### LOW: `TestDangerousAllowPrivate_ScanCmd` only tests struct initialization

This test constructs a `ScanCmd` and asserts field values but never calls `Run()`. The compiler already ensures the field exists — consider either removing or adding a comment explaining why it's intentionally shallow.

### LOW: `TestDangerousAllowPrivate_SameOutputForPublicURLs` runs with `probe=false`

Both calls have probing disabled, so `allowPrivate` is inert in both paths. The test proves nothing about the feature. Run with `probe=true` to be meaningful.

---

### Positive notes
- Migration from `&probe.OptionsProbe{}` to `NewOptionsProbe(cfg)` fixes a pre-existing bug where probes bypassed `withDefaults()`
- Flag correctly scoped to probes only (crawling unaffected)
- `CrawlCmd` correctly does not expose the flag
- No injection risk in the no-op validator

santiago-praetorian:
[LAB-1461: Feature: Add --dangerous-allow-private flag to disable SSRF protections for local testing](https://linear.app/praetorianlabs/issue/LAB-1461/feature-add-dangerous-allow-private-flag-to-disable-ssrf-protections)

santiago-praetorian (approved):
All six findings from the previous review have been addressed:

- **SSRF bypass**: Both layers (URLValidator + ssrfSafeDialContext) are now disabled when the flag is set
- **Loopback test**: `TestDangerousAllowPrivate_PrivateIPProbe` validates probes reach a local httptest server
- **Warning gating**: Warning only prints when both `allowPrivate` and `probe` are true
- **Options struct**: `generateSpecOptions` replaces consecutive bool parameters
- **SameOutputForPublicURLs**: Now runs with `probe: true`
- **ScanCmd test**: Doc comment explains intentionally shallow scope

All tests pass. Code is cleanly scoped and well-tested.

PR #39, opened 2026-03-15
feat(test): add end-to-end live test suite

santiago-praetorian:
## Summary

Addresses [LAB-1250](https://linear.app/praetorianlabs/issue/LAB-1250/create-end-to-end-live-test-suite)

- Add `test/` directory with automated end-to-end live test infrastructure modeled after Hadrian's live test suite
- `setup-live-targets.sh` builds binaries, resolves port conflicts, starts services, writes config
- `run-live-tests.sh` crawls targets, generates specs, validates output, prints summary table
- Custom REST API target with core endpoints + edge case endpoints (response types, URL handling, crawl behavior, classifier, spec generation)
- Custom SOAP/WSDL service target with 3 operations
- Synthetic import fixtures (Burp XML, HAR, mitmproxy) with expected output files for exact comparison
- 17 test targets across 5 categories: Import, Generation, Crawl behavior, Classifier, Spec generation

## Test plan

- [x] All 17 tests pass locally (`./test/run-live-tests.sh`)
- [x] Offline tests work without Chrome (`--targets import-burp,import-har,...,classifier-edge,spec-edge --no-build`)
- [x] Setup/teardown lifecycle works (`--teardown` flag)
- [x] Go unit tests unaffected (`go test ./...`)
- [x] REST API and SOAP service compile and respond correctly
- [ ] Reviewer runs `./test/setup-live-targets.sh && ./test/run-live-tests.sh` on their machine


praetorian-peter-mueller:
@santiago-praetorian Can you convert this PR to a draft until #44 is merged? At that time, we'll rebase and run the end-to-end testing?

praetorian-peter-mueller:
Hi @santiago-praetorian — given we're holding off on GraphQL temporarily, I'm proceeding with the review.

Feedback: I ran the test suite and there were no issues with Vespasian.

```
  Test Summary
  TARGET                  STATUS    ENDPOINTS   EXPECTED   DURATION
  ----------------------  --------  ----------  ---------  --------
  rest-api                PASS      8           8          76s
  soap-service            PASS      3           3          45s
  import-burp             PASS      5           5          0s
  import-har              PASS      3           3          0s
  import-base64           PASS      2           2          0s
  import-mitmproxy        PASS      3           3          0s
  import-unicode          PASS      3           3          0s
  import-duplicates       PASS      2           2          0s
  import-malformed        PASS      0           0          0s
  import-empty            PASS      0           0          0s
  generate-rest           PASS      8           8          0s
  generate-wsdl           PASS      3           3          0s
  edge-cases              PASS      -           -          186s
  crawl-depth             PASS      -           -          165s
  crawl-unreachable       PASS      0           0          32s
  classifier-edge         PASS      -           -          0s
  spec-edge               PASS      -           -          0s

  Total: 17 passed, 0 failed, 0 skipped
```

I did notice that some of the shell scripts under `test/` were incompatible with the ancient version of bash that is preinstalled on my macOS laptop. We could beat these scripts up until they're sufficiently portable, or we could just assume that the user is running a version of bash that was shipped in the last 20 years. I vote the latter.

That being said, if you start getting many bug reports about mysterious syntax errors in the `test/*.sh` scripts, you might need to look into bash-3.2 portability.

praetorian-peter-mueller (dismissed):
LGTM approved

santiago-praetorian:
## Code Review Fixes

Pushed `439ba29` addressing all findings from code review.

### Blocking (6 fixes)
1. **Service startup failures now abort setup** — `setup-live-targets.sh` no longer silently continues when a service fails to start
2. **`curl -sf` → `curl -s`** — removed `-f` flag from 4 curl calls that could abort under `set -e`
3. **XML injection in SOAP fault** — added `xmlEscape()` using `xml.EscapeText` for untrusted header values
4. **Shell-interpolated Python paths** — replaced all `python3 -c "...open('$var')..."` with `sys.argv` + heredoc pattern; added `json_field`, `json_len`, `json_valid` helpers
5. **import-malformed tests actual broken formats** — now tests truncated XML + invalid JSON alongside sparse-data fixtures
6. **YAML key detection at root level** — `validate.sh` uses `^key\s*:` with `re.MULTILINE` instead of substring search

### Medium (4 fixes)
7. **REST API errors return JSON** — `http.Error()` → `writeJSON()` across 7 call sites
8. **SOAP unknown action returns 400** — was incorrectly returning 500; method-not-allowed now returns XML
9. **`.gitignore` covers `node_modules/`** — added `test/graphql-server/node_modules/`
10. **Summary table sorted alphabetically** — deterministic output across runs

### Low (3 fixes)
11. **DRY: Python JSON helpers** — 9 duplicated one-liners replaced with 3 shared functions
12. **`paths: {}` detection** — handles inline empty YAML mapping
13. **Path regex matching quotes** — requires `"..."`, `'...'`, or unquoted (no longer accepts `"/path':`)

### README updates
- Go version 1.24 → 1.25 (matches `go.mod`)
- `--targets` expanded from 7 to all 20, grouped by category
- Output tree + directory structure covers all 20 test targets
- Added Expected Results section
- Fixtures listing expanded to all 19 files

### Other
- GraphQL health check: `GET /graphql` → `GET /` (Apollo Server 4 doesn't serve GET on `/graphql`)
- Summary table column width: 22 → 26 chars (fits `generate-graphql-imports`)

santiago-praetorian:
## Add graphql-server test target (`a503ebb`)

The `test/graphql-server/` directory was missing from the PR. Added 8 source files:

- `server.js` — Apollo Server with queries, mutations, enums, unions, nested types
- `package.json` / `package-lock.json` — Node.js dependencies
- `reference-capture.json` — fixed capture for deterministic `generate-graphql` tests
- `expected-paths.json` — expected operations for validation
- `expected-spec.graphql` — expected SDL for byte-identical comparison
- `test-burp.xml` — Burp XML import test data for `generate-graphql-imports`
- `test-traffic.har` — HAR import test data for `generate-graphql-imports`

Runtime artifacts (`capture.json`, `test-capture.json`) excluded — those are generated by the test runner. `node_modules/` covered by `.gitignore` from previous commit.

logan-bayless:
## PR Review: feat(test): add end-to-end live test suite

Great work on this — the test infrastructure is well-architected: clean setup/teardown lifecycle, deterministic fixtures for offline testing, port auto-resolution, and solid coverage across REST, SOAP, and GraphQL. The shared validation library with embedded Python is a nice pattern.

I found 10 issues during review (two passes). I've fixed 9 locally on the branch and need your help on 1.

### Needs Author Action

**SOAP `reference-capture.json` has truncated response bodies** — All three SOAP response bodies only contain `<id>1</id>`, missing the `<name>` and `<email>` fields the server actually returns. The `generate-wsdl` deterministic test validates against incomplete type information. Could you regenerate `test/soap-service/reference-capture.json` from the running soap-service and then regenerate `test/soap-service/expected-spec.xml` via `vespasian generate wsdl`?

### Resolved — Key Finding(s)

**`set -e` aborts entire test suite on any validation failure** — `validate.sh` had 8 functions where `result=$(python3 ...)` would abort the script under `set -euo pipefail` when the Python script exited non-zero. Every `if [ $? -ne 0 ]` check after them was dead code — never reached on failure. The existing `validate_path_coverage` function had the correct pattern (`|| missing=1`); the other 8 didn't. Fixed with `|| rc=$?`.

### Other Fixes (applied)

| Finding | Severity | Fix |
|---------|----------|-----|
| Server-side XSS in GraphQL server (`"${id}"` in `<script>` blocks) | LOW | `JSON.stringify(id)` |
| Client-side XSS via `innerHTML` across all GraphQL HTML pages | LOW | Added `esc()` helper using `createTextNode`; wrapped all string data |
| `--help` listed 10 of 20 supported targets | LOW | Updated to match README |
| Logging functions duplicated across setup + runner scripts | HYGIENE | Extracted to `test/common.sh` |
| Hardcoded port 4000 in GraphQL server startup log | LOW | Dynamic `port` variable |
| GraphQL server stderr sent to `/dev/null` (hides startup errors) | LOW | Redirect to `.graphql-server.log` |
| SOAP responses not XML-escaped (inconsistent with fault escaping) | HYGIENE | Added `xmlEscape()` calls |
| Dead cleanup trap (`PIDS_TO_CLEANUP` never populated) | LOW | Removed dead code |

logan-bayless:
## Update: All 10 findings now resolved

After posting the review above, we built and ran the soap-service locally to regenerate the reference capture from real traffic. The response bodies now include the full `<name>` and `<email>` fields.

The WSDL generator produces an identical `expected-spec.xml` — it infers operations from SOAPAction headers, not response bodies — so the expected spec is unchanged. But the reference capture is now a faithful recording of actual server output rather than a synthetic stub.

No action needed from the author. All 10 findings are fixed on the branch and ready for commit.

logan-bayless (dismissed):
Approving with fixes applied. I've pushed commit 2ae4621 to this branch with all 10 review findings resolved — please review that commit yourself before merging to make sure the changes look right to you. The review comments above have the full breakdown.

santiago-praetorian:
## Review Fixes (28ce3b4)

Pushed fixes for 6 medium-severity findings from backend + security review:

1. **Add `sync.Mutex` to REST API handlers** — `users`, `products`, `orders` slices are now protected against concurrent access from the crawler
2. **Convert 3 remaining `python3 -c` blocks to heredoc+`sys.argv`** — SOAP capture builder, `test_classifier_edge_cases`, and `test_spec_edge_cases` no longer interpolate `${port}` into Python code strings
3. **Replace `eval` with `declare -g` in config loader** — defense-in-depth against command execution via crafted config values
4. **Fix dead `$?` check in `test_graphql_server`** — `set -e` would abort before the `if [ $? -ne 0 ]` block; now uses `|| rc=$?` pattern consistent with `validate.sh`
5. **Narrow `pkill -f` fallback pattern** — includes `SCRIPT_DIR` in match to avoid killing unrelated processes on shared machines
6. **Freeze GraphQL seed data** — `Object.freeze()` on seed arrays; mutations operate on deep-copied working arrays to prevent cross-test state pollution

Also added `timeout=30` to `urllib.request.urlopen` in the GraphQL capture builder.

logan-bayless:
## PR Review (Round 2): feat(test): add end-to-end live test suite

Found 5 additional issues, fixed 4 in commit 7d4b0a8.

### Unfixed: method data in expected-paths.json is dead (LOW)

`expected-paths.json` stores HTTP methods for each path (`"/api/users": ["GET"]`), but `validate_path_coverage` only iterates the dict keys — the method arrays are never read. A regression that drops POST from a path would pass validation silently.

**Why not fixed here:** The fix is either (a) strip the method arrays from all `expected-paths.json` files and simplify the validation, or (b) implement method-level validation in `validate_path_coverage`. Both change the fixture contract and validation behavior — better done as a targeted follow-up than bolted onto a test infrastructure PR.

### Findings

| # | Title | Severity | Status |
|---|-------|----------|--------|
| 1 | XSS: `JSON.stringify(id)` in `<script>` tags allows `</script>` breakout in `userPage`/`postPage` | LOW | Fixed |
| 2 | `set -e` dead `$?` check in `test_graphql_server` (line 662) — introspection failure aborts test instead of soft warning | MEDIUM | Fixed |
| 3 | HTTP method values in `expected-paths.json` never validated by `validate_path_coverage` | LOW | Not fixed |
| 4 | `python3 -c` with double-quoted string in `test_crawl_unreachable` — inconsistent with heredoc pattern | HYGIENE | Fixed |
| 5 | Dead code: `users` seed data tags spread (SEED_USERS has no `tags` property) | HYGIENE | Fixed |

### Details

**Finding 2 (MEDIUM)** is the most impactful. Under `set -euo pipefail`, the command substitution `introspection_check=$(python3 ...)` aborts the script when python3 exits non-zero. The `if [ $? -ne 0 ]` check after it was unreachable — the introspection check was intended as a soft warning but was actually a hard abort. Fixed with `|| rc=$?` to match the pattern used consistently throughout `validate.sh`.

### Accountability: our initial review missed these AND Santiago's 6 findings

Our initial review (commit `2ae4621`) approved this PR after finding 10 issues. Santiago then found 6 more issues in commit `28ce3b4` — mutex gaps, shell injection, eval in config loader, another dead `$?` check, broad `pkill` pattern, and mutable seed data — that we should have caught. All 5 findings in this round were also present when we approved.

The root cause is that our first review read all 41 files but analyzed them superficially. We scanned for keyword-level patterns instead of tracing how the code actually executes. Concrete failures:

- **We never traced `set -e` behavior.** `validate.sh` uses `|| rc=$?` across 9 call sites. We read both files but never checked whether `run-live-tests.sh` followed the same pattern — it didn't.
- **We never audited pattern consistency.** Santiago's mutex finding existed because some handlers had `mu.Lock()` and others didn't. A grep for `mu.Lock` would have revealed the gap instantly.
- **We never traced variable flow to interpreter sinks.** Santiago found 3 remaining `python3 -c` shell injection sites that were in functions we read.

We've updated our review process to require execution path tracing, grep-verified pattern audits, and a written analysis state file before triage. This second pass used that process. We're posting this to be transparent about the gap — not to excuse it.

All Go tests pass, JS and shell syntax verified.

santiago-praetorian:
I have used Claude to review the code, and no changes were required. I have also run the tests and they all passed.

logan-bayless (approved):
Approving with round 2 fixes applied in commit 7d4b0a8. One LOW finding (unused method data in expected-paths.json) deferred as a follow-up — see review comment for details.

PR #43, opened 2026-03-18
feat: add --deduplicate flag to generate and scan commands

blayne:
## Description

Allow disabling deduplication so all classified requests are preserved. This is needed for GraphQL where all requests share the same METHOD:path key, causing the deduplicator to collapse them into a single entry.

## Changes

- Added Deduplicate bool field to both GenerateCmd and ScanCmd structs, defaulting to true, with CLI help text
- Updated generateSpec function signature to accept a new deduplicate bool parameter
- Made deduplication conditional in generateSpec — classify.Deduplicate() is now only called when deduplicate is true, instead of always running
- Passed c.Deduplicate through from both GenerateCmd.Run() and ScanCmd.Run() to generateSpec()
- Updated test call sites in TestGenerateSpec and TestGenerateSpec_EmptyRequests to pass true for the new deduplicate parameter
- Re-aligned struct field tags in GenerateCmd and ScanCmd to accommodate the new longer field name

The motivation is to support GraphQL APIs where all requests share the same METHOD:path key, causing the deduplicator to incorrectly collapse distinct operations into a single entry.

## Testing

- [X] Unit tests pass (`make test`)
- [X] Linter passes (`make lint`)
- [X] Manual verification (describe below)

Deduplication works well for GraphQL SDL generation.

## Checklist

- [ ] Code follows project conventions
- [ ] Tests added/updated for new functionality
- [ ] Documentation updated (if applicable)

### DIFF
--- a/cmd/vespasian/main.go
+++ b/cmd/vespasian/main.go
@@ -466,12 +466,13 @@ func (c *ImportCmd) Run() error {
 
 // GenerateCmd generates API specifications from captured traffic.
 type GenerateCmd struct {
-	APIType    string  `arg:"" enum:"rest,wsdl" help:"API type to generate"`
-	Capture    string  `arg:"" help:"Capture file path"`
-	Output     string  `short:"o" help:"Output file path"`
-	Confidence float64 `default:"0.5" help:"Minimum confidence threshold"`
-	Probe      bool    `default:"true" help:"Enable endpoint probing"`
-	Verbose    bool    `short:"v" help:"Enable verbose logging"`
+	APIType     string  `arg:"" enum:"rest,wsdl" help:"API type to generate"`
+	Capture     string  `arg:"" help:"Capture file path"`
+	Output      string  `short:"o" help:"Output file path"`
+	Confidence  float64 `default:"0.5" help:"Minimum confidence threshold"`
+	Probe       bool    `default:"true" help:"Enable endpoint probing"`
+	Deduplicate bool    `default:"true" help:"Deduplicate classified endpoints before probing (default: true)"`
+	Verbose     bool    `short:"v" help:"Enable verbose logging"`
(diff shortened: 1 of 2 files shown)


santiago-praetorian (changes requested):
## Review Findings

### Low: No test case exercises `deduplicate=false`

`cmd/vespasian/main_test.go:475,496` — Both `TestGenerateSpec` and `TestGenerateSpec_EmptyRequests` pass `true` for the `deduplicate` parameter. The `deduplicate=false` branch is never exercised by a test. Adding a table-driven case with `deduplicate: false` in `TestGenerateSpec` would confirm the conditional works as expected.

### Low: `TestGenerateCmdRun_ValidCapture` does not set `Deduplicate` field

`cmd/vespasian/main_test.go:534-538` — The `GenerateCmd` struct is constructed without setting `Deduplicate`. Go zero-values `bool` to `false`, so this test runs with deduplication disabled — the opposite of the CLI default (`true`). Setting `Deduplicate: true` explicitly would make the test intent clear and match the CLI default behavior.

### Nit: Redundant default in help text

`cmd/vespasian/main.go:474,530` — The help string `"Deduplicate classified endpoints before probing (default: true)"` manually states the default, but Kong already renders default values from the `default` struct tag. This results in the help output showing the default twice. Remove `(default: true)` from the help text.

santiago-praetorian:
[LAB-220](https://linear.app/praetorianlabs/issue/LAB-220/implement-graphql-classifier-and-sdl-generator-introspection-inference)

blayne:
@santiago-praetorian PLMK if the additional commit does not address your comments.

santiago-praetorian (approved):
All previous review findings have been addressed. Clean, minimal change — LGTM.

PR #44, opened 2026-03-18
feat: graphql support

blayne:
### GraphQL Support

These changes add end-to-end GraphQL API detection, probing, and spec generation to Vespasian, along with supporting infrastructure improvements.

### Core Features

- GraphQL Classifier (pkg/classify/graphql.go) — Dedicated heuristic classifier that identifies GraphQL traffic using layered signals: URL path matching (/graphql), request body analysis (detecting
query/mutation/subscription syntax), and response structure inspection (data/errors top-level keys). Produces confidence scores from 0.70 (path only) to 0.95+ (multiple signals).
- GraphQL Probe (pkg/probe/graphql.go) — Introspection-based schema discovery probe that sends a standard GraphQL introspection query to classified endpoints and parses the full type system response
into structured schema data.
- GraphQL Generator (pkg/generate/graphql/) — Two-path SDL spec generation: prefers introspection results when available, falls back to traffic-based inference. The inference engine (infer.go) uses
gqlparser for proper AST-level query parsing to extract operation types, field names, arguments (with declared variable types), and return types from observed request/response pairs. The SDL emitter
(sdl.go) renders the introspection schema into canonical GraphQL SDL.

### CLI Integration

- Added graphql as a recognized API type in the generate and scan commands.
- Added --deduplicate flag to the generate and scan commands.
- Updated help text to list all available --api-type options.

### Dependencies

- Upgraded katana from v1.4.0 to v1.5.0.
- Added github.com/vektah/gqlparser/v2 for GraphQL query parsing.

### Lint & Quality

- Fixed gofmt, goimports, errcheck, gosec, and gocritic warnings across all new GraphQL files.
- Reduced cyclomatic complexity in ClassifyDetail, inferSDL, and inferFieldsFromResponse by extracting helper functions.

## Changes

- Add GraphQLClassifier with multi-signal heuristic detection (path, body syntax, response structure) and confidence scoring
- Add GraphQLProbe for introspection-based schema discovery against classified endpoints
- Add GraphQLGenerator with two-path SDL output: full introspection schema rendering, or traffic-based inference using gqlparser AST parsing
- Add graphql as a recognized --api-type in generate and scan CLI commands
- Add --deduplicate flag to generate and scan commands
- List available API type options in generate --help text
- Upgrade katana dependency from v1.4.0 to v1.5.0
- Add gqlparser/v2 dependency for GraphQL query parsing
- Fix lint issues (gofmt, goimports, errcheck, gosec, gocritic) across all new GraphQL files
[...]

PR #45, opened 2026-03-22
fix: route SOAP/WSDL traffic to WSDL generator in scan command (LAB-1392)

santiago-praetorian:
## Summary

- The `scan` command was hardcoded to `APIType: "rest"`, causing SOAP/WSDL traffic to be misrouted to the REST classifier/generator with empty output
- Added `detectAPIType()` that runs the WSDL classifier against captured traffic before invoking `generateSpec()`
- If any request matches the WSDL classifier above the confidence threshold, the pipeline routes to the WSDL generator; otherwise falls back to REST
- Added `--api-type` flag (`auto|rest|wsdl|graphql`, default `auto`) so users can override auto-detection
- Introduced `apiType*` constants for compile-time safety across the routing pipeline
- Added `graphql` to the enum in preparation for PR #44 (feat/graphql-support)

## Changes

- `cmd/vespasian/main.go`:
  - Added `apiTypeAuto`/`apiTypeREST`/`apiTypeWSDL`/`apiTypeGraphQL` constants
  - Added `--api-type` flag to `ScanCmd` (default `auto`)
  - Added `detectAPIType()` function for WSDL auto-detection
  - Added `apiTypeDisplayName()` for verbose output
  - Updated `ScanCmd.Run()` to auto-detect API type instead of hardcoding `"rest"`
  - Added `graphql` to `GenerateCmd` and `ScanCmd` enum tags
- `cmd/vespasian/main_test.go`: Added 11 test cases covering detection logic, WSDL generation pipeline, explicit override, display names, and end-to-end scan regression

## Test plan

- [x] `TestDetectAPIType` — 7 subtests covering empty, REST, SOAP/SOAPAction, WSDL URL, SOAP envelope, mixed traffic, threshold
- [x] `TestGenerateSpec_WSDLType` — end-to-end WSDL generation with negative OpenAPI assertion
- [x] `TestScanPipeline_WSDLDetection` — full pipeline regression test for LAB-1392
- [x] `TestDetectAPIType_ExplicitOverride` — `--api-type` flag bypasses auto-detection
- [x] `TestAPITypeDisplayName` — display name mapping (REST, WSDL, GraphQL, unknown)
- [x] All existing tests pass (no regressions)
- [ ] Manual: `vespasian scan` against SOAP endpoint produces WSDL output
- [x] Manual: `vespasian scan` against REST endpoint still produces OpenAPI output

Fixes: LAB-1392


santiago-praetorian:
## Merge strategy with PR #44 (feat/graphql-support)

This PR adds `graphql` to the `--api-type` enum and the auto-detection framework. PR #44 adds the actual GraphQL classifier, probe, and generator. Depending on merge order:

- **If PR #44 merges first**: This PR needs to resolve the enum tag conflict (both already have `graphql`) and add a `GraphQLClassifier` check to `detectAPIType`.
- **If this PR merges first**: PR #44 gets the auto-detection framework for free and just needs to add a `classifiersForType` case + `detectAPIType` GraphQL check.

eli-wald:
Attempting to validate patch. Running `vespasian scan` (as well as `vespasian scan --api-type wsdl`) on SOAP services does not produce any output documents. Observed this when pointing at SOAP API endpoints as well as when pointing scan directly to wsdl file.

eli-wald (commented):
# PR #45 Review: fix: route SOAP/WSDL traffic to WSDL generator in scan command (LAB-1392)

**Reviewer:** Eli Wald
**Date:** 2026-03-23
**Verdict:** Request Changes

---

## Summary

This PR fixes the scan command's hardcoded `"rest"` API type by adding `detectAPIType()` to auto-detect WSDL traffic, and adds an `--api-type` flag for manual override. The approach is sound and well-structured, but the core bug (LAB-1392) is not actually fixed — manual testing confirms that `vespasian scan` and `vespasian scan --api-type wsdl` against SOAP endpoints still produces no output, and scans against WSDL documents end with `"generate failed: no SOAP operations found in traffic"`.

---

## CRITICAL: WSDL generation pipeline still fails end-to-end

### Problem

The routing fix is correct — `detectAPIType()` properly identifies SOAP traffic and routes it to the WSDL generator. However, the **WSDL generator itself** fails when processing crawl-captured SOAP traffic. Running:

```bash
vespasian scan [--api-type wsdl] <soap-endpoint> 
```

produces no output document. Running:

```bash
vespasian scan [--api-type wsdl]  <WSDL-document>
```

against WSDL documents hosted on SOAP endpoints produces the error:

```
vespasian: error: generate failed: no SOAP operations found in traffic
```

**Root cause:** The WSDL generator's `InferWSDL()` function (in `pkg/generate/wsdl/infer.go`) extracts SOAP operations from either the `SOAPAction` header or the first child element of the SOAP body. Crawl-captured requests typically have:

1. **Empty request bodies** — the crawler observes responses but doesn't replay SOAP requests, so `Body` is empty/nil
2. **No SOAPAction header** — many SOAP services don't require it, and the crawler doesn't synthesize one
3. **No WSDLDocument** — the probe strategy may fail or not be enabled, and the crawler doesn't fetch `?wsdl` as a request body

When all extraction methods fail for every classified request, `InferWSDL()` returns the "no SOAP operations found" error.

This means the PR's routing fix is necessary but not sufficient. The `TestGenerateSpec_WSDLType` and `TestScanPipeline_WSDLDetection` tests pass because they construct synthetic `ObservedRequest` objects with pre-populated SOAP bodies — traffic that real crawls don't produce in this form.

### Suggested fix

The fix likely needs to happen in the WSDL probe strategy or the crawler itself:

- **Option A:** When `--api-type wsdl` or auto-detection selects WSDL, the probe strategy should attempt to fetch the WSDL document from `<url>?wsdl` and populate `WSDLDocument`, which the generator's Phase 1 returns directly.
- **Option B:** Teach `InferWSDL()` to also inspect **response** bodies (not just request bodies) — crawled SOAP responses contain the envelope structure that reveals operation names.
- **Option C:** Ensure the WSDL probe (`NewWSDLProbe`) is wired up correctly and actually reaches the target (especially on localhost targets where SSRF protection may block it without `--dangerous-allow-private`).

This is a blocking issue — the PR description's manual test checkbox `[ ] Manual: vespasian scan against SOAP endpoint produces WSDL output` is unchecked, confirming this hasn't been validated.

---

## HIGH: Mixed REST+SOAP traffic always routes to WSDL

`detectAPIType()` returns `"wsdl"` on the **first** WSDL match, with no consideration for what fraction of traffic is SOAP vs REST:

```go
func detectAPIType(requests []crawl.ObservedRequest, threshold float64) string {
    wsdlClassifier := &classify.WSDLClassifier{}
    for _, req := range requests {
        if isAPI, confidence := wsdlClassifier.Classify(req); isAPI && confidence >= threshold {
            return apiTypeWSDL
        }
    }
    return apiTypeREST
}
```

If a target has 99 REST endpoints and 1 XML endpoint that happens to use `text/xml` content-type (confidence 0.85), the entire scan routes to the WSDL generator and all REST endpoints are lost.

The "mixed traffic" test case explicitly tests and asserts this behavior, but the test name ("mixed traffic with SOAP present returns wsdl") treats it as correct when it's actually a data loss bug. A scan against a real-world application with a `/health` endpoint returning XML would lose all REST API discovery.

**Suggestion:** Count WSDL vs REST matches and route based on the majority, or generate both specs (as future work). At minimum, emit a warning in verbose mode when mixed traffic is detected and WSDL is chosen despite being the minority.

---

## MEDIUM: Test assertions don't match real-world failure mode

`TestGenerateSpec_WSDLType` and `TestScanPipeline_WSDLDetection` construct synthetic SOAP traffic with:
- Pre-populated `SOAPAction` headers
- Complete SOAP envelope bodies

These tests pass because the hand-crafted traffic provides everything `InferWSDL()` needs. But real crawl output lacks request bodies and often lacks `SOAPAction` headers, so these tests don't exercise the actual failure path. They verify the *routing* is correct (good) but falsely suggest the *generation* works end-to-end (misleading).

**Suggestion:** Add a test with minimal SOAP signals (e.g., just `Content-Type: text/xml` and an empty body) to demonstrate that `InferWSDL` fails, confirming the known limitation. This prevents future developers from assuming the pipeline is fully functional.


---

## Positive notes

- The `detectAPIType()` approach is clean — lightweight first-pass classification with early return is the right pattern
- The `--api-type` flag with `auto` default is good UX for power users
- The dual-pass comment (detectAPIType vs generateSpec's RunClassifiers) is helpful
- API type constants improve compile-time safety
- `apiTypeDisplayName()` is a nice touch for verbose output
- Test coverage for the routing logic itself is thorough (7 subtests for detectAPIType)
- The PR correctly does not modify the WSDL classifier or generator, keeping scope focused on the routing layer

---

## Action items

1. **[Blocking]** Validate that the WSDL generation pipeline actually produces output for real SOAP traffic (not just synthetic test data). If it doesn't (which manual testing confirms), either fix the generation pipeline in this PR or split into two PRs: one for routing (this) and one for generation, with this PR's description clearly stating the limitation.
2. **[High]** Address the mixed-traffic data loss issue in `detectAPIType()`.
3. **[Medium]** Add a test case with realistic (body-less) SOAP traffic to document the known limitation.

santiago-praetorian:
## Review Response: WSDL Generation Pipeline Fix

Addressed all three review findings plus a deeper root cause discovered during manual testing.

### Changes

**1. CRITICAL — WSDL generation now works with real crawl traffic**
`pkg/generate/wsdl/infer.go`: `extractOperation()` now has a third fallback — when both SOAPAction header and request body are empty (typical for crawl-captured traffic), it extracts the operation name from the **response body's** SOAP envelope. Response elements like `<GetUserResponse>` have the "Response" suffix stripped to recover the operation name.

**2. HIGH — Mixed traffic no longer always routes to WSDL**
`cmd/vespasian/main.go`: `detectAPIType()` now counts WSDL and REST classifier matches across all requests. WSDL only wins when it has matches AND they represent the majority. Previously, a single XML endpoint caused all REST discovery to be lost.

**3. Active WSDL discovery probe (root cause fix)**
`cmd/vespasian/main.go`: Added `probeWSDLDocument()` that actively fetches `<targetURL>?wsdl` during the scan pipeline's detection phase. SOAP services return HTML for browser GETs, so crawl traffic rarely contains WSDL signals. When the probe finds a valid WSDL document, it injects a synthetic request so the generator's Phase 1 returns it directly. Includes verbose logging for debugging.

**4. MEDIUM — Realistic crawl-traffic tests**
Added 11 new tests covering: response-body operation extraction, crawled `?wsdl` document passthrough, full scan pipeline with realistic traffic, WSDL discovery probe (valid/invalid/404), end-to-end discovery probe integration, mixed-traffic majority detection, and known-limitation documentation.

### Manual Test Results

| Test | Target | Result |
|------|--------|--------|
| Public SOAP scan | dneonline.com/calculator.asmx | 1206 bytes inferred WSDL (probe connection reset, fallback succeeded) |
| Local SOAP scan | localhost:9090 mock | 7880 bytes full WSDL via probe |
| Explicit --api-type wsdl | localhost:9090 mock | 7880 bytes full WSDL |
| REST scan | localhost:5001 VAmPI | 379 bytes OpenAPI 3.0.3 |
| REST no WSDL leakage | localhost:5001 VAmPI | OpenAPI only, no WSDL |
| Generate from capture | WSDL capture file | 7880 bytes full WSDL |

### Exported symbol

`pkg/probe/validate.go`: Added `ValidateProbeURL()` as a public wrapper around `validateProbeURL()` for SSRF validation from outside the probe package.

All 124 unit tests pass. No regressions.

eli-wald (approved):
## Previously flagged issues — status

### [CRITICAL — RESOLVED] WSDL generation pipeline still fails end-to-end

Fixed via two complementary approaches:

1. **`extractOperation()` in `pkg/generate/wsdl/infer.go`** now falls back to response bodies when request bodies are empty, stripping the `Response` suffix (e.g., `GetUserResponse` → `GetUser`). This handles crawl-captured traffic correctly.

2. **`probeWSDLDocument()` in `main.go`** actively fetches `<targetURL>?wsdl` and injects the document as a synthetic request, so the generator's Phase 1 returns it directly without needing to infer operations.

Both paths are well-tested (`TestGenerateSpec_WSDLFromResponseBody`, `TestGenerateSpec_WSDLFromCrawledWSDLDocument`, `TestScanPipeline_WSDLDiscoveryProbe`).

Further, manual testing on dummy endpoints (localhost:8081 - https://github.com/mukkul007/Vulnerable-Web-Services and dvws.local - https://github.com/snoopysecurity/dvws-node) and public SOAP demo server (dneonline.com/calculator.asmx) all succeed.

### [HIGH — RESOLVED] Mixed REST+SOAP traffic always routes to WSDL

`detectAPIType()` now counts both WSDL and REST matches and requires WSDL to be the majority (`wsdlCount > 0 && wsdlCount >= restCount`). The test "minority SOAP in mostly REST traffic returns rest" covers this directly.

### [MEDIUM — RESOLVED] Test assertions don't match real-world failure mode

`TestGenerateSpec_WSDLEmptyBodyNoResponse` explicitly documents the known limitation (empty body + no SOAP response → error). `TestScanPipeline_RealisticCrawlTraffic` exercises the real-world pattern.

PR #46, opened 2026-03-24
feat: add ASCII banner on startup

santiago-praetorian:
## Summary
- Add styled ASCII art banner (FIGlet slant font) displayed on stderr at startup with version info
- Add `--no-banner` CLI flag to suppress the banner for scripted/automated usage
- Follow the same pattern used in hadrian (PR #39)

## Test plan
- [x] `TestPrintBanner` — verifies banner contains company name and version
- [x] `TestPrintBannerTo` — verifies banner output via io.Writer
- [x] `TestNoBannerFlag` — verifies NoBanner defaults to false
- [x] `TestBannerSuppression` — verifies banner suppressed when flag is set
- [x] Manual: `vespasian version` shows banner + version
- [x] Manual: `vespasian --no-banner version` suppresses banner

Closes LAB-1485

### DIFF
--- a/cmd/vespasian/display.go
+++ b/cmd/vespasian/display.go
@@ -1,0 +1,51 @@
+// Copyright 2026 Praetorian Security, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+// ANSI escape codes for banner styling.
+const (
+	bannerColorRed   = "\033[31m"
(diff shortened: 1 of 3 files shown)


logan-bayless (approved):
## PR Review: APPROVED

Clean, well-structured PR. Follows the established hadrian pattern, outputs to stderr (won't interfere with piped output), provides a `--no-banner` escape hatch. All tests pass, `go vet` clean. No security concerns. Two non-blocking observations below.

logan-bayless (commented):
### Non-blocking: `TestBannerSuppression` tests local bool, not actual flag wiring

`TestBannerSuppression` (display_test.go:81-94) creates a local `noBanner := true` and tests that the banner doesn't print when the variable is true. This tests basic boolean logic rather than exercising the actual `CLI.NoBanner` field or kong parsing path.

Not a real risk given the simplicity of the flag — `TestNoBannerFlag` at least verifies the default. A nice-to-have improvement would be to set `CLI.NoBanner = true`, call the conditional, and verify suppression — then reset it.

---

### Non-blocking: `captureStderr` mutates global `os.Stderr` (not parallel-safe)

`captureStderr` (display_test.go:26-46) swaps `os.Stderr` with a pipe, which is a global mutation. If any tests using this helper are ever annotated with `t.Parallel()`, they'll race on the global.

No bug today since none of these tests use `t.Parallel()`. The `printBannerTo(w)` API already exists for clean testability — `TestPrintBannerTo` uses it correctly. `TestPrintBanner` is the only test that actually needs the stderr capture.

PR #47, opened 2026-03-24
fix: prevent data race on Katana's global CustomFieldsMap (LAB-1650)

santiago-praetorian:
## Summary

- Add early context check in `Crawl()` before Katana initialization to prevent goroutine leaks that cause data races on the unprotected global `CustomFieldsMap`
- Treat context cancellation with zero results as graceful shutdown in `doCrawl()` (previously required partial results)

## Root Cause

Katana's `output.CustomFieldsMap` is a global `map[string]CustomFieldConfig` without synchronization. Signal-path tests cancel the context before calling `Crawl()`, but `Crawl()` still launches `engine.Crawl()` which spawns parser goroutines. These goroutines leak past the test boundary and race with the next test's `NewCrawlerOptions()` → `loadCustomFields()` write.

## Test plan

- [x] `make test` (`go test -race ./...`) passes consistently
- [x] 15 consecutive runs with `-race -count=1` show zero failures
- [x] All existing tests pass including `TestDoCrawl_GracefulShutdown*` and `TestCrawl_SignalPath_*`

Fixes [LAB-1650](https://linear.app/praetorianlabs/issue/LAB-1650/fix-data-race-in-crawl-tests-caused-by-katanas-global-customfieldsmap)

### DIFF
--- a/cmd/vespasian/main.go
+++ b/cmd/vespasian/main.go
@@ -229,12 +229,14 @@ func doCrawl(ctx context.Context, stderr io.Writer, targetURL string, opts crawl
 	}
 
 	if err != nil {
-		if (errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)) && len(requests) > 0 {
-			noun := "results"
-			if len(requests) == 1 {
-				noun = "result"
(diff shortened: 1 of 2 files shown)


blayne:
LGTM

blayne (approved):
LGTM

PR #52, opened 2026-04-02
fix: interleave port resolution with service startup

praetorian-peter-mueller:
## Summary
- Port resolution for all three test services (rest-api, soap-service, graphql-server) happened upfront before any service started
- When default ports (8990-8992) were occupied, `find_available_port` resolved all three to the same port (e.g. 8993) since none had started yet
- The first service grabbed the port; the remaining two crashed with `EADDRINUSE`, causing GraphQL live tests to fail
- Fix: resolve each port immediately before starting that service, so subsequent checks see prior services as occupied

## Test plan
- [x] `run-live-tests.sh --targets graphql-server,generate-graphql,generate-graphql-imports` — all 3 pass
- [x] Full `run-live-tests.sh` suite to verify no regressions

### DIFF
--- a/test/setup-live-targets.sh
+++ b/test/setup-live-targets.sh
@@ -376,39 +376,41 @@ main() {
         exit 0
     fi
 
-    # ── Port resolution ──
-    log_header "Resolving Ports"
-
-    REST_API_PORT=$(find_available_port "$DEFAULT_REST_API_PORT")
-    if [ -z "$REST_API_PORT" ]; then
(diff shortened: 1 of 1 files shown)


santiago-praetorian (approved):
Looks good — the fix correctly interleaves port resolution with service startup so `find_available_port` sees prior services as occupied. Clean, minimal change that directly addresses the root cause.

I ran the end-to-end test cases and the results matched previous results.

PR #56, opened 2026-04-03
fix: close DNS rebinding SSRF gap in probeWSDLDocument

praetorian-peter-mueller:
## Summary
- `probeWSDLDocument` was the only probe path using a plain `http.Client` without `ssrfSafeDialContext`, creating a TOCTOU DNS rebinding window
- Export `SSRFSafeDialContext` from `pkg/probe` and wire it into `probeWSDLDocument`'s transport
- When `allowPrivate` is true, uses default transport (matching existing behavior)

## Test plan
- [x] `make check` passes (fmt, vet, lint, test)
- [x] Out-of-band WSDL probe testing against test SOAP service

### DIFF
--- a/cmd/vespasian/main.go
+++ b/cmd/vespasian/main.go
@@ -827,8 +827,15 @@ func probeWSDLDocument(targetURL string, allowPrivate bool, verbose bool) []byte
 		}
 	}
 
+	transport := &http.Transport{
+		DialContext: probe.SSRFSafeDialContext,
+	}
+	if allowPrivate {
+		transport = &http.Transport{}
+	}
 	client := &http.Client{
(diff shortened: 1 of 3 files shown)


santiago-praetorian (approved):
Clean fix — wires the existing `SSRFSafeDialContext` into `probeWSDLDocument`'s transport, closing the TOCTOU DNS rebinding window. Approach matches the pattern already in `pkg/probe/types.go`.

Ran end-to-end live WSDL probe testing and results matched previous runs.

Minor suggestion for a follow-up: consider adding transport-level timeouts (`TLSHandshakeTimeout`, `ResponseHeaderTimeout`) to the new transport to match the existing pattern in `types.go`.

PR #63, opened 2026-04-12
feat: JS API endpoint extraction for SPA crawling

santiago-praetorian:
## Summary

Addresses [LAB-1505](https://linear.app/praetorianlabs/issue/LAB-1505/vespasian-cannot-discover-api-endpoints-from-spa-javascript-bundles)
Related to [LAB-1368](https://linear.app/praetorianlabs/issue/LAB-1368/missing-feature-vespasian-cannot-gather-concatenated-urls)

Replaces #42 (closed after history squash).

- Add post-crawl JS extraction step that discovers API endpoints embedded in SPA JavaScript bundles
- Re-fetches truncated JS bodies (>1MB bundles) with a 10MB limit since crawl truncates at 1MB
- Extracts API paths via 3 strategies: quoted strings, template literals, full URLs
- Discovers service prefixes from JS concatenation patterns and crawl results (multi-service SPAs)
- Probes discovered paths with raw HTTP GET, filters out 404s from wrong prefix combinations
- Enables `ScrapeJSLuiceResponses` in Katana for improved SPA route extraction

**Result against OWASP crAPI**: 29 API endpoints discovered from a SPA that previously yielded 1 (`manifest.json`).

## Test plan

- [ ] All unit tests pass (`go test ./...` — 8 packages)
- [ ] 19 test cases covering: path extraction, template literals, full URLs, service prefix detection, prefix from crawl results, truncated body re-fetch, max endpoints, context cancellation, probe failure
- [ ] Tested against OWASP crAPI (29/35 endpoints discovered via headless crawl)
- [ ] Reviewer runs `bin/vespasian scan <target-spa> -v` against a SPA target


santiago-praetorian:
[LAB-1505](https://linear.app/praetorianlabs/issue/LAB-1505/vespasian-cannot-discover-api-endpoints-from-spa-javascript-bundles)

blayne (changes requested):
# PR Review

**PR:** https://github.com/praetorian-inc/vespasian/pull/63
**Linear:** https://linear.app/praetorianlabs/issue/LAB-1505 (related: LAB-1368)
**Head SHA:** d817b4d668f8a6011b8f59e20abed1c953bae16c
**Reviewed at:** 2026-04-30T22:20:47Z

|---|---|---|---|---|---|
| Count | 0 | 1 | 8 | 20 | 1 |
| Blockers | 0 | 1 | 8 | 3 | 0 |

---

## Blockers

### QUAL-007 — medium — QUAL

**File:** `pkg/crawl/jsreplay.go:565-597`

**Issue:** The probing loop iterates over `allPaths` (a `map[string]bool`) with `for path := range allPaths`. Go map iteration order is randomized per process. When there are more discovered paths than MaxEndpoints, which paths get probed (and which get silently dropped) changes between runs of the same target. This makes the output non-reproducible: two runs of `vespasian scan` against the same SPA can produce different endpoint sets. For a security assessment tool, reproducibility is important to avoid missing endpoints in one run that would have been found in another.

**Recommendation:** Collect the map keys into a slice and sort it before iterating: `paths := make([]string, 0, len(allPaths)); for p := range allPaths { paths = append(paths, p) }; sort.Strings(paths); for _, path := range paths { ... }`. This ensures deterministic selection when MaxEndpoints is reached.

```
	allPaths := make(map[string]bool)
	...
	for path := range allPaths {  // non-deterministic iteration order
		if probed >= cfg.MaxEndpoints {
			break
		}
		...
```

---

### SEC-BE-002 — medium — SEC-BE

**File:** `pkg/crawl/jsreplay.go:226-658`

**Issue:** The fullURLPattern (lines 226-233) extracts arbitrary http(s)://host URLs from the SPA's JavaScript and the probe loop forwards them to probeURL, which sets every header from cfg.Headers (line 734-736) on the outbound request. cfg.Headers is populated from bs.opts.Headers (cmd/vespasian/main.go line 642), which comes from the user-supplied --header flag and is documented for cases such as 'Authorization: Bearer ...' or 'Cookie: session=...'. There is NO check that the destination host of an extracted full-URL matches the user's target host. A malicious or compromised SPA can therefore include a string like '"https://attacker.example.com/api/v1/sink"' in its bundle, causing Vespasian to send the operator's Authorization / Cookie / X-API-Key headers to attacker-controlled infrastructure during a scan — a credential exfiltration primitive. fetchJSBody has the same issue when re-fetching JS files at URLs derived from <script src="..."> tags pointing offsite (line 535 / 600).

**Recommendation:** Constrain header forwarding to the originally targeted host. Recommended approach: (a) capture the scan's target origin (scheme+host[+port]) once at the start of ReplayJSExtracted, (b) before sending headers in probeURL/fetchJSBody, compare the resolved request URL's host against the target origin and only send cfg.Headers when they match, (c) optionally accept an explicit allowlist of additional same-tenant hosts via JSReplayConfig. Additionally, log a warning when a discovered full URL points to a different origin than the target so the operator can audit cross-origin probing.

```
var fullURLPattern = regexp.MustCompile(
	`["'` + "`]" +
		`(https?://[a-zA-Z0-9._-]+(?::[0-9]+)?` +
for k, v := range cfg.Headers {
	req.Header.Set(k, v)
}
result = append(result, ObservedRequest{
	Method:   "GET",
	URL:      fullURL,
	Headers:  cfg.Headers,
	Response: *resp,
	Source:   "js-extract",
})
```

---

### SEC-BE-004 — high — SEC-BE

**File:** `pkg/crawl/jsreplay.go:500-661`

**Issue:** ReplayJSExtracted issues outbound HTTP requests (fetchJSBody and probeURL) to URLs that are derived from regex extraction of attacker-controlled JavaScript bundles (the target SPA). The full-URL extraction pattern (fullURLPattern at lines 226-233) and the script-src extractor (extractScriptURLs at line 174) accept ANY http://host or https://host taken from the JS body or HTML, so a malicious SPA can embed strings like 'http://169.254.169.254/api/v1/...', 'http://10.0.0.1/api/...', or 'http://attacker.com/api/...' and Vespasian will fetch them. The repo already has SSRF protection (pkg/probe/validate.go: ValidateProbeURL and SSRFSafeDialContext, which block RFC 1918, loopback, link-local including AWS metadata 169.254.169.254, and re-validate on connect to defeat DNS rebinding). The probe phase below this code wires AllowPrivate: c.DangerousAllowPrivate (cmd/vespasian/main.go line 652), but the new ReplayJSExtracted call (lines 641-645) does NOT receive that flag, does NOT import pkg/probe, and does NOT call ValidateProbeURL or use SSRFSafeDialContext for its http.Client.Transport. Risk: AWS/GCP metadata service exfiltration, internal-network port scanning, and exploitation of internal HTTP services from the Vespasian operator's machine — exactly the threat model the existing probe SSRF protection was built for.

**Recommendation:** (1) Plumb DangerousAllowPrivate into JSReplayConfig (e.g., add AllowPrivate bool field) and pass it from ScanCmd.Run(). (2) Before calling fetchJSBody and probeURL, call probe.ValidateProbeURL(url) (skip if AllowPrivate is true) and skip the URL on error. (3) Configure cfg.Client.Transport with a net.Dialer{ Control: ... } or DialContext: probe.SSRFSafeDialContext (when !AllowPrivate) to defeat DNS rebinding between the validation lookup and the actual TCP connect. (4) Apply the same validation to URLs returned by extractScriptURLs and to fullURL in the probe loop (line 632-634).

```
import (
	"bytes"
	"context"
	"fmt"
	"io"
	"net/http"
	"net/url"
	"regexp"
	"strings"
	"time"
)
resp, err := cfg.Client.Do(req) //nolint:gosec // G704: intentional outbound fetch of JS file
resp, err := cfg.Client.Do(req) //nolint:gosec // G704: intentional outbound probe of discovered endpoint
requests = crawl.ReplayJSExtracted(genCtx, requests, crawl.JSReplayConfig{
		Headers: bs.opts.Headers,
		Verbose: c.Verbose,
		Stderr:  os.Stderr,
	})
```

---

### SEC-BE-007 — medium — SEC-BE

**File:** `pkg/crawl/jsreplay.go:626-658`

**Issue:** Even setting aside SSRF and header exfiltration, the probe loop (lines 626-658) issues GETs to any full URL extracted from JavaScript regardless of whether it points to the user's target. Combined with MaxEndpoints=500 default, an attacker controlling the SPA can use Vespasian as a request reflector / scanner: embed 500 different attacker-chosen URLs in the JS bundle, each of which will be fetched from the operator's IP. This is useful for attribution attacks, log poisoning of third-party services, abuse-of-trust toward third-party APIs that whitelist the operator's IP, and consuming the operator's API quota. The existing probe pipeline operates only on classified endpoints derived from same-target traffic; this new path has no equivalent same-origin invariant.

**Recommendation:** Default to same-origin probing: only probe URLs whose host matches the scan target's host (or a configured allowlist). For full URLs extracted from JS that are off-origin, either skip them entirely or require an explicit --allow-cross-origin-js flag. This also reduces the SSRF and header-leak blast radius and aligns with the Vespasian threat model (probe the target, not the internet).

```
for path := range allPaths {
	if probed >= cfg.MaxEndpoints {
		break
	}
	fullURL := path
	if !strings.HasPrefix(path, "http://") && !strings.HasPrefix(path, "https://") {
		fullURL = baseURL + path
	}
	resp := probeURL(ctx, cfg, fullURL)
```

---

### TEST-001 — medium — TEST

**File:** `pkg/crawl/jsreplay.go:51-70`

**Issue:** withDefaults sets the documented default values for Timeout (10s), MaxEndpoints (500), Client (with `CheckRedirect: ErrUseLastResponse`), and Stderr (io.Discard). These are the contract the package documents in JSReplayConfig field comments, but there is no test that verifies the actual numeric/behavioral defaults. A regression that silently lowered MaxEndpoints to 50, removed the no-redirect policy, or changed Timeout to 1s would not be caught — every other test either supplies a Client explicitly or never inspects the resolved config. Coverage shows 90% of withDefaults statements executed but no direct assertion on the resolved values.

**Recommendation:** Add TestJSReplayConfig_WithDefaults that calls `JSReplayConfig{}.withDefaults()` and asserts: Timeout == 10*time.Second, MaxEndpoints == 500, Client != nil, Stderr == io.Discard, and that the resolved Client's CheckRedirect returns http.ErrUseLastResponse. Also add a case asserting that non-zero user values are preserved (e.g., MaxEndpoints: 7 stays 7).

```
func (cfg JSReplayConfig) withDefaults() JSReplayConfig {
	if cfg.Timeout == 0 {
		cfg.Timeout = 10 * time.Second
	}
	if cfg.MaxEndpoints == 0 {
		cfg.MaxEndpoints = 500
	}
```

---

### TEST-002 — low — TEST

**File:** `pkg/crawl/jsreplay.go:113-117`

**Issue:** extractScriptURLs returns nil when url.Parse(pageURL) fails, but no test passes an unparsable pageURL. Existing tests pass valid URLs only. The branch is small but defensive; without a test, a refactor that, for example, swallowed the parse error and used an empty base could silently produce incorrect resolved URLs.

**Recommendation:** Add a sub-case in TestExtractScriptURLs (or a new TestExtractScriptURLs_InvalidPageURL) that passes a URL string url.Parse rejects (e.g., a string with a control byte / scheme:// followed by an invalid host) and asserts the function returns nil.

```
func extractScriptURLs(htmlBody []byte, pageURL string) []string {
	base, err := url.Parse(pageURL)
	if err != nil {
		return nil
	}
```

---

### TEST-003 — low — TEST

**File:** `pkg/crawl/jsreplay.go:459-469`

**Issue:** ReplayJSExtracted scans requests to determine baseURL and returns the input unchanged if it cannot find any non-empty URL. There is no test for: (a) `requests == nil` or empty slice, (b) all requests with empty URL fields. TestReplayJSExtracted_NoJSFiles is close but supplies a request with a valid URL and HTML response — the baseURL discovery loop succeeds. A regression that changed early-return semantics (e.g., panicking on empty input or mutating the slice) would not be caught.

**Recommendation:** Add TestReplayJSExtracted_EmptyInput with `requests := []ObservedRequest{}` and assert `result` is an empty slice (or equal to input). Add a case where every request has URL == "" and assert the input is returned unchanged.

```
	baseURL := ""
	for _, req := range requests {
		if req.URL != "" {
			baseURL = resolveBaseURL(req.URL)
			break
		}
	}
	if baseURL == "" {
		return requests
	}
```

---

### TEST-004 — medium — TEST

**File:** `pkg/crawl/jsreplay.go:567-583`

**Issue:** The PR description's test plan claims '19 test cases covering: ... probe failure', but the test file does not exercise either of the two skip paths in ReplayJSExtracted: (1) probeURL returning nil (network error / unreachable host), and (2) the explicit `if resp.StatusCode == http.StatusNotFound { continue }` branch — which is the core mechanism the PR uses to filter out wrong service-prefix combinations (the central design claim in LAB-1505 and the PR body: 'Filters out 404 responses from wrong prefix combinations'). Neither behavior is asserted: no test gives a JS body whose extracted paths include both a 200 endpoint and a 404 endpoint and verifies that only the 200 is appended to result. Without this, a regression that started including 404s (polluting downstream classification with phantom endpoints) would not be caught.

**Recommendation:** Add TestReplayJSExtracted_Filters404 that wires up a handler returning 200/JSON for one path (e.g., /workshop/api/products) and 404 for another (e.g., /identity/api/products), feeds JS that produces both via prefix expansion, and asserts the result contains exactly the 200 endpoint and not the 404. Add TestReplayJSExtracted_ProbeNetworkError that points the client at a closed/refusing server (or a transport that returns an error) and asserts the result equals the original requests (probe failures are silently dropped).

```
		resp := probeURL(ctx, cfg, fullURL)
		if resp == nil {
			continue
		}

		// Skip 404 responses — these are typically wrong service prefix
		// combinations (e.g., /identity/api/shop/products when the correct
		// prefix is /workshop/). Keeping them would pollute the spec with
		// endpoints that don't actually exist on that service.
		if resp.StatusCode == http.StatusNotFound {
			continue
		}
```

---

### TEST-005 — low — TEST

**File:** `pkg/crawl/jsreplay.go:631-633`

**Issue:** fetchJSBody has an explicit `if resp.StatusCode >= 400 { return nil }` guard (jsreplay.go:631-633), but no test triggers it. The two existing fetch tests (TestFetchJSBody_RejectsHTML, TestFetchJSBody_RejectsHTMLWithoutContentType) both serve 200 responses with HTML bodies. A regression that removed the >=400 guard would still pass all current tests as long as the server returned a non-HTML body — for example, a 500 error page in JSON would suddenly start being parsed for API paths. Coverage report shows fetchJSBody at 73.9%.

**Recommendation:** Add TestFetchJSBody_RejectsErrorStatus with a server returning 404 and 500 (with non-HTML bodies) and assert fetchJSBody returns nil for both. This locks in the error-status guard independently of the HTML-content guard.

```
	if resp.StatusCode >= 400 {
		return nil
	}
```

---

### TEST-007 — medium — TEST

**File:** `pkg/crawl/jsreplay_test.go:388-415`

**Issue:** TestReplayJSExtracted_MaxEndpoints feeds 5 paths with MaxEndpoints=2 but asserts only `LessOrEqual(len(result), 3)` and `LessOrEqual(callCount, 2)`. Because the production loop iterates a `map[string]bool` (non-deterministic order) and skips probes whose response is nil or 404, the assertions pass even if the cap is silently broken or never reached (e.g., 0 probes still satisfies the bound). The test handler always returns 200/JSON, so the cap is the only thing that can stop probing — the assertions should be exact equalities (`Equal(t, 3, len(result))` and `Equal(t, 2, callCount)`). As written, a regression that increments past MaxEndpoints by one would still pass the upper-bound check on average runs, and a regression that never honors MaxEndpoints (0 probes) trivially passes both `LessOrEqual` checks.

**Recommendation:** Assert exact values: `assert.Equal(t, 3, len(result))` (1 original + exactly 2 probed) and `assert.Equal(t, 2, callCount)`. This converts the test from a one-sided bound into a precise behavioral contract on the MaxEndpoints cap.

```
	cfg := JSReplayConfig{Client: srv.Client(), MaxEndpoints: 2}
	result := ReplayJSExtracted(context.Background(), requests, cfg)

	assert.LessOrEqual(t, len(result), 3) // 1 original + max 2 probed
	assert.LessOrEqual(t, callCount, 2)
```

---

### REQ-001 — medium — REQ

**File:** `pkg/crawl/jsreplay.go:154-161`

**Issue:** templateLiteralPattern terminates the match at the first `${` interpolation, so `` `/api/users/${id}/profile` `` extracts only `/api/users`, not `/api/users/{id}/profile`. This is the dominant SPA URL-building idiom; in practice it means parameterized resource paths after the interpolation site are silently dropped. Combined with the LAB-1368 gap above, the PR's coverage of parameterized URL construction is narrower than the related ticket implies.

**Recommendation:** Either (a) extend the template-literal scan to continue past `${...}` and reassemble the literal segments into a path-template (e.g., turn `${id}` into `{id}`), or (b) document this limitation explicitly in the package comment / README and link a follow-up Linear issue. Add a regression test using a representative `\`/api/users/${id}/profile\`` literal asserting the reassembled output.

```
var templateLiteralPattern = regexp.MustCompile(
	"`" +
		`(/?` +
			`(?:[a-zA-Z0-9_-]+/)*` +
			`(?:api/|v[1-9][0-9]*/|rest/|rpc/|graphql)` +
			`[a-zA-Z0-9/_\{}.:-]*)` +
		"(?:`|\\$\\{)",
```

---

### REQ-002 — medium — REQ

**File:** `pkg/crawl/jsreplay.go:176-178`

**Issue:** LAB-1368's stated example — `"/api/posts/".concat(id, "/comment")` — would not be reconstructed into `/api/posts/{id}/comment`. The servicePrefixPattern only matches `+`-operator concatenation between two QUOTED string literals; method-call concatenation via `String.prototype.concat` is not matched. The literal `/api/posts/` is still picked up by apiPathPattern, so the prefix surfaces, but the parameterized suffix `/comment` is lost. Since LAB-1368 is the related ticket explicitly cited in the PR description, this is a partial alignment with the stated requirement.

**Recommendation:** Either (a) extend extraction to recognize the `.concat(...)` form (e.g., `"...".concat(...)`), capturing the literal arguments and reconstructing the parameterized path; (b) update the LAB-1368 ticket to scope it explicitly to `+` and template-literal forms and link a follow-up issue for `.concat`; or (c) add a comment in jsreplay.go calling out that `.concat` is intentionally out of scope and pointing at the follow-up ticket.

```
// servicePrefixPattern matches service prefix strings concatenated with API paths.
// E.g., "identity/" + "api/auth/login" — captures "identity/".
var servicePrefixPattern = regexp.MustCompile(
```

---

## Non-blocking

### QUAL-001 — low — QUAL

**File:** `pkg/crawl/doc.go`

**Issue:** The PR adds a new exported type (JSReplayConfig) and a new exported function (ReplayJSExtracted) to the crawl package, which is a significant addition to the public API. CLAUDE.md states: 'After modifying a Go source file, update its package's doc.go if the change affects the package's public API or purpose.' The doc.go currently lists only Crawler, BrowserManager, ObservedRequest, and ObservedResponse. The new JS extraction step is a distinct new pipeline stage that should be documented there.

**Recommendation:** Add a description of the JS extraction step to pkg/crawl/doc.go. For example, add to the Key types list: '- [JSReplayConfig] and [ReplayJSExtracted] implement a post-crawl JS bundle scanning step that extracts API paths from SPA JavaScript bundles and probes them with raw HTTP requests to discover endpoints missed by the headless browser.'

```
// Key types:
//   - [Crawler] orchestrates a headless browser crawl ...
//   - [BrowserManager] manages Chrome process lifecycle ...
//   - [ObservedRequest] and [ObservedResponse] represent captured HTTP traffic.
// (JSReplayConfig and ReplayJSExtracted are absent)
```

---

### QUAL-002 — low — QUAL

**File:** `pkg/crawl/jsreplay.go:95-106`

**Issue:** isHTMLResponse (lines 95-106) and isJSResponse (lines 189-200) are structurally identical: both lowercase the content-type, strip a charset suffix at ';', then loop over a slice of known types. The only difference is the slice iterated. This is 6+ lines of copy-paste logic that will need to be maintained in lockstep. If the parsing logic needs to change (e.g., handle quoted charset params or whitespace around the semicolon) it must be updated in two places.

**Recommendation:** Extract a shared helper: `func matchesContentType(contentType string, types []string) bool`. Both isHTMLResponse and isJSResponse become one-liners calling that helper. This also makes it trivial to add additional content-type groups in the future.

```
func isHTMLResponse(contentType string) bool {
	ct := strings.ToLower(contentType)
	if idx := strings.Index(ct, ";"); idx != -1 {
		ct = strings.TrimSpace(ct[:idx])
	}
	for _, htmlCT := range htmlContentTypes {
		if ct == htmlCT {
			return true
		}
	}
	return false
}
// ... identical block at lines 189-200 for isJSResponse
```

---

### QUAL-003 — low — QUAL

**File:** `pkg/crawl/jsreplay.go:203-211`

**Issue:** apiPathPattern matches any quoted string containing 'api/', 'v1/', 'rest/', 'rpc/', or 'graphql' with a preceding optional path prefix. The pattern does not anchor to word boundaries or require the string to stand alone. This means string literals used in comments, error messages, or locale strings within JS (e.g., `"Error: call /api/v1/auth failed"`, `"See docs at /v1/guide"`) will produce false-positive path matches that get probed. The 404-filter mitigates this at runtime, but these extra probes consume MaxEndpoints quota and generate noise in verbose output. This is a known tradeoff documented in the PR (wrong combinations filtered by classifier), but the impact on MaxEndpoints is undocumented.

**Recommendation:** This is an inherent tradeoff of regex-based extraction vs. AST parsing. At a minimum, add a comment near the pattern explaining that false positives are expected and that the 404 filter is the primary defence. Consider adding a minimum path-depth filter (e.g., at least 2 path segments after the API indicator) to reduce shallow false positives.

```
var apiPathPattern = regexp.MustCompile(
	`["']` +
		`(/?` +
		`(?:[a-zA-Z0-9_-]+/)*` +
		`(?:api/|v[1-9][0-9]*/|rest/|rpc/|graphql)` +
		`[a-zA-Z0-9/_\{}.:-]*)` +
	`["']`,
)
```

---

### QUAL-004 — nit — QUAL

**File:** `pkg/crawl/jsreplay.go:237-239`

**Issue:** The comment for servicePrefixPattern says 'E.g., "identity/" + "api/auth/login"' using double-quote examples, which matches the pattern. The pattern itself only matches single or double-quoted strings (not template literals/backticks). This is correct and consistent, but the comment could be more explicit that template literal concatenations (e.g., `identity/` + `api/...`) are NOT captured by this pattern.

**Recommendation:** Add a note to the comment: '// Note: backtick template literals are not matched; concatenation must use single or double quotes.'

```
// servicePrefixPattern matches service prefix strings concatenated with API paths.
// E.g., "identity/" + "api/auth/login" — captures "identity/".
var servicePrefixPattern = regexp.MustCompile(
	<regex literal — see pkg/crawl/jsreplay.go line 240 for verbatim>
```

---

### QUAL-005 — low — QUAL

**File:** `pkg/crawl/jsreplay.go:564-597`

**Issue:** JSReplayConfig.MaxEndpoints is documented as 'limits the number of URLs probed' (line 98) and its default is set in withDefaults. However, the counter (`probed`) is only incremented after a non-nil, non-404 response is appended to `result` (line 596). If many probed URLs return nil (network errors, timeouts) or 404, the loop will issue far more than MaxEndpoints actual HTTP requests. In a target environment with many unreachable or slow-to-timeout endpoints, this can cause the step to run significantly longer than expected. The field name and documentation imply a hard cap on outbound probes, not a cap on results.

**Recommendation:** Either: (a) rename the field to MaxResults and update the comment to clarify it limits the number of successful results appended, or (b) introduce a separate probe counter that increments on every probeURL call (not just successful ones) and break when that counter reaches MaxEndpoints. Option (b) makes the behaviour match the documented name.

```
	// MaxEndpoints limits the number of URLs probed. Defaults to 500.
	MaxEndpoints int

	probed := 0
	for path := range allPaths {
		if probed >= cfg.MaxEndpoints {
			break
		}
		...
		if resp.StatusCode == http.StatusNotFound {
			continue  // probed NOT incremented
		}
		result = append(result, ...)
		probed++  // only incremented on success
```

---

### QUAL-006 — low — QUAL

**File:** `pkg/crawl/jsreplay.go:564-598`

**Issue:** When more than MaxEndpoints paths are discovered, the loop breaks silently. The verbose logf before the loop reports the total count of unique API paths found ('js-extract: N unique API paths found across all JS files'), but there is no warning message when the cap is actually reached and paths are silently dropped. A user running without -v will have no indication that endpoint discovery was truncated, which could lead to missed findings in a security assessment.

**Recommendation:** After the probing loop completes, check if `probed >= cfg.MaxEndpoints && len(allPaths) > cfg.MaxEndpoints` and emit a warning (to cfg.Stderr, unconditionally — not only in verbose mode) such as: `fmt.Fprintf(cfg.Stderr, "js-extract: warning: probed %d/%d discovered paths (MaxEndpoints limit reached; increase with --max-js-endpoints)\n", probed, len(allPaths))`.

```
	logf("js-extract: %d unique API paths found across all JS files\n", len(allPaths))

	probed := 0
	for path := range allPaths {
		if probed >= cfg.MaxEndpoints {
			break  // silent truncation — no warning emitted
		}
		...
```

---

### QUAL-008 — low — QUAL

**File:** `pkg/crawl/jsreplay.go:606-697`

**Issue:** fetchJSBody (lines 606-649) and probeURL (lines 664-697) share an identical 10-line HTTP request setup pattern: context timeout, http.NewRequestWithContext, header injection loop, cfg.Client.Do with error check, and defer body drain+close. This is 3+ contiguous identical logical blocks repeated verbatim in two functions within the same file. If the header injection logic or drain pattern needs to change, both must be updated.

**Recommendation:** Extract a doRequest helper with signature similar to: `func doRequest(ctx context.Context, cfg JSReplayConfig, rawURL string) (*http.Response, context.CancelFunc, error)`. Both fetchJSBody and probeURL call this helper for the common setup, then apply their own response-specific logic (status-code filtering, HTML rejection, body limit) after the call.

```
// In fetchJSBody:
	reqCtx, cancel := context.WithTimeout(ctx, cfg.Timeout)
	defer cancel()
	req, err := http.NewRequestWithContext(reqCtx, http.MethodGet, rawURL, nil)
	for k, v := range cfg.Headers { req.Header.Set(k, v) }
	resp, err := cfg.Client.Do(req)
	defer func() { io.Copy(...); resp.Body.Close() }()
// Identical block appears in probeURL at lines 666-684
```

---

### SEC-BE-001 — low — SEC-BE

**File:** `pkg/crawl/jsreplay.go:95-131`

**Issue:** Per-request controls exist (cfg.Timeout default 10s, MaxEndpoints default 500, maxJSBodySize 10 MB, maxReplayBodySize 1 MB), but there is no overall deadline on ReplayJSExtracted. Worst-case sequential execution: 500 endpoints * 10s = ~83 minutes, plus per-JS-file fetches. A malicious SPA serving slow-loris responses or many large JS files can therefore keep the tool busy long after the user expects the scan to be done. Because requests are sequential, this is operator-side resource burn rather than a service DoS, but it makes scan time predictability dependent on attacker behavior.

**Recommendation:** Either (a) add an overall context.WithTimeout around the probe loop derived from a JSReplayConfig.MaxTotalTime field (e.g., min(MaxEndpoints * Timeout, configurable cap)), or (b) document the worst-case time bound clearly. If sequential semantics aren't required, a small bounded worker pool (3-5) with a shared deadline would be more efficient and naturally cap total wall-clock time.

```
if cfg.Timeout == 0 {
	cfg.Timeout = 10 * time.Second
}
if cfg.MaxEndpoints == 0 {
	cfg.MaxEndpoints = 500
}
```

---

### SEC-BE-003 — low — SEC-BE

**File:** `pkg/crawl/jsreplay.go:410-422`

**Issue:** addPath (lines 410-422) accepts a string raw and treats anything starting with 'http://' or 'https://' as a full URL, then passes it through to the probe loop. The fullURLPattern regex enforces basic shape but does not reject e.g. URL credentials ('http://user:pass@host/...'), IPv6 zone identifiers, or unusual ports (the pattern allows :[0-9]+). Combined with no host validation, this gives the JS bundle author considerable control over the requests Vespasian will issue. While Go's net/http rejects some malformed URLs at the request-construction step, the precondition that 'we only probe sane URLs' is not actually enforced here.

**Recommendation:** Parse each extracted full URL with url.Parse before adding it to allPaths and reject (a) URLs with non-empty u.User (credentials), (b) URLs whose Host is empty, (c) URLs with schemes other than http/https. This is a defense-in-depth measure that complements (not replaces) the SSRF / same-origin checks in the higher-severity findings.

```
addPath := func(raw string) {
	if isStaticFile(raw) {
		return
	}
	// Full URLs are kept as-is (they include scheme+host).
	if strings.HasPrefix(raw, "http://") || strings.HasPrefix(raw, "https://") {
		raw = strings.TrimRight(raw, "/")
		if !seen[raw] {
			seen[raw] = true
			paths = append(paths, raw)
		}
		return
	}
```

---

### SEC-BE-005 — low — SEC-BE

**File:** `pkg/crawl/jsreplay.go:540-615`

**Issue:** logf prints attacker-controlled strings (URLs extracted from JS, Content-Type values, JS file URLs from <script src>) directly to cfg.Stderr (default os.Stderr) when verbose mode is on. The strings are not sanitized for terminal control characters; a malicious SPA could embed ANSI escape sequences or NUL bytes inside a quoted string that satisfies one of the extraction patterns and have them rendered when the operator runs scan with -v. Impact is limited (operator's terminal only, mostly cosmetic clobbering) but worth flagging because the new code path materially expands the attack surface for what gets logged.

**Recommendation:** Before logging extracted strings, run them through strconv.Quote (or strip non-printable characters / 0x00-0x1F except \t\n) so terminal escape sequences cannot be injected via the JS bundle. Apply consistently to logf calls at lines 540-543, 553, 555, 567-568, 578, 580-581, 586, 591, 605, 608, 611, 615.

```
logf("js-extract: discovered %d JS URLs from HTML <script> tags\n", len(htmlJSURLs))
for u := range htmlJSURLs {
	logf("  %s\n", u)
}
logf("js-extract: extracted %d API paths from %s\n", len(paths), jsURL)
for _, p := range paths {
	logf("  %s\n", p)
}
```

---

### SEC-BE-006 — low — SEC-BE

**File:** `pkg/crawl/jsreplay.go:562-596`

**Issue:** When req.URL is identified as JS (by extension) or has Content-Type=application/javascript but the response body is empty (line 576-588), ReplayJSExtracted re-fetches that URL with cfg.Headers attached (fetchJSBody at line 667-710). The set of URLs Katana flags this way includes anything ending in .js or .mjs, which can include cross-origin CDN URLs (e.g., the test on line 1310 includes 'https://cdn.example.com/lib.js'). This means user-supplied auth headers flow to third-party CDNs and any host the SPA links a script tag to. Same root cause as finding 'User-supplied auth headers exfiltrated...' but reachable through a separate code path that does not require fullURLPattern matching.

**Recommendation:** Restrict fetchJSBody header forwarding (and re-fetching at all) to same-origin JS files relative to the scan target. Cross-origin script src='...' URLs from <script> tags should either be skipped or fetched without cfg.Headers. The same applies to URLs returned by extractScriptURLs (line 174) when they resolve to a different origin than the page URL.

```
if len(jsBody) == 0 || len(jsBody) >= MaxResponseBodySize {
	if len(jsBody) == 0 {
		logf("js-extract: empty body, fetching %s\n", req.URL)
	} else {
		logf("js-extract: body truncated at %d bytes, re-fetching %s\n",
			MaxResponseBodySize, req.URL)
	}
	fullBody := fetchJSBody(ctx, cfg, req.URL)
```

---

### SEC-BE-008 — low — SEC-BE

**File:** `pkg/crawl/jsreplay.go:650-656`

**Issue:** result is constructed with Headers: cfg.Headers (line 653). cfg.Headers is the same shared map referenced by the caller and by every appended ObservedRequest. If cfg.Headers contains an Authorization header, the resulting capture.json (when persisted) will contain that secret in every probed entry's Headers field. The classifier and writeOutput downstream may serialize this. This is consistent with how the existing crawl pipeline records request headers, but the new code intentionally copies a reference rather than a defensive copy, so any later mutation by the caller mutates already-observed records.

**Recommendation:** (1) Defensively copy cfg.Headers into a per-request map when building ObservedRequest. (2) Separately, consider redacting sensitive headers (Authorization, Cookie, X-API-Key, etc.) before persisting capture.json, or document that capture.json may contain credentials and should be treated as sensitive.

```
result = append(result, ObservedRequest{
	Method:   "GET",
	URL:      fullURL,
	Headers:  cfg.Headers,
	Response: *resp,
	Source:   "js-extract",
})
```

---

---

## Ticket alignment

### REQ-001 — medium

**File:** `pkg/crawl/jsreplay.go:154`

**Status:** partially implemented

**Evidence:** templateLiteralPattern terminates the match at the first `${` interpolation, so `` `/api/users/${id}/profile` `` extracts only `/api/users`, not `/api/users/{id}/profile`. This is the dominant SPA URL-building idiom; in practice it means parameterized resource paths after the interpolation site are silently dropped. Combined with the LAB-1368 gap above, the PR's coverage of parameterized URL construction is narrower than the related ticket implies.

**Recommendation:** Either (a) extend the template-literal scan to continue past `${...}` and reassemble the literal segments into a path-template (e.g., turn `${id}` into `{id}`), or (b) document this limitation explicitly in the package comment / README and link a follow-up Linear issue. Add a regression test using a representative `\`/api/users/${id}/profile\`` literal asserting the reassembled output.

---

### REQ-002 — medium

**File:** `pkg/crawl/jsreplay.go:176`

**Status:** partially implemented

**Evidence:** LAB-1368's stated example — `"/api/posts/".concat(id, "/comment")` — would not be reconstructed into `/api/posts/{id}/comment`. The servicePrefixPattern only matches `+`-operator concatenation between two QUOTED string literals; method-call concatenation via `String.prototype.concat` is not matched. The literal `/api/posts/` is still picked up by apiPathPattern, so the prefix surfaces, but the parameterized suffix `/comment` is lost. Since LAB-1368 is the related ticket explicitly cited in the PR description, this is a partial alignment with the stated requirement.

**Recommendation:** Either (a) extend extraction to recognize the `.concat(...)` form (e.g., `"...".concat(...)`), capturing the literal arguments and reconstructing the parameterized path; (b) update the LAB-1368 ticket to scope it explicitly to `+` and template-literal forms and link a follow-up issue for `.concat`; or (c) add a comment in jsreplay.go calling out that `.concat` is intentionally out of scope and pointing at the follow-up ticket.

---

### REQ-003 — low

**File:** `pkg/crawl/jsreplay_test.go`

**Status:** partially implemented

**Evidence:** Both LAB-1505 and the PR description state a concrete acceptance result: "29 API endpoints discovered" against OWASP crAPI from a SPA that previously yielded 1. Nothing in the PR encodes that target — there is no fixture (e.g., a captured copy of the crAPI JS bundle and a golden output of expected paths), no integration test, and no documented manual reproduction script. Future regressions in extraction logic will not be caught.

**Recommendation:** Capture a small slice of the crAPI JS bundle (or a redacted equivalent) as a test fixture and add an integration-style test that calls extractAPIPaths against it and asserts a stable set of path strings. Alternatively, document the manual reproduction command in CLAUDE.md or pkg/crawl/doc.go so a reviewer can verify the claim.

---

## Tests & documentation

### TEST-001 — medium — TEST

**File:** `pkg/crawl/jsreplay.go:51-70`

**Issue:** withDefaults sets the documented default values for Timeout (10s), MaxEndpoints (500), Client (with `CheckRedirect: ErrUseLastResponse`), and Stderr (io.Discard). These are the contract the package documents in JSReplayConfig field comments, but there is no test that verifies the actual numeric/behavioral defaults. A regression that silently lowered MaxEndpoints to 50, removed the no-redirect policy, or changed Timeout to 1s would not be caught — every other test either supplies a Client explicitly or never inspects the resolved config. Coverage shows 90% of withDefaults statements executed but no direct assertion on the resolved values.

**Recommendation:** Add TestJSReplayConfig_WithDefaults that calls `JSReplayConfig{}.withDefaults()` and asserts: Timeout == 10*time.Second, MaxEndpoints == 500, Client != nil, Stderr == io.Discard, and that the resolved Client's CheckRedirect returns http.ErrUseLastResponse. Also add a case asserting that non-zero user values are preserved (e.g., MaxEndpoints: 7 stays 7).

```
func (cfg JSReplayConfig) withDefaults() JSReplayConfig {
	if cfg.Timeout == 0 {
		cfg.Timeout = 10 * time.Second
	}
	if cfg.MaxEndpoints == 0 {
		cfg.MaxEndpoints = 500
	}
```

---

### TEST-002 — low — TEST

**File:** `pkg/crawl/jsreplay.go:113-117`

**Issue:** extractScriptURLs returns nil when url.Parse(pageURL) fails, but no test passes an unparsable pageURL. Existing tests pass valid URLs only. The branch is small but defensive; without a test, a refactor that, for example, swallowed the parse error and used an empty base could silently produce incorrect resolved URLs.

**Recommendation:** Add a sub-case in TestExtractScriptURLs (or a new TestExtractScriptURLs_InvalidPageURL) that passes a URL string url.Parse rejects (e.g., a string with a control byte / scheme:// followed by an invalid host) and asserts the function returns nil.

```
func extractScriptURLs(htmlBody []byte, pageURL string) []string {
	base, err := url.Parse(pageURL)
	if err != nil {
		return nil
	}
```

---

### TEST-003 — low — TEST

**File:** `pkg/crawl/jsreplay.go:459-469`

**Issue:** ReplayJSExtracted scans requests to determine baseURL and returns the input unchanged if it cannot find any non-empty URL. There is no test for: (a) `requests == nil` or empty slice, (b) all requests with empty URL fields. TestReplayJSExtracted_NoJSFiles is close but supplies a request with a valid URL and HTML response — the baseURL discovery loop succeeds. A regression that changed early-return semantics (e.g., panicking on empty input or mutating the slice) would not be caught.

**Recommendation:** Add TestReplayJSExtracted_EmptyInput with `requests := []ObservedRequest{}` and assert `result` is an empty slice (or equal to input). Add a case where every request has URL == "" and assert the input is returned unchanged.

```
	baseURL := ""
	for _, req := range requests {
		if req.URL != "" {
			baseURL = resolveBaseURL(req.URL)
			break
		}
	}
	if baseURL == "" {
		return requests
	}
```

---

### TEST-004 — medium — TEST

**File:** `pkg/crawl/jsreplay.go:567-583`

**Issue:** The PR description's test plan claims '19 test cases covering: ... probe failure', but the test file does not exercise either of the two skip paths in ReplayJSExtracted: (1) probeURL returning nil (network error / unreachable host), and (2) the explicit `if resp.StatusCode == http.StatusNotFound { continue }` branch — which is the core mechanism the PR uses to filter out wrong service-prefix combinations (the central design claim in LAB-1505 and the PR body: 'Filters out 404 responses from wrong prefix combinations'). Neither behavior is asserted: no test gives a JS body whose extracted paths include both a 200 endpoint and a 404 endpoint and verifies that only the 200 is appended to result. Without this, a regression that started including 404s (polluting downstream classification with phantom endpoints) would not be caught.

**Recommendation:** Add TestReplayJSExtracted_Filters404 that wires up a handler returning 200/JSON for one path (e.g., /workshop/api/products) and 404 for another (e.g., /identity/api/products), feeds JS that produces both via prefix expansion, and asserts the result contains exactly the 200 endpoint and not the 404. Add TestReplayJSExtracted_ProbeNetworkError that points the client at a closed/refusing server (or a transport that returns an error) and asserts the result equals the original requests (probe failures are silently dropped).

```
		resp := probeURL(ctx, cfg, fullURL)
		if resp == nil {
			continue
		}

		// Skip 404 responses — these are typically wrong service prefix
		// combinations (e.g., /identity/api/shop/products when the correct
		// prefix is /workshop/). Keeping them would pollute the spec with
		// endpoints that don't actually exist on that service.
		if resp.StatusCode == http.StatusNotFound {
			continue
		}
```

---

### TEST-005 — low — TEST

**File:** `pkg/crawl/jsreplay.go:631-633`

**Issue:** fetchJSBody has an explicit `if resp.StatusCode >= 400 { return nil }` guard (jsreplay.go:631-633), but no test triggers it. The two existing fetch tests (TestFetchJSBody_RejectsHTML, TestFetchJSBody_RejectsHTMLWithoutContentType) both serve 200 responses with HTML bodies. A regression that removed the >=400 guard would still pass all current tests as long as the server returned a non-HTML body — for example, a 500 error page in JSON would suddenly start being parsed for API paths. Coverage report shows fetchJSBody at 73.9%.

**Recommendation:** Add TestFetchJSBody_RejectsErrorStatus with a server returning 404 and 500 (with non-HTML bodies) and assert fetchJSBody returns nil for both. This locks in the error-status guard independently of the HTML-content guard.

```
	if resp.StatusCode >= 400 {
		return nil
	}
```

---

### TEST-006 — low — TEST

**File:** `pkg/crawl/jsreplay_test.go:375-386`

**Issue:** TestReplayJSExtracted_NoJSFiles only asserts `assert.Len(t, result, 1)`. It does not verify that the original request is preserved unchanged (URL, source, response body). A regression that returned a non-nil-but-mutated single element (e.g., emptied URL or rewritten Source) would still pass. This is the canonical no-op contract for the function and deserves a content assertion.

**Recommendation:** After the length assertion, also assert `assert.Equal(t, requests[0], result[0])` (or assert URL/Source/StatusCode individually) to lock in that the input is returned verbatim when there is no JS to process.

```
	result := ReplayJSExtracted(context.Background(), requests, JSReplayConfig{})
	assert.Len(t, result, 1)
```

---

### TEST-007 — medium — TEST

**File:** `pkg/crawl/jsreplay_test.go:388-415`

**Issue:** TestReplayJSExtracted_MaxEndpoints feeds 5 paths with MaxEndpoints=2 but asserts only `LessOrEqual(len(result), 3)` and `LessOrEqual(callCount, 2)`. Because the production loop iterates a `map[string]bool` (non-deterministic order) and skips probes whose response is nil or 404, the assertions pass even if the cap is silently broken or never reached (e.g., 0 probes still satisfies the bound). The test handler always returns 200/JSON, so the cap is the only thing that can stop probing — the assertions should be exact equalities (`Equal(t, 3, len(result))` and `Equal(t, 2, callCount)`). As written, a regression that increments past MaxEndpoints by one would still pass the upper-bound check on average runs, and a regression that never honors MaxEndpoints (0 probes) trivially passes both `LessOrEqual` checks.

**Recommendation:** Assert exact values: `assert.Equal(t, 3, len(result))` (1 original + exactly 2 probed) and `assert.Equal(t, 2, callCount)`. This converts the test from a one-sided bound into a precise behavioral contract on the MaxEndpoints cap.

```
	cfg := JSReplayConfig{Client: srv.Client(), MaxEndpoints: 2}
	result := ReplayJSExtracted(context.Background(), requests, cfg)

	assert.LessOrEqual(t, len(result), 3) // 1 original + max 2 probed
	assert.LessOrEqual(t, callCount, 2)
```

---

### TEST-008 — low — TEST

**File:** `pkg/crawl/jsreplay_test.go:417-437`

**Issue:** TestReplayJSExtracted_ContextCancellation cancels the context before calling ReplayJSExtracted and asserts `Len(result, 1)`. However, the same outcome (length 1) would result from any code path that simply found no JS paths to probe — the test does not distinguish 'context cancellation skipped probe' from 'extraction succeeded but probe network call failed for some other reason'. There is no instrumentation (e.g., a server hit counter) to prove zero probe attempts were made when ctx is canceled, nor an assertion that result[0] equals the original request.

**Recommendation:** Add a hit counter on the test server and assert `callCount == 0` (the canceled context must short-circuit the probe before any HTTP request is sent). Also assert `result[0]` equals the original request to confirm input passthrough.

```
	ctx, cancel := context.WithCancel(context.Background())
	cancel()
...
	result := ReplayJSExtracted(ctx, requests, JSReplayConfig{Client: srv.Client()})
	assert.Len(t, result, 1)
```

---

### DOC-001 — low — DOC

**File:** `CLAUDE.md:68`

**Issue:** CLAUDE.md (the project's authoritative architecture overview) describes pkg/crawl as: "Headless browser crawling via Katana, capture file I/O ... browser manager with Chrome lifecycle". This PR adds a meaningful third responsibility — post-crawl JS extraction with raw HTTP probing of API paths embedded in SPA bundles — that is not reflected. The repo's own dev-workflow rule (CLAUDE.md: "After adding or changing features, review README.md and CLAUDE.md for accuracy and update them if needed") is therefore not satisfied.

**Recommendation:** Update the pkg/crawl bullet in CLAUDE.md and consider adding a short "Pipeline" sub-bullet under the scan command that mentions JS-extracted endpoint replay so the architecture description matches what the binary now does.

```
- **pkg/crawl**: Headless browser crawling via Katana, capture file I/O (`ObservedRequest` JSON format), browser manager with Chrome lifecycle
```

---

### DOC-002 — low — DOC

**File:** `README.md`

**Issue:** The PR materially changes user-visible behavior of the scan command on SPAs (29 vs 1 endpoints discovered on crAPI). README has no reference to JS-bundle extraction, the 10 MB JS body limit, the MaxEndpoints cap, or the new debug log line family ("js-extract: ..."). Operators who tune verbose output or wonder why a SPA scan now produces dramatically more endpoints have no documented behavior to reference.

**Recommendation:** Add a short note in README under Usage (or a new "How it works" / "SPA support" section) describing the post-crawl JS extraction step — what it probes, what it filters, and how MaxEndpoints / Timeout interact.

```
(README.md does not mention 'jsreplay', 'js-extract', 'SPA bundle', or 'JavaScript bundle')
```

---

### DOC-003 — low — DOC

**File:** `pkg/crawl/doc.go`

**Issue:** Two new public symbols (JSReplayConfig, ReplayJSExtracted) were added to package crawl but the package-level godoc in pkg/crawl/doc.go was not updated to describe the JS-replay capability. The repository's own CLAUDE.md says: "After modifying a Go source file, update its package's doc.go if the change affects the package's public API or purpose." — this PR materially expands the package's public API without updating doc.go.

**Recommendation:** Add a short paragraph to pkg/crawl/doc.go describing the post-crawl JS-extraction step and add [JSReplayConfig] and [ReplayJSExtracted] to the "Key types" / new "Pipeline steps" list.

```
// Key types:
//   - [Crawler] orchestrates a headless browser crawl ...
//   - [BrowserManager] manages Chrome process lifecycle ...
//   - [ObservedRequest] and [ObservedResponse] represent captured HTTP traffic.
(JSReplayConfig and ReplayJSExtracted are absent)
```

---

blayne:
I've run this PR against crAPI, and I can see that it returns the endpoints whereas the code prior to this PR does not.

santiago-praetorian:
## Verification update — ready for re-review

All findings from the original review have been addressed across 16 iterations of incremental fixes (commits `d437ac1` → `a3eb7e2`). Subsequent CodeRabbit and GHAS reviewer comments are also addressed.

### End-to-end A/B against OWASP crAPI

Built both binaries from source and ran `vespasian scan http://127.0.0.1:9888 --dangerous-allow-private --max-pages 30 --timeout 5m` against the same crAPI deployment:

| Build | Endpoints in spec |
|---|---|
| `main` (current tip) | **2** (`/manifest.json`, `/chatbot/genai/state`) |
| `feat/js-api-extraction-v2` (`a3eb7e2`) | **44** |
| **delta** | **+42** |

This exceeds the original LAB-1505 target ("29 endpoints discovered"). The gain comes from the post-crawl JS-replay step extracting 34 raw API paths from crAPI's `main.8c78208c.js` bundle (1.65 MB after re-fetch past the 1 MB crawl truncation), then expanding them via discovered service prefixes (`identity/`, `workshop/`, `community/`) and filtering 404s.

### Live test suite

`./test/run-live-tests.sh` — **21 passed, 0 failed, 0 skipped**:

```
TARGET                      STATUS    ENDPOINTS  EXPECTED
classifier-edge             PASS      -          -
crawl-depth                 PASS      -          -
crawl-unreachable           PASS      0          0
edge-cases                  PASS      -          -
generate-graphql            PASS      8          8
generate-graphql-imports    PASS      2          2
generate-rest               PASS      8          8
generate-wsdl               PASS      3          3
graphql-server              PASS      8          8
import-base64               PASS      2          2
import-burp                 PASS      5          5
import-duplicates           PASS      2          2
import-empty                PASS      0          0
import-har                  PASS      3          3
import-malformed            PASS      0          0
import-mitmproxy            PASS      3          3
import-mitmproxy-native     PASS      3          3
import-unicode              PASS      3          3
rest-api                    PASS      8          8
soap-service                PASS      3          3
spec-edge                   PASS      -          -
```

### Iter-by-iter convergence

| Iter | Findings | Real blockers |
|---|---|---|
| 1 (your review) | 30 | 12 |
| 3 | 8 | 4 |
| 5 | 7 | 4 |
| 7 | 4 | 0 |
| 8 | 3 | 1 (auto-promote) |
| 9 | 1 | 0 |
| 13 | 1 | 1 (auto-promote) |
| 14 | 0 | 0 |
| 16 | 1 info | 0 |

### What landed beyond your original 30 findings

- **Strategy 3 prefix discovery** (hardened): catches SPAs that declare service prefixes as runtime constants (the crAPI pattern). Frequency = 1, per-bundle cap = 8 (TOTAL across strategies), API-indicator filter, 404-filter as the noise backstop.
- **Same-origin gate**: now canonicalizes default ports (`https://example.com` ≡ `https://example.com:443`) and lower-cases scheme/host.
- **Probe gate**: `--probe=false` keeps the scan passive (no JS-replay outbound HTTP).
- **3xx redirect follow** in `fetchJSBody`: bounded depth (5), per-hop SSRF + same-origin re-validation.
- **Interpolation-aware template-literal walker**: nested `${...}` + escaped backticks handled correctly.
- **CI-green**: `make check` (fmt + vet + golangci-lint + `go test -race`) passes.

CC @blayne — addressed everything from the 2026-04-30 review plus all subsequent bot/CI feedback. Ready when you are. 🙏

blayne (approved):

PR #64, opened 2026-04-13
Add Claude PR Assistant workflow

nsportsman:
## Summary
- Adds the standard `claude-code.yml` GitHub Actions workflow used across our other repos (guard-core, guard-tabularium, guard-cli, janus-framework)
- Uses the shared reusable workflow from `praetorian-inc/public-workflows`
- Triggers on PR open/synchronize and review comments

## Test plan
- [ ] Verify the `ANTHROPIC_API_KEY` secret is configured in the repo settings
- [ ] Open a test PR to confirm the workflow triggers

### DIFF
--- a/.github/workflows/claude-code.yml
+++ b/.github/workflows/claude-code.yml
@@ -1,0 +1,16 @@
+name: Claude PR Assistant
+
+on:
+  pull_request:
+    types: [synchronize, opened]
+  pull_request_review_comment:
+    types: [created]
+
+jobs:
+  claude-code-action:
+    uses: praetorian-inc/public-workflows/.github/workflows/claude-code.yml@main
+    permissions:
+      contents: write
+      pull-requests: write
+    secrets:
+      ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}


santiago-praetorian:
Thanks for setting this up @nsportsman. A few questions/items before merging:

**Reusable workflow pinned to `@main` (security concern)**

```yaml
uses: praetorian-inc/public-workflows/.github/workflows/claude-code.yml@main
```

`@main` is a mutable reference — anyone who can push to `public-workflows/main` can immediately alter the code that runs here with access to `secrets.ANTHROPIC_API_KEY`. Can we pin this to a tagged version (`@v1.0.0` or similar) instead of `@main`? If `public-workflows` doesn't publish tags yet, pinning to a commit SHA works too.

**`contents: write` permission — intentional?**

```yaml
permissions:
  contents: write
  pull-requests: write
```

`contents: write` is what allows Claude to push commits (e.g., responding to `@claude fix this` by pushing a fix to the PR branch). If the intent is for Claude to write code on the team's behalf, keeping `write` is correct. If the intent is review-and-comment only, `contents: read` is enough and reduces the `GITHUB_TOKEN`'s blast radius. Which one matches the intended use?

**No concurrency control**

Every push to a PR branch triggers a fresh Claude review. Rapid fix-up commits will queue parallel runs, each burning API credits on a diff that's about to be stale. Do you want to address this by adding a concurrency group that cancels in-progress runs?

```yaml
concurrency:
  group: claude-${{ github.event.pull_request.number }}
  cancel-in-progress: true
```

**`reopened` event type not included**

```yaml
types: [synchronize, opened]
```

If a PR is closed and then reopened (without new pushes), Claude won't re-run. Do you want to address this by adding `reopened` to the list, or is that workflow not a concern here?

PR #66, opened 2026-04-16
feat: replace Katana headless crawl with concurrent go-rod engine

WizardML7:
## Summary

Replaces Katana's serial headless crawl loop with a concurrent go-rod engine that runs 10 browser tabs in parallel (configurable via `--concurrency`). Katana processed URLs one at a time with a ~3s DOM stability wait per page — the new engine overlaps these waits, achieving a **7.5x speedup** on benchmarks (67s → 9s on the REST test target).

Katana's `standard.Engine` is retained for `--headless=false` — full Katana removal is a separate ticket.

## Changes

- **Concurrent crawl engine** (`pkg/crawl/engine.go`): go-rod worker pool with CDP network interception
- **URL frontier** (`pkg/crawl/frontier.go`): thread-safe queue with dedup, scope filtering, depth tracking
- **jsluice integration** (`pkg/crawl/jsextract.go`): JS AST parsing of external and inline scripts for endpoint discovery
- **Form discovery** (`pkg/crawl/forms.go`): extracts forms and emits synthetic POST requests
- **SSRF protection** (`pkg/crawl/scope.go`): DNS-resolution-based private IP blocking, critical for Guard platform deployments
- **New CLI flag**: `--concurrency` (default 10, max 50)

## Security

Passed backend security review. All HIGH findings addressed:
- H-1: Race condition in CDP event handler — fixed with atomic complete flag
- H-2: SSRF on crawled URLs — added private IP blocklist with DNS resolution
- H-3: Unbounded response body — truncation at collection time

## Test plan

- [ ] `make check` passes (fmt, vet, lint, test with `-race`)
- [ ] 8/8 expected REST API paths discovered against built-in test target
- [ ] 7.5x speedup verified: 67s (Katana) → 9s (go-rod) on same target
- [ ] E2E scan against snyk.io produces valid OpenAPI spec
- [ ] SSRF tests verify private IP rejection and `--dangerous-allow-private` bypass

### DIFF
--- a/cmd/vespasian/main.go
+++ b/cmd/vespasian/main.go
@@ -385,7 +385,8 @@ func setupBrowserAndSignals(rawHeaders []string, crawlOpts CrawlOptions, extraOp
 
 // CrawlCmd crawls a web application to capture HTTP traffic.
 type CrawlCmd struct {
-	URL string `arg:"" help:"Target URL to crawl"`
+	URL                   string `arg:"" help:"Target URL to crawl"`
+	DangerousAllowPrivate bool   `help:"Disable SSRF protection for crawling, allowing private/localhost targets. WARNING: Do not use on production systems." name:"dangerous-allow-private"`
 	CrawlOptions
 }
 
@@ -396,13 +397,14 @@ func (c *CrawlCmd) Run() error {
 	}
 
 	bs, err := setupBrowserAndSignals(c.Header, c.CrawlOptions, crawl.CrawlerOptions{
-		Depth:       c.Depth,
-		MaxPages:    c.MaxPages,
-		Timeout:     c.Timeout,
-		Scope:       c.Scope,
-		Headless:    c.Headless,
-		Proxy:       c.Proxy,
-		Concurrency: c.Concurrency,
+		Depth:        c.Depth,
+		MaxPages:     c.MaxPages,
+		Timeout:      c.Timeout,
+		Scope:        c.Scope,
+		Headless:     c.Headless,
+		Proxy:        c.Proxy,
+		Concurrency:  c.Concurrency,
+		AllowPrivate: c.DangerousAllowPrivate,
 	})
 	if err != nil {
 		return err
--- a/cmd/vespasian/main.go
+++ b/cmd/vespasian/main.go
@@ -563,13 +563,14 @@ func (c *ScanCmd) Run() error { //nolint:gocyclo // top-level orchestration
 	}
 
 	bs, err := setupBrowserAndSignals(c.Header, c.CrawlOptions, crawl.CrawlerOptions{
-		Depth:       c.Depth,
-		MaxPages:    c.MaxPages,
-		Timeout:     c.Timeout,
-		Scope:       c.Scope,
-		Headless:    c.Headless,
-		Proxy:       c.Proxy,
-		Concurrency: c.Concurrency,
+		Depth:        c.Depth,
+		MaxPages:     c.MaxPages,
+		Timeout:      c.Timeout,
+		Scope:        c.Scope,
+		Headless:     c.Headless,
+		Proxy:        c.Proxy,
+		Concurrency:  c.Concurrency,
+		AllowPrivate: c.DangerousAllowPrivate,
 	})
 	if err != nil {
 		return err
--- a/cmd/vespasian/main.go
+++ b/cmd/vespasian/main.go
@@ -275,6 +275,7 @@ type CrawlOptions struct {
 	Scope       string        `default:"same-origin" enum:"same-origin,same-domain" help:"Crawl scope"`
 	Headless    bool          `default:"true" help:"Use headless browser"`
 	Proxy       string        `help:"Proxy address for headless browser (e.g., http://127.0.0.1:8080). Note: TLS certificate verification is disabled during crawls."`
+	Concurrency int           `default:"10" help:"Number of concurrent browser tabs for headless crawling"`
 	NoRequestID bool          `name:"no-request-id" help:"Disable automatic X-Vespasian-Request-Id header"`
 	Verbose     bool          `short:"v" help:"Enable verbose logging"`
 }
@@ -395,12 +396,13 @@ func (c *CrawlCmd) Run() error {
 	}
 
 	bs, err := setupBrowserAndSignals(c.Header, c.CrawlOptions, crawl.CrawlerOptions{
-		Depth:    c.Depth,
-		MaxPages: c.MaxPages,
-		Timeout:  c.Timeout,
-		Scope:    c.Scope,
-		Headless: c.Headless,
-		Proxy:    c.Proxy,
+		Depth:       c.Depth,
+		MaxPages:    c.MaxPages,
+		Timeout:     c.Timeout,
+		Scope:       c.Scope,
+		Headless:    c.Headless,
+		Proxy:       c.Proxy,
+		Concurrency: c.Concurrency,
 	})
 	if err != nil {
 		return err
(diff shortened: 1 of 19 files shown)


eli-wald (changes requested):
# PR Review: `feat: replace Katana headless crawl with concurrent go-rod engine`

**PR**: praetorian-inc/vespasian#66
**Author**: WizardML7 (Domenic Lo Iacono)
**Stats**: +3,056 / -94 across 19 files

## Overview

Replaces Katana's serial headless crawl (1 tab at a time, ~3s DOM wait each) with a concurrent go-rod engine running N browser tabs in parallel. Katana's `standard.Engine` is retained for `--headless=false`. The PR also adds jsluice integration, form discovery, SSRF protection, and a URL frontier with proper completion detection.

## Strengths

- **Clean architecture**: Well-decomposed into focused files — `engine.go` (worker pool), `frontier.go` (thread-safe queue), `network.go` (CDP capture), `scope.go` (SSRF + scope), `links.go` (DOM extraction), `jsextract.go` (jsluice), `forms.go` (form discovery). Each has a clear single responsibility.
- **Excellent test coverage**: 7 new test files with 1,250+ lines of tests. Unit tests for every new component plus integration tests for the engine. The frontier tests are particularly thorough (blocking/unblocking, concurrent access, active-worker-prevents-early-done).
- **Security findings addressed**: The H-1 (race on CDP complete flag), H-2 (SSRF), and H-3 (unbounded body) fixes are clean and well-commented.
- **Frontier completion detection**: The `MarkActive`/`MarkIdle` pattern with `sync.Cond` correctly solves the "empty queue but workers still processing" problem — a subtle bug that many crawl implementations get wrong.
- **Backward compatibility**: Retaining Katana's standard engine for `--headless=false` is the right incremental approach.

## Issues & Suggestions

### 1. `scopeChecker` signature inconsistency (scope.go:109)

```go
func scopeChecker(seedURL string, scope string, allowPrivate bool) (func(string) bool, error)
```

But the integration test at `engine_integration_test.go:1074` calls it with only 2 args:

```go
scopeFn, err := scopeChecker(root.URL, "same-origin")
```

This won't compile. The test needs `scopeChecker(root.URL, "same-origin", true)`.

### 2. Double truncation in `mapNetworkToObservedRequest` (network.go)

The response body is truncated in `NetworkLoadingFinished` handler and again in `mapNetworkToObservedRequest`. The request body is also truncated in `mapNetworkToObservedRequest` but never at capture time. The double truncation is harmless but inconsistent — consider truncating only at the mapping stage for clarity.

### 3. Frontier queue growth (frontier.go)

The queue uses slice-based FIFO with `f.queue = f.queue[1:]` in Pop. This doesn't release memory from consumed entries — the underlying array grows but never shrinks. For a large crawl with thousands of URLs flowing through, this could hold onto significant memory. Consider using a ring buffer or periodically compacting:

```go
// In Pop, periodically compact to release consumed slots:
if len(f.queue) > 0 && cap(f.queue) > 4*len(f.queue) {
    compact := make([]urlEntry, len(f.queue))
    copy(compact, f.queue)
    f.queue = compact
}
```

### 4. `AllowPrivate` only wired in `ScanCmd`, not `CrawlCmd` (main.go)

In `ScanCmd.Run()`:

```go
Concurrency:  c.Concurrency,
AllowPrivate: c.DangerousAllowPrivate,
```

But in `CrawlCmd.Run()`, `AllowPrivate` is not set. If someone runs `vespasian crawl --dangerous-allow-private`, it won't work for headless mode. This seems like an oversight.

### 5. Event listener goroutine leak (engine.go:224)

```go
go waitEvents()
```

The `waitEvents` function (from `page.EachEvent`) blocks for the page's lifetime. When the page is closed via `page.Close()`, this goroutine should exit — but if `page.Close()` fails or the page is already detached, the goroutine may leak. Consider using the page's context to bound it, or at least document the lifecycle expectation.

### 6. `enrichFromPage` silently drops errors (engine.go:769-773)

```go
links, err := extractLinks(page)
if err != nil {
    return captured, nil
}
```

If link extraction fails (e.g., page crashed), all captured results are returned but with zero links and no logging. This is fine for resilience, but a verbose-mode log would help debugging.

### 7. Form `ContentType` not set in `formsToObservedRequests` for GET (forms.go)

POST forms get `content-type` set in headers, but GET forms don't get `content-type` cleared/set. Minor, but could confuse downstream classification if `ContentType` field carries a default.

## Security Considerations

- **SSRF protection is solid**: DNS-resolution-based check, covers all RFC 1918 ranges plus link-local (cloud metadata 169.254.169.254), IPv6 ULA/loopback. DNS failure defaults to reject — correct.
- **The `complete` flag atomic pattern (H-1)** correctly prevents double-finalization of CDP events under lock before the blocking CDP call.
- **Body truncation (H-3)** at `MaxResponseBodySize` prevents memory exhaustion from hostile XHR responses.
- **Consider**: The `isPrivateHost` function does DNS resolution synchronously. A malicious page with many cross-origin links could cause the scope checker to make hundreds of DNS queries. This is bounded by `maxPages` but worth noting.

## Verdict

This is a well-structured, well-tested PR. The architecture is clean, concurrency is handled carefully, and the security findings are properly addressed. The compilation error in the integration test (#1) and the `AllowPrivate` wiring gap (#4) should be fixed before merge. The rest are minor improvements.

eli-wald (approved):
# PR #66 Follow-up Review: Commit `949151d`

**Commit**: `949151d fix: address code review feedback from PR #66`
**Files changed**: 7 files, +59 / -24

## Issue Resolution Status

### 1. `scopeChecker` signature inconsistency — RESOLVED

The integration test at `engine_integration_test.go:269` now passes all 3 arguments:

```go
scopeFn, err := scopeChecker(root.URL, "same-origin", true)
```

Verified: `go vet -tags integration ./pkg/crawl/` passes cleanly.

### 2. Double truncation in `mapNetworkToObservedRequest` — RESOLVED

Truncation is now applied consistently at collection time only:
- Request body: truncated in `NetworkRequestWillBeSent` handler via `truncateBody([]byte(e.Request.PostData))`
- Response body: truncated in `NetworkLoadingFinished` handler (unchanged from before)
- `mapNetworkToObservedRequest` now passes bodies through without re-truncating

The test was updated from `TestMapNetworkToObservedRequest_BodyTruncation` to `TestMapNetworkToObservedRequest_BodyPassthrough` to reflect the new contract, and a new `TestTruncateBody_AtCollectionTime` test was added to explicitly verify the truncation function.

### 3. Frontier queue growth — RESOLVED

The compaction logic was added to `Pop()` in `frontier.go`:

```go
if cap(f.queue) > 4*len(f.queue) && len(f.queue) > 0 {
    compact := make([]urlEntry, len(f.queue))
    copy(compact, f.queue)
    f.queue = compact
}
```

This matches the suggested approach. The `len(f.queue) > 0` guard prevents unnecessary allocation when the queue is about to be empty.

### 4. `AllowPrivate` not wired in `CrawlCmd` — RESOLVED

`CrawlCmd` now has the `DangerousAllowPrivate` field and passes it through:

```go
type CrawlCmd struct {
    URL                   string `arg:"" help:"Target URL to crawl"`
    DangerousAllowPrivate bool   `help:"..." name:"dangerous-allow-private"`
    CrawlOptions
}
```

And in `CrawlCmd.Run()`:

```go
AllowPrivate: c.DangerousAllowPrivate,
```

### 5. Event listener goroutine lifecycle — RESOLVED

The `go waitEvents()` call now has a comment documenting the lifecycle:

```go
// Start the event listener in a goroutine. The goroutine exits when
// the page is closed (deferred above) or the page context expires.
// go-rod's EachEvent internally listens on the page's CDP session,
// which is torn down by page.Close().
go waitEvents()
```

This clarifies that the goroutine is bounded by the page lifecycle — no code change needed, just documentation.

### 6. `enrichFromPage` silently drops errors — RESOLVED

Link extraction errors are now logged when stderr is available:

```go
links, err := extractLinks(page)
if err != nil {
    if stderr != nil {
        fmt.Fprintf(stderr, "link extraction failed for %s: %v\n", pageURL, err)
    }
    return captured, nil
}
```

The function signature was updated to accept `stderr io.Writer` and the call site in `visitPage` passes `e.opts.Stderr`.

### 7. Form `ContentType`/Headers for GET forms — RESOLVED

The empty `Headers` map is now only initialized for POST forms, not unconditionally:

```go
// Before: Headers always initialized
obs := ObservedRequest{
    Headers: map[string]string{},
}

// After: Headers only initialized for POST
if f.Method == "POST" {
    obs.Headers = map[string]string{}
    ...
}
```

GET forms no longer carry an empty `Headers` map with no `content-type`, avoiding potential confusion in downstream classification.

## Verdict

All 7 issues from the original review have been addressed in commit `949151d`. The fixes are clean and well-scoped — no unnecessary changes beyond what was flagged. The integration test compiles cleanly with `-tags integration`. This PR is ready to merge.

PR #68, opened 2026-04-16
fix: propagate session cookies across headless crawl requests (LAB-2222)

eli-wald:
## Summary

- Fixes [LAB-2222](https://linear.app/praetorianlabs/issue/LAB-2222/session-cookie-not-propagated-across-crawl-requests-webgoat): session cookies passed via `-H "Cookie: ..."` were not propagated to subsequent requests during headless browser crawls
- Cookie headers are now injected directly into Chrome's cookie store via `Storage.setCookies` CDP protocol, bypassing the unreliable `Network.setExtraHTTPHeaders` path
- Katana's default incognito mode is disabled when cookies are present so injected cookies persist across all browser pages

## Root Cause

Katana's `addHeadersToPage` uses Chrome's `Network.setExtraHTTPHeaders` to set custom headers on each page. This does not reliably propagate `Cookie` headers across redirects and Fetch API-intercepted requests in headless Chrome's incognito mode. Additionally, Katana creates an incognito browser context by default, which starts with an empty cookie jar isolated from any external cookie injection.

## Changes

| File | Change |
|------|--------|
| `pkg/crawl/cookies.go` (new) | `ExtractCookieHeader` separates Cookie from other headers; `ParseCookiesToParams` converts cookie header values to CDP `NetworkCookieParam` entries |
| `pkg/crawl/cookies_test.go` (new) | 19 unit tests for cookie extraction and parsing |
| `pkg/crawl/browser.go` | Added `rod.Browser` connection to `BrowserManager`; added `SetCookies()` method for CDP cookie injection |
| `pkg/crawl/crawler.go` | Before Katana starts: extracts cookies, injects via CDP, sets `HeadlessNoIncognito` when cookies present |

## Test plan

- [x] Unit tests: 19 new tests for cookie parsing/extraction (all pass)
- [x] Full test suite: all existing tests pass (0 regressions)
- [x] WebGoat (JSESSIONID cookie auth): authenticated crawl captures 14 pages of authenticated content vs. 1 request (404 error) previously
- [x] Juice Shop (JWT Bearer + cookie): scan with auth produces 16 classified API endpoints with full OpenAPI spec (non-headless); headless mode limited by separate SPA crawling issues (LAB-2222 scope is cookie propagation only)
- [x] Non-Cookie custom headers (`Authorization`, `X-Custom`) continue to work alongside cookie injection


santiago-praetorian (changes requested):
Reviewed the four changed files against LAB-2222's scope (session cookie propagation in headless crawl). Tests pass; the approach (CDP `Storage.setCookies` + `HeadlessNoIncognito`) correctly addresses the root cause and the `HeadlessNoIncognito` interaction is verified against vendored Katana v1.5.0.

One real issue worth fixing before merge, plus a few optional polish items.

## Requested change

**`pkg/crawl/crawler.go:134-143` — whitespace-only Cookie header triggers `Storage.clearCookies`**

`ParseCookiesToParams` filters out whitespace-only pairs and returns a **nil** slice (not an empty slice) when the entire Cookie value is `"  ;  ;  "` or similar. Rod's `Browser.SetCookies` does a strict nil check:

```go
if cookies == nil { return proto.StorageClearCookies{}.Call(b) }
```

So a malformed-but-non-empty Cookie header will invoke `Storage.clearCookies` on the browser instead of being a no-op. Practical impact today is near-zero (the jar is empty at this point in the lifecycle), but it's a fragile contract if injection ever moves or gets called twice. Two-line guard:

```go
cookies, parseErr := ParseCookiesToParams(targetURL, cookieValue)
if parseErr != nil {
    return nil, fmt.Errorf("parse cookie header: %w", parseErr)
}
if len(cookies) > 0 {
    if err := browserMgr.SetCookies(cookies); err != nil {
        return nil, fmt.Errorf("inject cookies into browser: %w", err)
    }
    headers = remaining
    hasCookies = true
}
```

## Optional (nice-to-have, not blocking)

- **`pkg/crawl/cookies.go:72-79`** — one-line comment that `HttpOnly` is deliberately omitted (so the crawler allows JS `document.cookie` reads on apps that read auth state via JS). `HttpOnly` does not gate outbound request attachment, so the omission is correct — worth saying so to prevent a future "hardening" regression.
- **`pkg/crawl/cookies.go:73`** — one-line comment that `Path: "/"` is intentional (target URL's path is discarded by design for session-cookie use cases).
- **`pkg/crawl/cookies.go:75`** — doc note on `ParseCookiesToParams` that `Domain: u.Hostname()` produces a host-only cookie (no leading dot), so subdomain redirects won't carry the cookie. Correct for LAB-2222's scope; helps future operators whose targets redirect across subdomains.
- **`pkg/crawl/cookies_test.go:93`** — test uses hardcoded `k == "Cookie" || k == "cookie" || k == "COOKIE"`; prefer `strings.EqualFold` to match production behavior.

## Verified non-findings

- Error paths (`parse cookie header: %w`, `inject cookies into browser: %w`) do not leak cookie values — `url.Parse` errors cover only the target URL; rod surfaces CDP errors that don't echo input.
- `Secure: u.Scheme == "https"` is correct.
- `CustomHeaders` injection surface is unchanged — cookie values are removed from `headers` before `ToStringSlice`.
- `HeadlessNoIncognito: hasCookies` correctly prevents Katana from creating an incognito context on the remote browser (verified against `katana@v1.5.0/pkg/engine/hybrid/hybrid.go:75-76`).
- `BrowserManager.Close()` not calling the new `rod.Browser.Close()` is cosmetic — `Kill()` SIGKILLs the process and the WS gets EOF immediately.
- No new dependencies.

santiago-praetorian (changes requested):

**PR:** https://github.com/praetorian-inc/vespasian/pull/68  
**Linear:** https://linear.app/praetorianlabs/issue/LAB-2222/session-cookie-not-propagated-across-crawl-requests-webgoat  
**Head SHA:** `edaa2352a4a85ac92e3a66dc7524ae695f836395`  
**Reviewed at:** 2026-04-22T03:13:22Z

**Recommendation:** **REQUEST CHANGES**

|---|---|---|---|---|---|
| Count | 0 | 1 | 2 | 6 | 1 |
| Blockers | 0 | 1 | 2 | 0 | 0 |

---

## Blockers

### TEST-001 — medium — TEST

**File:** `pkg/crawl/browser.go:133-138`

**Issue:** SetCookies is a new exported method on BrowserManager. The only test that calls it runs inside TestCrawler_CookieHeaderPropagatesAcrossRedirects, which is gated behind //go:build integration (see separate finding). Coverage output confirms SetCookies is at 0.0% under the default build: `github.com/praetorian-inc/vespasian/pkg/crawl/browser.go:133: SetCookies 0.0%`. In particular, the explicit error branch `if b.browser == nil { return fmt.Errorf("browser not connected") }` has no test at all. Since the only non-integration code path that would reach this branch is a BrowserManager constructed by tests or callers that bypass NewBrowserManager, the guard currently exists as defensive dead-weight with no verification that it returns the expected error message (string is relied on for wrapping in crawler.go:153 `inject cookies: %w`).

**Recommendation:** Add a small non-integration unit test that constructs a BrowserManager literal with browser=nil and asserts SetCookies returns a non-nil error matching 'browser not connected' (or use errors.Is / strings.Contains). This takes ~10 lines and gives the branch a canary, plus documents the nil-safety contract. Example: `bm := &BrowserManager{}; err := bm.SetCookies(nil); require.Error(t, err); require.Contains(t, err.Error(), "not connected")`.

```
func (b *BrowserManager) SetCookies(cookies []*proto.NetworkCookieParam) error {
	if b.browser == nil {
		return fmt.Errorf("browser not connected")
	}
	return b.browser.SetCookies(cookies)
github.com/praetorian-inc/vespasian/pkg/crawl/browser.go:133:	SetCookies			0.0%
```

---

### TEST-004 — medium — TEST

**File:** `pkg/crawl/crawler.go:141-157`

**Issue:** The actual bug fix for LAB-2222 lives in crawlHeadless at pkg/crawl/crawler.go:141-157: (1) ExtractCookieHeader separates the Cookie header from other headers, (2) ParseCookiesToParams converts it to NetworkCookieParam entries, (3) browserMgr.SetCookies injects them, and (4) `Headers: extraHeaders` is passed to newRodEngine instead of the original c.opts.Headers (so Cookie is not doubly-injected via Network.setExtraHTTPHeaders). The two cookies_test.go helpers are tested in isolation, but no test exercises the composition — specifically, no test proves that when Headers={"Cookie": "X", "Authorization": "Y"} is passed to crawlHeadless/NewCrawler, (a) SetCookies is called once with the parsed X, (b) the engine is configured with a Headers map that no longer contains Cookie but still contains Authorization, and (c) errors from ParseCookiesToParams or SetCookies propagate as `parse cookies: ...` / `inject cookies: ...`. This wiring is the failure surface the PR targets, and a regression here (e.g., someone refactoring Headers plumbing) would not be caught by the cookies_test.go unit tests. The integration-tagged TestCrawler_CookieHeaderPropagatesAcrossRedirects covers the end-to-end behavior only when the integration tag is enabled.

**Recommendation:** Add a non-integration-tagged test (either table-driven inside crawler_test.go, or a new file) that exercises the crawlHeadless wiring with a seam around SetCookies. Two reasonable options: (1) refactor crawlHeadless to accept a cookie-injector interface so tests can pass a fake; (2) keep BrowserManager as the concrete dependency but wrap SetCookies/wsURL behind a small interface and inject it into Crawler for tests. Either way, add at least three cases: Cookie header present (SetCookies called, engine headers stripped), no Cookie header (SetCookies not called, engine headers unchanged), and ParseCookiesToParams error (crawlHeadless returns wrapped 'parse cookies:' error). This costs ~50 LoC and closes the regression gap even when the integration tag is off.

```
	cookieValue, extraHeaders := ExtractCookieHeader(c.opts.Headers)
	if cookieValue != "" {
		params, cerr := ParseCookiesToParams(targetURL, cookieValue)
		if cerr != nil {
			return nil, fmt.Errorf("parse cookies: %w", cerr)
		if serr := browserMgr.SetCookies(params); serr != nil {
			return nil, fmt.Errorf("inject cookies: %w", serr)
		Headers:       extraHeaders,
```

---

### TEST-005 — high — TEST

**File:** `pkg/crawl/engine_integration_test.go:1`

**Issue:** TestCrawler_CookieHeaderPropagatesAcrossRedirects is the only test that exercises the LAB-2222 WebGoat-style flow end-to-end: it stands up a Spring-Security-like httptest server that 302s unauthenticated requests to /login and returns the protected payload only when JSESSIONID=TESTSESSION123 is attached, then drives a real Crawler with Headers={"Cookie": "JSESSIONID=..."} and asserts /protected returns 200. This test directly regresses LAB-2222 and would fail on the pre-fix code (where the cookie went through Network.setExtraHTTPHeaders and was dropped on the redirect). However, the entire file is gated by //go:build integration (line 1), so a standard `go test ./...` run skips it. The project's CI does not appear to invoke `-tags integration` for this package, which means the explicit regression barrier for the bug this PR claims to fix provides no automated protection against re-introduction. The PR body claims '19 unit tests for cookie extraction and parsing' as the safety net, but those tests only cover pure-Go helpers — none of them prove the cookie actually propagates across a redirect. If Katana/go-rod is later upgraded or crawler.go:141-155 is refactored such that SetCookies() is no longer called for Cookie headers, no default-build test will catch it.

**Recommendation:** Either (a) add a non-integration-tagged unit test that exercises the wiring in crawlHeadless (e.g., a fake BrowserManager with an injectable SetCookies spy, asserting that when Headers contains 'Cookie' the spy is called with the parsed NetworkCookieParam and Headers passed to the engine no longer contains 'Cookie'), or (b) make the integration build tag part of the standard CI pipeline for this package and document it in the Makefile/CI config so the regression test is always executed. Option (a) is cheaper and keeps the fast test loop green-field friendly; option (b) is a one-line CI change but requires Chrome on the runner.

```
//go:build integration
// TestCrawler_CookieHeaderPropagatesAcrossRedirects regresses LAB-2222: a
// session cookie supplied via -H "Cookie: ..." must be attached to every
Headers:      map[string]string{"Cookie": "JSESSIONID=" + sessionID},
if !seenProtected {
t.Errorf("expected authenticated 200 on /protected; session cookie did not propagate. Captured:\n  %s", strings.Join(seen, "\n  "))
```

---



## Non-blocking

### QUAL-001 — low — QUAL

**File:** `pkg/crawl/browser.go:124-127`

**Issue:** NewBrowserManager now connects a rod.Browser (stored in b.browser) but BrowserManager.Close() only calls Kill() and cleanup() on the launcher, never calling b.browser.Close(). The rod.Browser.Close() sends a CDP Browser.close command and cleans up internal goroutines/connections. While Kill() terminates the Chrome process (which causes the WebSocket to drop), the rod.Browser object may have background goroutines that remain live until the connection drops and their read loops exit with an error. In practice this is not a resource leak because the process death is immediate, but it is a deviation from the pattern established by the rest of the rod API where callers are expected to close the Browser before the process exits. This also makes the Close() contract inconsistent: it kills the process but leaves an open client object.

**Recommendation:** Add b.browser.Close() in BrowserManager.Close() (or in cleanup()) before calling launcher.Kill(). Ignore the error from browser.Close() since the process may already be dead: e.g., `_ = b.browser.Close()`. This makes the shutdown sequence symmetric with construction and aligns with rod's documented lifecycle.

```
func (b *BrowserManager) Close() {
	b.Kill()
	b.cleanup()
}
// rod.Browser.Close() not called; b.browser goroutines are abandoned until kill drops the WebSocket
```

---

### QUAL-002 — low — QUAL

**File:** `pkg/crawl/crawler.go`

**Issue:** The PR body's Changes table states 'sets HeadlessNoIncognito when cookies present' for crawler.go, but no such flag is set anywhere in the diff. This claim appears to be stale: PR #66 replaced Katana's hybrid headless engine (which used incognito browser contexts) with a go-rod engine that uses the default browser context (non-incognito). Accordingly, HeadlessNoIncognito is irrelevant on the headless path, but the PR description still advertises it as a change. The functionality is correct — go-rod's default context is not incognito, so cookies injected via Storage.setCookies are visible to all pages — but the description creates confusion for future readers trying to understand what was changed.

**Recommendation:** Update the PR description (and any associated Linear comment) to remove the HeadlessNoIncognito claim and instead explain that the headless path now uses go-rod's default browser context (which is not incognito), so injected cookies are shared without needing that flag.

```
// PR body Changes table: 'pkg/crawl/crawler.go | Before Katana starts: extracts cookies, injects via CDP, sets HeadlessNoIncognito when cookies present'
// Actual diff: no HeadlessNoIncognito assignment anywhere in crawler.go
```

---

### QUAL-003 — nit — QUAL

**File:** `pkg/crawl/crawler.go:149-154`

**Issue:** The cookie injection block uses cerr and serr as error variable names to avoid shadowing the outer err variable. Go idiom strongly prefers reusing err with short-scoped if-initializer statements (if err := f(); err != nil { ... }) or assigning to a fresh err in a nested block, rather than introducing unique names like cerr/serr. These names are unfamiliar to Go readers and make the code harder to scan.

**Recommendation:** Use the inline if-initializer form to keep err as the error name:

  if err := browserMgr.SetCookies(params); err != nil {
      return nil, fmt.Errorf("inject cookies: %w", err)
  }

For ParseCookiesToParams, assign to params and err separately:

  params, err := ParseCookiesToParams(targetURL, cookieValue)
  if err != nil {
      return nil, fmt.Errorf("parse cookies: %w", err)
  }

```
		params, cerr := ParseCookiesToParams(targetURL, cookieValue)
		if cerr != nil {
			return nil, fmt.Errorf("parse cookies: %w", cerr)
		}
		if serr := browserMgr.SetCookies(params); serr != nil {
			return nil, fmt.Errorf("inject cookies: %w", serr)
```

---

### QUAL-004 — low — QUAL

**File:** `pkg/crawl/engine_integration_test.go:371-380`

**Issue:** TestCrawler_CookieHeaderPropagatesAcrossRedirects checks that a /protected URL appears in results with StatusCode == 200 to confirm cookie propagation. However, the test server's /protected handler returns 200 for both the authenticated case ('Protected content' body) and could be confused if a caching layer or proxy intercept intervened. More importantly, the test doesn't verify the response body contains 'Protected content', which is the only definitive proof that the authenticated server handler actually ran. A status code of 200 alone is a necessary but not sufficient condition — the assertion matches the test's stated intent but is weaker than it could be.

**Recommendation:** Add a body content check alongside the status code assertion:

  if r.URL == srv.URL+"/protected" && r.Response.StatusCode == 200 &&
      strings.Contains(string(r.Response.Body), "Protected content") {
      seenProtected = true
  }

This makes the regression test conclusively verify that the authenticated handler was reached, not just that something returned 200.

```
		if r.URL == srv.URL+"/protected" && r.Response.StatusCode == 200 {
			seenProtected = true
			break
		}
```

---



## Ticket alignment

_None — LAB-2222 acceptance criterion met; regression test exercises the Spring-Security-style redirect flow and PR author reports 14 authenticated pages captured against WebGoat._



## Tests & documentation

### TEST-001 — medium — TEST

**File:** `pkg/crawl/browser.go:133-138`

**Issue:** SetCookies is a new exported method on BrowserManager. The only test that calls it runs inside TestCrawler_CookieHeaderPropagatesAcrossRedirects, which is gated behind //go:build integration (see separate finding). Coverage output confirms SetCookies is at 0.0% under the default build: `github.com/praetorian-inc/vespasian/pkg/crawl/browser.go:133: SetCookies 0.0%`. In particular, the explicit error branch `if b.browser == nil { return fmt.Errorf("browser not connected") }` has no test at all. Since the only non-integration code path that would reach this branch is a BrowserManager constructed by tests or callers that bypass NewBrowserManager, the guard currently exists as defensive dead-weight with no verification that it returns the expected error message (string is relied on for wrapping in crawler.go:153 `inject cookies: %w`).

**Recommendation:** Add a small non-integration unit test that constructs a BrowserManager literal with browser=nil and asserts SetCookies returns a non-nil error matching 'browser not connected' (or use errors.Is / strings.Contains). This takes ~10 lines and gives the branch a canary, plus documents the nil-safety contract. Example: `bm := &BrowserManager{}; err := bm.SetCookies(nil); require.Error(t, err); require.Contains(t, err.Error(), "not connected")`.

```
func (b *BrowserManager) SetCookies(cookies []*proto.NetworkCookieParam) error {
	if b.browser == nil {
		return fmt.Errorf("browser not connected")
	}
	return b.browser.SetCookies(cookies)
github.com/praetorian-inc/vespasian/pkg/crawl/browser.go:133:	SetCookies			0.0%
```

---

### TEST-002 — low — TEST

**File:** `pkg/crawl/cookies_test.go:22-100`

**Issue:** ExtractCookieHeader explicitly documents determinism as a correctness property: `// Iterate in sorted key order so concatenation of differently-cased // "Cookie" headers is deterministic across runs (Go map iteration is // randomized). Matters for duplicate cookie precedence.` The implementation sorts keys before iterating, which is the sole reason the result is stable run-to-run. However, no test case in TestExtractCookieHeader inputs more than one Cookie header with different casings (e.g., both 'Cookie' and 'cookie' in the same map), so the sort-and-join contract is not asserted. Since Go map iteration is randomized, removing the `sort.Strings(keys)` line today would not cause any existing test to fail deterministically — only flakes over many runs. This is the exact kind of behavior that belongs to a regression test: the property exists specifically because duplicate-casing cookie precedence matters for LAB-2222 clones.

**Recommendation:** Add a subtest with two Cookie headers of different casings and assert the concatenation order is fixed. Example: `{name: "duplicate cookie headers concatenated in sorted key order", headers: map[string]string{"Cookie": "a=1", "cookie": "b=2"}, wantCookie: "a=1; b=2" /* Cookie (uppercase 'C') sorts before cookie (lowercase 'c') */, wantRemainCount: 0}`. Optionally run it in a loop (e.g., 100 iterations) to drive out the Go map randomization.

```
// Iterate in sorted key order so concatenation of differently-cased
// "Cookie" headers is deterministic across runs (Go map iteration is
// randomized). Matters for duplicate cookie precedence.
keys := make([]string, 0, len(headers))
sort.Strings(keys)
```

---

### TEST-003 — low — TEST

**File:** `pkg/crawl/cookies_test.go:119-210`

**Issue:** ParseCookiesToParams explicit branches are covered: invalid URL (':://invalid'), bare hostname ('example.com'), scheme-only ('http://'), non-http scheme ('ftp://'), empty cookie value, whitespace-only pairs, cookie without '=' value, and cookie with multiple '=' in value. Coverage sits at 95.2%. However, one silent-skip branch is asserted only by 'count' not by behavior: a pair like '=bare-value' (empty name after split) is silently `continue`d (cookies.go:83-85 `if name == "" { continue }`). The current 'whitespace-only pairs' case covers the trimmed-empty branch, but there is no explicit case for a valued pair with an empty name (e.g., `cookieValue: "=abc; valid=yes"`) that would assert exactly one param comes back and it is 'valid'. Not a coverage gap in percentage terms, but an intent gap — a future refactor could change the empty-name branch to emit an error without any test failing.

**Recommendation:** Add one more subtest to TestParseCookiesToParams: `{name: "pair with empty name silently skipped", targetURL: "https://example.com", cookieValue: "=orphan; valid=yes", wantCount: 1}`. Trivial addition; documents the intentional silent-skip behavior.

```
parts := strings.SplitN(pair, "=", 2)
name := strings.TrimSpace(parts[0])
if name == "" {
	continue
}
github.com/praetorian-inc/vespasian/pkg/crawl/cookies.go:56:	ParseCookiesToParams		95.2%
```

---

### TEST-004 — medium — TEST

**File:** `pkg/crawl/crawler.go:141-157`

**Issue:** The actual bug fix for LAB-2222 lives in crawlHeadless at pkg/crawl/crawler.go:141-157: (1) ExtractCookieHeader separates the Cookie header from other headers, (2) ParseCookiesToParams converts it to NetworkCookieParam entries, (3) browserMgr.SetCookies injects them, and (4) `Headers: extraHeaders` is passed to newRodEngine instead of the original c.opts.Headers (so Cookie is not doubly-injected via Network.setExtraHTTPHeaders). The two cookies_test.go helpers are tested in isolation, but no test exercises the composition — specifically, no test proves that when Headers={"Cookie": "X", "Authorization": "Y"} is passed to crawlHeadless/NewCrawler, (a) SetCookies is called once with the parsed X, (b) the engine is configured with a Headers map that no longer contains Cookie but still contains Authorization, and (c) errors from ParseCookiesToParams or SetCookies propagate as `parse cookies: ...` / `inject cookies: ...`. This wiring is the failure surface the PR targets, and a regression here (e.g., someone refactoring Headers plumbing) would not be caught by the cookies_test.go unit tests. The integration-tagged TestCrawler_CookieHeaderPropagatesAcrossRedirects covers the end-to-end behavior only when the integration tag is enabled.

**Recommendation:** Add a non-integration-tagged test (either table-driven inside crawler_test.go, or a new file) that exercises the crawlHeadless wiring with a seam around SetCookies. Two reasonable options: (1) refactor crawlHeadless to accept a cookie-injector interface so tests can pass a fake; (2) keep BrowserManager as the concrete dependency but wrap SetCookies/wsURL behind a small interface and inject it into Crawler for tests. Either way, add at least three cases: Cookie header present (SetCookies called, engine headers stripped), no Cookie header (SetCookies not called, engine headers unchanged), and ParseCookiesToParams error (crawlHeadless returns wrapped 'parse cookies:' error). This costs ~50 LoC and closes the regression gap even when the integration tag is off.

```
	cookieValue, extraHeaders := ExtractCookieHeader(c.opts.Headers)
	if cookieValue != "" {
		params, cerr := ParseCookiesToParams(targetURL, cookieValue)
		if cerr != nil {
			return nil, fmt.Errorf("parse cookies: %w", cerr)
		if serr := browserMgr.SetCookies(params); serr != nil {
			return nil, fmt.Errorf("inject cookies: %w", serr)
		Headers:       extraHeaders,
```

---

### TEST-005 — high — TEST

**File:** `pkg/crawl/engine_integration_test.go:1`

**Issue:** TestCrawler_CookieHeaderPropagatesAcrossRedirects is the only test that exercises the LAB-2222 WebGoat-style flow end-to-end: it stands up a Spring-Security-like httptest server that 302s unauthenticated requests to /login and returns the protected payload only when JSESSIONID=TESTSESSION123 is attached, then drives a real Crawler with Headers={"Cookie": "JSESSIONID=..."} and asserts /protected returns 200. This test directly regresses LAB-2222 and would fail on the pre-fix code (where the cookie went through Network.setExtraHTTPHeaders and was dropped on the redirect). However, the entire file is gated by //go:build integration (line 1), so a standard `go test ./...` run skips it. The project's CI does not appear to invoke `-tags integration` for this package, which means the explicit regression barrier for the bug this PR claims to fix provides no automated protection against re-introduction. The PR body claims '19 unit tests for cookie extraction and parsing' as the safety net, but those tests only cover pure-Go helpers — none of them prove the cookie actually propagates across a redirect. If Katana/go-rod is later upgraded or crawler.go:141-155 is refactored such that SetCookies() is no longer called for Cookie headers, no default-build test will catch it.

**Recommendation:** Either (a) add a non-integration-tagged unit test that exercises the wiring in crawlHeadless (e.g., a fake BrowserManager with an injectable SetCookies spy, asserting that when Headers contains 'Cookie' the spy is called with the parsed NetworkCookieParam and Headers passed to the engine no longer contains 'Cookie'), or (b) make the integration build tag part of the standard CI pipeline for this package and document it in the Makefile/CI config so the regression test is always executed. Option (a) is cheaper and keeps the fast test loop green-field friendly; option (b) is a one-line CI change but requires Chrome on the runner.

```
//go:build integration
// TestCrawler_CookieHeaderPropagatesAcrossRedirects regresses LAB-2222: a
// session cookie supplied via -H "Cookie: ..." must be attached to every
Headers:      map[string]string{"Cookie": "JSESSIONID=" + sessionID},
if !seenProtected {
t.Errorf("expected authenticated 200 on /protected; session cookie did not propagate. Captured:\n  %s", strings.Join(seen, "\n  "))
```

---

### DOC-001 — low — DOC

**File:** `pkg/crawl/doc.go:36-47`

**Issue:** The PR adds three new exported symbols to the crawl package: ExtractCookieHeader, ParseCookiesToParams, and BrowserManager.SetCookies. doc.go was updated with a Session-cookie helpers section that covers ExtractCookieHeader and ParseCookiesToParams, but omits BrowserManager.SetCookies. That method is the actual CDP injection entry-point — the helpers just produce its input. A reader scanning doc.go for 'how do I set cookies on the browser' will not find the answer. Per CLAUDE.md: 'After modifying a Go source file, update its package doc.go if the change affects the package's public API or purpose.'

**Recommendation:** Add a bullet to the Session-cookie helpers list in doc.go calling out BrowserManager.SetCookies as the CDP injection step, e.g., '- [BrowserManager.SetCookies] injects parsed cookie params into Chrome's cookie store via the Storage.setCookies CDP command, surviving redirects and Fetch interceptions.'

```
// Session-cookie helpers (LAB-2222) let callers bootstrap Chrome's cookie
//   - [ExtractCookieHeader] separates Cookie values (case-insensitively)
//   - [ParseCookiesToParams] converts a Cookie header value into CDP
```

---



3 review-thread findings were filtered because the PR history shows them as resolved by the author (see `<worktree>/.review/addressed.json`).


---

santiago-praetorian (approved):
**Approved.** LAB-2222 is solved cleanly — session cookies now survive Spring-Security-style 302 redirects via CDP `Storage.setCookies`, and the wiring is covered both by default-suite `TestApplyCookieHeader_*` subtests and the integration-tagged `TestCrawler_CookieHeaderPropagatesAcrossRedirects` (real `http.StatusFound` + `"Protected content"` body assertion). All prior-review findings (TEST-001..005, QUAL-001/003/004) have been addressed in commits `617102f` and `1cf02d6`.

## Non-blocking findings (low / nit)

All three are fine to land in-PR or defer to a follow-up ticket — none of them block approval.

### DOC-001 · low · `pkg/crawl/doc.go:36-46`

The session-cookie helpers section documents `[ExtractCookieHeader]` and `[ParseCookiesToParams]` as godoc-linked bullets but omits `[ApplyCookieHeader]` (mentioned in prose only) and `[BrowserManager.SetCookies]` (absent entirely). Godoc links enable IDE navigation and `pkg.go.dev` cross-referencing.

Suggested two bullets to add:

```go
//   - [ApplyCookieHeader] orchestrates the full extract/parse/inject
//     pipeline; callers pass browserMgr.SetCookies as the injector.
//   - [BrowserManager.SetCookies] injects parsed cookies into Chrome's
//     cookie store via the Storage.setCookies CDP command, surviving
//     redirects and Fetch interceptions.
```

### QUAL-001 · low · PR description / commit `2e04a94`

The Changes table and commit body still claim Katana's `HeadlessNoIncognito` is disabled when cookies are present. No such code exists in the diff (the headless path is now go-rod, not Katana's hybrid engine). A future maintainer reading `git log` to understand how cookies persist would be misled. Recommend amending the PR description and/or adding a clarification commit that replaces the incognito claim with: cookies are injected via `Storage.setCookies` into the `rod.Browser` instance before navigation; incognito mode is not relevant to this fix.

### QUAL-002 · nit · `pkg/crawl/crawler.go:149-152`

`ApplyCookieHeader` error is returned bare:

```go
extraHeaders, err := ApplyCookieHeader(c.opts.Headers, targetURL, browserMgr.SetCookies)
if err != nil {
    return nil, err
}
```

Every other error in `crawlHeadless` is wrapped (`scope setup: %w`, `create engine: %w`, etc.). One-line fix to match the surrounding style:

```go
return nil, fmt.Errorf("apply cookies: %w", err)
```

`ApplyCookieHeader`'s inner prefixes (`parse cookies:`, `inject cookies:`) are preserved via `%w`, and operators get a clear call-site label in the error chain.

## Before merging — please rebase on `main`

This branch is ~8 commits behind `main`. Please **rebase before merging**. The rebase pulls in:

- PR #84 (LAB-2308) — adds `--dangerous-allow-private` on every `vespasian crawl` invocation in `test/run-live-tests.sh`, so local live tests pass without manual flag-patching.
- PR #91 — `claude-code.yml` workflow updates.

Without the rebase, `./test/run-live-tests.sh --targets rest-api,soap-service,graphql-server` fails locally for reasons unrelated to this PR (silent empty-capture on localhost seeds — separately tracked as LAB-2438 / PR #92, which also lands a loud error in the engine).

If you have local working-tree edits to `test/run-live-tests.sh` from your own live-test workaround (e.g., adding the flag to the rest-api crawl by hand), discard them before the rebase — `main` already has them.

---

_Review generated by `praetorian-capability-pr-review`. Full roll-up and agent artifacts at `.worktrees/praetorian-capability-pr-review/pr-68/` in the reviewer's workspace._

PR #89, opened 2026-04-21
fix(importer): support mitmproxy native .mitm flow format (LAB-2309)

### LINKED ISSUE #85: mitmproxy import doesn't work?
I'm using `mitmproxy` version 12.2.1 (macos, installed through brew) and trying to import flows just throws an error:

```
vespasian: error: import failed: mitmproxy importer: expected JSON array or object, got "2"
```

This doesn't seem to change if I use `--verbose` so any help trying to figure it out would be great 😄

santiago-praetorian:
## Summary

- Fixes [#85](https://github.com/praetorian-inc/vespasian/issues/85) / [LAB-2309](https://linear.app/praetorianlabs/issue/LAB-2309): `vespasian import mitmproxy <file>.mitm` failed with `expected JSON array or object, got "2"` because the importer only accepted the JSON export. mitmproxy's native `File > Save` format is binary tnetstring, which starts with an ASCII digit (the length prefix — exactly what "2" was).
- Adds a minimal tnetstring decoder (`pkg/importer/tnetstring.go`) and wires it into `MitmproxyImporter.Import` via first-byte dispatch. JSON path unchanged.
- Hardened against untrusted input: per-element size cap (64 MB), depth cap (64), list/dict cardinality cap (1 M), stream-level flow count cap (500 K).
- Extensive test coverage including a **real mitmproxy-produced fixture** vendored from `mitmproxy/test/mitmproxy/data/dumpfile-7.mitm` (MIT license), plus a CI drift check that ensures the engineer-generated fixture matches its generator output byte-for-byte.

## What's in the diff

- `pkg/importer/tnetstring.go` — decoder (subset of format used by `HTTPFlow.get_state()`: bytes, string, int, float, bool, null, list, dict)
- `pkg/importer/mitmproxy.go` — format dispatch + native flow → `ObservedRequest` translation
- `internal/tnetenc/` — shared encoder used by test helpers and the fixture generator
- `pkg/importer/testhelpers_test.go` — `withTempCap` helper + raw tnetstring-element builders
- `test/fixtures/sample-mitmproxy.mitm` — generated fixture for the import-mitmproxy-native live test
- `test/fixtures/real-mitmproxy.mitm` — vendored real mitmproxy output
- `test/fixtures/gen_mitmproxy_native/` — generator program so the sample fixture can be regenerated
- `test/run-live-tests.sh` — new `import-mitmproxy-native` target

## Test plan

- [x] `make check` (fmt, vet, golangci-lint, `go test -race ./...`) — 0 issues
- [x] `pkg/importer` coverage: **92.9%** (CI threshold 80%)
- [x] `internal/tnetenc` coverage: **100%**
- [x] Live tests: `./test/run-live-tests.sh --targets import-mitmproxy,import-mitmproxy-native` — both PASS
- [x] End-to-end: `./bin/vespasian import mitmproxy <real-.mitm-file>` succeeds against four real dumpfiles from upstream mitmproxy (`dumpfile-7.mitm`, `-10.mitm`, `-019.mitm`, `-19.mitm`)
- [x] Regression test `TestMitmproxyImporter_Native_RegressionLAB2309` asserts the exact failure mode in the original issue (file starts with ASCII digit) now decodes cleanly
- [x] Drift check `TestMitmproxyFixture_MatchesGenerator` ensures `sample-mitmproxy.mitm` stays in sync with `gen_mitmproxy_native/main.go`

## Acknowledgements

Thanks to @yaleman for reporting the issue and providing a clear reproduction.


eli-wald (changes requested):

**PR:** https://github.com/praetorian-inc/vespasian/pull/89
**Linear:** https://linear.app/praetorianlabs/issue/LAB-2309/external-issue-mitmproxy-import-doesnt-work
**Head SHA:** `094972f935b4bce79146021125c9192e10109708`
**Reviewed at:** 2026-04-22

Main takeaways are to implement some additional scheme checking in `pkg/importer/mitmproxy.go` and add some additional test coverage. Looks good otherwise.

| Totals   | critical | high | medium | low | nit |
|----------|----------|------|--------|-----|-----|
| Count    | 0        | 0    | 2      | 10  | 7   |
| Blockers | 0        | 0    | 2      | 3   | 2   |

Acceptance criteria (LAB-2309): **all 3 implemented.**
- AC1 — native `.mitm` import works via first-byte dispatch into a tnetstring decoder. ✓
- AC2 — JSON export path preserved. ✓ (`TestMitmproxyImporter_*_JSON` still passes)
- AC3 — regression test `TestMitmproxyImporter_Native_RegressionLAB2309` asserts the ASCII-digit failure mode decodes cleanly. ✓

CodeRabbit bot left 5 earlier comments. All 5 are confirmed fixed in HEAD (`pairsParsed` counter replaces `len(result)`; `payloadPreview` truncates errors; key-existence asserts precede order checks; CONNECT URL shape characterized with tests; vendored fixture commit-pinned). Those findings are filtered out of this roll-up.

---

## Blockers

**File:** `pkg/importer/mitmproxy.go:258-264`
**CVSS:** `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:C/C:L/I:L/A:N` (base 5.0)

**Issue:** `flow.Request.Scheme` is taken verbatim from `reqMap["scheme"]` and passed to `constructURL` with no whitelist. mitmproxy itself only emits `http`/`https`, but the new native path accepts untrusted `.mitm` files (per the PR's own threat model). An attacker can forge scheme=`file`, `javascript`, `gopher`, `data`, `ldap`, `dict`, etc. The resulting `ObservedRequest.URL` carries that scheme into any downstream consumer that re-fetches, replays, or inspects it. Enforcing at the importer boundary removes a whole class of scheme-confusion bugs in downstream code that is not documented as having to re-validate.

**Recommendation:** In `flowFromNativeState` (and for parity in `importJSON.parseFlow`), reject flows whose scheme is not `http` or `https`:

```go
if scheme := tnetBytesOrString(reqMap["scheme"]); scheme != "http" && scheme != "https" {
    return mitmproxyFlow{}, fmt.Errorf("unsupported scheme %q", scheme)
}
```

```go
req := mitmproxyRequest{
    Method:  method,
    Scheme:  tnetBytesOrString(reqMap["scheme"]),
    Host:    tnetBytesOrString(reqMap["host"]),
    Port:    port,
    Path:    buildRequestPath(reqMap, method),
```

---

### TEST-001 — low — TEST (missing tests)

**File:** `pkg/importer/mitmproxy_fixture_test.go:61-86`

**Issue:** `TestMitmproxyFixture_MatchesGenerator` byte-compares the committed `sample-mitmproxy.mitm` against generator stdout, proving `committed == generated`. But no unit test actually feeds those bytes through `MitmproxyImporter.Import`. The end-to-end path for the generated fixture is only exercised in the gated live test `import-mitmproxy-native`. A generator that silently produces unparseable bytes would pass the drift check.

**Recommendation:** Add `TestMitmproxyImporter_Native_GeneratedFixture` that reads `test/fixtures/sample-mitmproxy.mitm`, imports it, and asserts on the three known flows.

---

### TEST-002 — low — TEST (missing tests)

**File:** `pkg/importer/mitmproxy_test.go:584-601`

**Issue:** First-byte dispatch has whitespace and quote coverage (`TestMitmproxyImporter_WhitespaceOnlyInput`, `TestMitmproxyImporter_InvalidFirstToken` with `"string value"`). Not covered: UTF-8 BOM (`0xEF 0xBB 0xBF`) prefix on an otherwise-valid JSON file, and a leading minus sign (`-1:x,`). Both should hit the "unrecognized format" path, but nothing pins that.

**Recommendation:** Add `TestMitmproxyImporter_BOMPrefixedInput` and a leading-`-` case to the existing invalid-first-token table.

---

### TEST-004 — low — TEST (missing tests)

**File:** `pkg/importer/mitmproxy_test.go:1131-1174`

**Issue:** `TestMitmproxyImporter_Native_MaxFlowsCap` tests only above-cap rejection. The exact boundary (`==N` accepted, `==N+1` rejected) is not pinned — unlike the element-count cap which has paired below-cap-accepted / above-cap-rejected tests.

**Recommendation:** Add a paired at-cap test:

```go
func TestMitmproxyImporter_Native_MaxFlowsCap_AtCap(t *testing.T) {
    withTempCap(t, &maxNativeFlows, 3)
    // stream exactly 3 valid flows, assert NoError + len(requests)==3
}
```

---

### TEST-005 — nit — TEST (missing tests)

**File:** `pkg/importer/tnetstring_test.go:148-174`

**Issue:** Malformed-input table covers non-digit prefix, missing colon, empty length, truncated payload, missing type byte, unknown type byte, prefix-too-long, and nested-length-exceeds-buffer. An explicit negative-length case (`-1:x,`) is not present — structurally covered by "non-digit prefix" (`a:hello,`) but not documented.

**Recommendation:** Add `{"negative length", "-1:x,"}` to `TestTnetstring_InvalidInputs`.

---

### TEST-006 — nit — TEST (missing tests)

**File:** `pkg/importer/tnetstring_test.go:164`

**Issue:** `maxLengthPrefixDigits = 9` prevents `strconv.Atoi` overflow by construction. `TestTnetstring_InvalidInputs` "prefix too long" uses 11 digits, so the exact boundary (9 digits accepted, 10 rejected) is not pinned. A future refactor that moved the digit check could slip through.

**Recommendation:** Add `TestTnetstring_LengthPrefixAtDigitCap` with paired 9-digit-accepted and 10-digit-rejected cases.

---

### TEST-007 — medium — TEST (missing tests)

**File:** `pkg/importer/tnetstring_test.go:272-302`

**Issue:** Rubric asks for paired at-cap acceptance / cap+1 rejection for every hard cap. Current coverage:
- `maxTnetstringElement` (64MB): `TestTnetstring_LengthTooLarge` tests cap+1 rejection; no at-cap acceptance.
- `maxTnetstringDepth` (64): `TestTnetstring_DepthLimitRejected` tests depth=128 rejection; no at-cap acceptance (depth=64 accepted, depth=65 rejected).

A refactor that flipped `>` to `>=` (or vice versa) would silently pass because only the far-side rejection is exercised.

**Recommendation:**
- Add `TestTnetstring_DepthAtCapAccepted` (depth=`maxTnetstringDepth` accepted) paired with `TestTnetstring_DepthAtCapPlusOneRejected` (depth=`maxTnetstringDepth+1` rejected).
- For the 64MB element cap, either gate a real-size test behind `testing.Short()` skip, or promote `maxTnetstringElement` to a mutable var so `withTempCap` can lower it and the boundary can be exercised quickly (mirrors the pattern already used for `maxTnetstringElements` and `maxNativeFlows`).

```go
func TestTnetstring_LengthTooLarge(t *testing.T) {
    input := fmt.Sprintf("%d:x,", maxTnetstringElement+1)
    nest := 2 * maxTnetstringDepth
```

---

## Non-blocking

### QUAL-001 — nit — QUAL

**File:** `pkg/importer/mitmproxy.go:107`

**Issue:** Verification note — the `//nolint:gocyclo // format parsing` directive on `importJSON` matches `.golangci.yml` (the project uses `gocyclo`, not `cyclop`). Annotation is correct; no action required. Recorded so a future reviewer sees the check was performed.

---

### QUAL-002 — low — QUAL

**File:** `pkg/importer/mitmproxy.go:108-119`

**Issue:** `Import()` peeks the first byte then dispatches to `importJSON()`, which calls `peekFirstNonWhitespace()` a second time on the same `bufReader`. The second call works because `UnreadByte()` left the byte available, but it's redundant. The inner `switch` has no `default` case — if a non-`[`/`{` byte ever reached `importJSON` (theoretically impossible given the caller guard) it would silently return `(nil, nil)`.

**Recommendation:** Pass `firstByte` as a parameter to `importJSON`, and add a `default` case that returns `fmt.Errorf("mitmproxy importer: importJSON called with unexpected first byte %q", firstByte)`.

---

### QUAL-003 — low — QUAL

**File:** `pkg/importer/mitmproxy.go:298-303`

**Issue:** For a non-CONNECT request with empty `path` and non-empty `authority`, `buildRequestPath` returns `authority` as the URL path, producing a malformed URL like `https://example.com/example.com:80`. Comment says this is deliberate ("fall back to authority so downstream URL parsing surfaces the data rather than dropping it"), but silently surfacing a malformed URL is arguably worse than an import error. No test covers this edge case.

**Recommendation:** Either return `/` (consistent with CONNECT handling) or add a test that pins the current behavior so it's documented as known-weird.

```go
// Non-CONNECT request missing a path is unusual; fall back to authority
// so downstream URL parsing surfaces the data rather than dropping it.
if authority := tnetBytesOrString(reqMap["authority"]); authority != "" {
    return authority
}
```

---

### QUAL-004 — nit — QUAL

**File:** `test/run-live-tests.sh:1280`

**Issue:** `test_import_mitmproxy_native` uses the same `expected-mitmproxy-capture.json` as `test_import_mitmproxy`. This works because both paths produce identical `ObservedRequest` output, but the sharing is implicit — a reader might not realize changes to either could break the other.

**Recommendation:** Add a one-line comment above the `expected_capture` assignment noting the deliberate sharing, or split into two fixture files.

---

### QUAL-005 — nit — QUAL

**File:** `pkg/importer/testhelpers_test.go:43-56`

**Issue:** `tnetListElement` and `tnetDictElement` build tnetstrings via `fmt.Sprintf("%d:%s]", len(body), body)` where `body` is `[]byte` — the exact pattern the production encoder deliberately avoided (bytes containing `%` break silently). Current call sites pass only ASCII-safe bytes so no test fails, but the helper is a latent bug for future callers composing payloads that include binary bodies or percent-encoded paths.

**Recommendation:** Switch the container helpers to explicit byte concatenation, mirroring `tnetenc.encodeWithMarker`.

```go
return []byte(fmt.Sprintf("%d:%s]", len(body), body))
return []byte(fmt.Sprintf("%d:%s}", len(body), body))
```

---

### QUAL-006 — low — QUAL

**File:** `pkg/importer/tnetstring.go:57-60`

**Issue:** `maxTnetstringElements` is a package-level `var` mutated by `withTempCap` in tests. "Production callers MUST treat this as read-only" is in `tnetstring.go`; "NOT PARALLEL-SAFE" is in `testhelpers_test.go`. The warnings are easy to miss because they live in separate files, and the pattern would race if any future test accidentally calls `t.Parallel()`. No current test uses `t.Parallel()`, so this is safe today.

**Recommendation:** Add a same-file comment on `maxTnetstringElements` (and `maxNativeFlows` in `mitmproxy.go`) pointing at the parallel-safety constraint.

---

### SEC-BE-002 — low — SEC-BE

**File:** `pkg/importer/mitmproxy.go:259`
**CVSS:** `CVSS:3.1/AV:L/AC:H/PR:N/UI:R/S:U/C:N/I:L/A:N` (base 2.5, normalized: `C:L → I:L`)

**Issue:** `reqMap["host"]` (attacker-controlled up to the 64MB per-element cap) is written directly into `url.URL.Host`. No length cap (RFC 1035 says 253), no hostname validation, no rejection of userinfo (`user:pass@attacker.example`), embedded ports, IPv6 bracket confusion, or control bytes. Risk is low because `ObservedRequest` is not auto-refetched today, but the field becomes a foot-gun for any future consumer (rate limiter, scope filter, logger) that trusts the host-derived slice. CVSS pass normalized the impact to `I:L` (integrity of URL data) rather than `C:L`, since the described harm is downstream filter/rate-limiter confusion, not information disclosure.

**Recommendation:** Validate host at the importer boundary: reject empty, cap length to 253, reject `@`, reject whitespace/control bytes. Apply in both native and JSON paths for parity.

---

### SEC-BE-003 — low — SEC-BE

**File:** `pkg/importer/mitmproxy.go:335-353`
**CVSS:** `CVSS:3.1/AV:L/AC:H/PR:N/UI:R/S:U/C:N/I:N/A:L` (base 2.5)

**Issue:** `nativeContent` base64-encodes the raw body only so `parseFlow`'s shared `decodeContent` can base64-decode it back. For a 64MB body, transient memory peaks at raw(64MB) + base64(~85MB) + decoded(64MB) ~= 213MB. Flows are processed sequentially so the spike does not accumulate, but peak per-flow is unnecessarily high on memory-constrained hosts.

**Recommendation:** Pass raw `[]byte` directly to `parseFlow`. Either change `mitmproxyRequest.Content`/`mitmproxyResponse.Content` from `*string` to `[]byte` for both paths (JSON path would decode once up front), or add a native-only parse path that skips the base64 round-trip.

---

### SEC-BE-004 — low — SEC-BE

**File:** `pkg/importer/tnetstring.go:99-102`
**CVSS:** `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:L` (base 3.3)

**Issue:** `decodeTnetstringStream` allocates `payload := make([]byte, length)` where `length` is bounded only by `maxTnetstringElement` (64MB), not by bytes actually remaining in the stream. An ~11-byte prefix like `67108864:` followed by immediate EOF forces a 64MB transient allocation before `io.ReadFull` fails. Single-import nuisance; becomes a memory-amplification primitive if the importer is ever wrapped behind a server endpoint accepting concurrent uploads. The inner decoder (`decodeTnetstringInner`) bounds length by `r.Len()` so it is not affected.

**Recommendation:** Either bound the stream-path length by remaining bytes where known, or grow the payload buffer in 1MB chunks so a truncated claim aborts before full allocation.

```go
length, err := readLengthPrefix(r, maxTnetstringElement)
if err != nil {
    return nil, err
}
payload := make([]byte, length)
if _, err := io.ReadFull(r, payload); err != nil {
    return nil, fmt.Errorf("tnetstring: read payload (len=%d): %w", length, err)
```

---

### SEC-BE-005 — nit — SEC-BE

**File:** `pkg/importer/tnetstring.go:191-194`
**CVSS:** `CVSS:3.1/AV:L/AC:H/PR:N/UI:R/S:U/C:N/I:N/A:N` (base 0.0)

**Issue:** `parseTnetPayload`'s `tnetBytes` case does `buf := make([]byte, len(payload)); copy(buf, payload)` even though both decoder paths allocate `payload` via fresh `make([]byte, length)`. The copy doubles transient memory for no aliasing benefit.

**Recommendation:** Return `payload` directly. Add a comment (or test assertion) confirming the decoder paths never share the payload slice with a parent buffer.

---

### TEST-003 — low — TEST

**File:** `pkg/importer/mitmproxy_test.go:806-847`

**Issue:** `TestMitmproxyImporter_Native_RegressionLAB2309` asserts the fixture starts with a digit and decodes cleanly. The comment mentions the pre-fix error message (`"expected JSON array or object, got \"2\""`) but the test doesn't pin it (via `NotContains` or equivalent). A future dispatch rewrite that accidentally succeeds on digit-prefixed input via an unrelated path would not trip this guard.

**Recommendation:** Either add `assert.NotContains(t, err.Error(), "expected JSON array or object")` on the error-free success path as defense-in-depth, or document the manual red-green verification in a comment (commit SHA of the reverted dispatch, date).

---

### DOC-001 — nit — DOC

**File:** `CLAUDE.md:75`

**Issue:** Phase 6 heuristic: new exported symbols (`internal/tnetenc.Encode`, `pkg/importer.ErrTnetstringDepth`) were added, and `CLAUDE.md`'s `pkg/importer` bullet still describes caps as just "size limits (500MB)". The PR adds per-element (64MB), list/dict cardinality (1M), and flow (500K) caps. `pkg/importer/doc.go` was correctly updated — only the project-overview doc is stale.

**Recommendation:** Update the `pkg/importer` bullet in `CLAUDE.md` (line 75) to mention per-element and cardinality caps alongside the 500MB file cap.

---

## Ticket alignment

_None._ All three LAB-2309 acceptance criteria are implemented — see the summary above.

---

## Tests & documentation

(TEST and DOC findings are enumerated above, interleaved with their severity and blocker status.)

---

5 CodeRabbit bot review comments (dict-pair counting, error-payload truncation, key-existence ordering, CONNECT URL shape, vendored-fixture provenance) are all confirmed fixed in HEAD SHA `094972f` and are filtered from this roll-up.

---

## Notes for the engineer

- The PR is substantively strong: 92.9% pkg coverage, 100% on `internal/tnetenc`, a real vendored fixture, a drift check between generator and committed fixture, and `make check` green. The blockers below are mostly hardening + at-cap boundary tests — none indicate the feature is broken.
- **SEC-BE-001** (scheme whitelist) is the one finding I would not merge without addressing. It is a genuine new attack surface introduced by accepting untrusted `.mitm` bytes without constraining the URL scheme the rest of the codebase will see.
- Several TEST blockers (TEST-005, TEST-006) are nit-severity but the skill's blocker formula treats any missing-test finding as a blocker. The engineer should use judgment — these are cheap additions but not strictly required.
- Non-blocker security findings (SEC-BE-002 through SEC-BE-005) are consistent cleanup opportunities; worth a short follow-up ticket so they aren't lost.

---

_Review generated by `capability-pr-review` (with agent substitutions noted above). Rerun against the same PR head SHA + Linear ticket state to reproduce this finding set._

eli-wald (changes requested):

**PR:** https://github.com/praetorian-inc/vespasian/pull/89
**Linear:** https://linear.app/praetorianlabs/issue/LAB-2309/external-issue-mitmproxy-import-doesnt-work
**Head SHA:** `094972f935b4bce79146021125c9192e10109708`
**Reviewed at:** 2026-04-22

Main takeaways are to implement some additional scheme checking in `pkg/importer/mitmproxy.go` and add some additional test coverage. Looks good otherwise.

| Totals   | critical | high | medium | low | nit |
|----------|----------|------|--------|-----|-----|
| Count    | 0        | 0    | 2      | 10  | 7   |
| Blockers | 0        | 0    | 2      | 3   | 2   |

Acceptance criteria (LAB-2309): **all 3 implemented.**
- AC1 — native `.mitm` import works via first-byte dispatch into a tnetstring decoder. ✓
- AC2 — JSON export path preserved. ✓ (`TestMitmproxyImporter_*_JSON` still passes)
- AC3 — regression test `TestMitmproxyImporter_Native_RegressionLAB2309` asserts the ASCII-digit failure mode decodes cleanly. ✓

CodeRabbit bot left 5 earlier comments. All 5 are confirmed fixed in HEAD (`pairsParsed` counter replaces `len(result)`; `payloadPreview` truncates errors; key-existence asserts precede order checks; CONNECT URL shape characterized with tests; vendored fixture commit-pinned). Those findings are filtered out of this roll-up.

---

## Blockers

**File:** `pkg/importer/mitmproxy.go:258-264`
**CVSS:** `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:C/C:L/I:L/A:N` (base 5.0)

**Issue:** `flow.Request.Scheme` is taken verbatim from `reqMap["scheme"]` and passed to `constructURL` with no whitelist. mitmproxy itself only emits `http`/`https`, but the new native path accepts untrusted `.mitm` files (per the PR's own threat model). An attacker can forge scheme=`file`, `javascript`, `gopher`, `data`, `ldap`, `dict`, etc. The resulting `ObservedRequest.URL` carries that scheme into any downstream consumer that re-fetches, replays, or inspects it. Enforcing at the importer boundary removes a whole class of scheme-confusion bugs in downstream code that is not documented as having to re-validate.

**Recommendation:** In `flowFromNativeState` (and for parity in `importJSON.parseFlow`), reject flows whose scheme is not `http` or `https`:

```go
if scheme := tnetBytesOrString(reqMap["scheme"]); scheme != "http" && scheme != "https" {
    return mitmproxyFlow{}, fmt.Errorf("unsupported scheme %q", scheme)
}
```

```go
req := mitmproxyRequest{
    Method:  method,
    Scheme:  tnetBytesOrString(reqMap["scheme"]),
    Host:    tnetBytesOrString(reqMap["host"]),
    Port:    port,
    Path:    buildRequestPath(reqMap, method),
```

---

### TEST-001 — low — TEST (missing tests)

**File:** `pkg/importer/mitmproxy_fixture_test.go:61-86`

**Issue:** `TestMitmproxyFixture_MatchesGenerator` byte-compares the committed `sample-mitmproxy.mitm` against generator stdout, proving `committed == generated`. But no unit test actually feeds those bytes through `MitmproxyImporter.Import`. The end-to-end path for the generated fixture is only exercised in the gated live test `import-mitmproxy-native`. A generator that silently produces unparseable bytes would pass the drift check.

**Recommendation:** Add `TestMitmproxyImporter_Native_GeneratedFixture` that reads `test/fixtures/sample-mitmproxy.mitm`, imports it, and asserts on the three known flows.

---

### TEST-002 — low — TEST (missing tests)

**File:** `pkg/importer/mitmproxy_test.go:584-601`

**Issue:** First-byte dispatch has whitespace and quote coverage (`TestMitmproxyImporter_WhitespaceOnlyInput`, `TestMitmproxyImporter_InvalidFirstToken` with `"string value"`). Not covered: UTF-8 BOM (`0xEF 0xBB 0xBF`) prefix on an otherwise-valid JSON file, and a leading minus sign (`-1:x,`). Both should hit the "unrecognized format" path, but nothing pins that.

**Recommendation:** Add `TestMitmproxyImporter_BOMPrefixedInput` and a leading-`-` case to the existing invalid-first-token table.

---

### TEST-004 — low — TEST (missing tests)

**File:** `pkg/importer/mitmproxy_test.go:1131-1174`

**Issue:** `TestMitmproxyImporter_Native_MaxFlowsCap` tests only above-cap rejection. The exact boundary (`==N` accepted, `==N+1` rejected) is not pinned — unlike the element-count cap which has paired below-cap-accepted / above-cap-rejected tests.

**Recommendation:** Add a paired at-cap test:

```go
func TestMitmproxyImporter_Native_MaxFlowsCap_AtCap(t *testing.T) {
    withTempCap(t, &maxNativeFlows, 3)
    // stream exactly 3 valid flows, assert NoError + len(requests)==3
}
```

---

### TEST-005 — nit — TEST (missing tests)

**File:** `pkg/importer/tnetstring_test.go:148-174`

**Issue:** Malformed-input table covers non-digit prefix, missing colon, empty length, truncated payload, missing type byte, unknown type byte, prefix-too-long, and nested-length-exceeds-buffer. An explicit negative-length case (`-1:x,`) is not present — structurally covered by "non-digit prefix" (`a:hello,`) but not documented.

**Recommendation:** Add `{"negative length", "-1:x,"}` to `TestTnetstring_InvalidInputs`.

---

### TEST-006 — nit — TEST (missing tests)

**File:** `pkg/importer/tnetstring_test.go:164`

**Issue:** `maxLengthPrefixDigits = 9` prevents `strconv.Atoi` overflow by construction. `TestTnetstring_InvalidInputs` "prefix too long" uses 11 digits, so the exact boundary (9 digits accepted, 10 rejected) is not pinned. A future refactor that moved the digit check could slip through.

**Recommendation:** Add `TestTnetstring_LengthPrefixAtDigitCap` with paired 9-digit-accepted and 10-digit-rejected cases.

---

### TEST-007 — medium — TEST (missing tests)

**File:** `pkg/importer/tnetstring_test.go:272-302`

**Issue:** Rubric asks for paired at-cap acceptance / cap+1 rejection for every hard cap. Current coverage:
- `maxTnetstringElement` (64MB): `TestTnetstring_LengthTooLarge` tests cap+1 rejection; no at-cap acceptance.
- `maxTnetstringDepth` (64): `TestTnetstring_DepthLimitRejected` tests depth=128 rejection; no at-cap acceptance (depth=64 accepted, depth=65 rejected).

A refactor that flipped `>` to `>=` (or vice versa) would silently pass because only the far-side rejection is exercised.

**Recommendation:**
- Add `TestTnetstring_DepthAtCapAccepted` (depth=`maxTnetstringDepth` accepted) paired with `TestTnetstring_DepthAtCapPlusOneRejected` (depth=`maxTnetstringDepth+1` rejected).
- For the 64MB element cap, either gate a real-size test behind `testing.Short()` skip, or promote `maxTnetstringElement` to a mutable var so `withTempCap` can lower it and the boundary can be exercised quickly (mirrors the pattern already used for `maxTnetstringElements` and `maxNativeFlows`).

```go
func TestTnetstring_LengthTooLarge(t *testing.T) {
    input := fmt.Sprintf("%d:x,", maxTnetstringElement+1)
    nest := 2 * maxTnetstringDepth
```

---

## Non-blocking

### QUAL-001 — nit — QUAL

**File:** `pkg/importer/mitmproxy.go:107`

**Issue:** Verification note — the `//nolint:gocyclo // format parsing` directive on `importJSON` matches `.golangci.yml` (the project uses `gocyclo`, not `cyclop`). Annotation is correct; no action required. Recorded so a future reviewer sees the check was performed.

---

### QUAL-002 — low — QUAL

**File:** `pkg/importer/mitmproxy.go:108-119`

**Issue:** `Import()` peeks the first byte then dispatches to `importJSON()`, which calls `peekFirstNonWhitespace()` a second time on the same `bufReader`. The second call works because `UnreadByte()` left the byte available, but it's redundant. The inner `switch` has no `default` case — if a non-`[`/`{` byte ever reached `importJSON` (theoretically impossible given the caller guard) it would silently return `(nil, nil)`.

**Recommendation:** Pass `firstByte` as a parameter to `importJSON`, and add a `default` case that returns `fmt.Errorf("mitmproxy importer: importJSON called with unexpected first byte %q", firstByte)`.

---

### QUAL-003 — low — QUAL

**File:** `pkg/importer/mitmproxy.go:298-303`

**Issue:** For a non-CONNECT request with empty `path` and non-empty `authority`, `buildRequestPath` returns `authority` as the URL path, producing a malformed URL like `https://example.com/example.com:80`. Comment says this is deliberate ("fall back to authority so downstream URL parsing surfaces the data rather than dropping it"), but silently surfacing a malformed URL is arguably worse than an import error. No test covers this edge case.

**Recommendation:** Either return `/` (consistent with CONNECT handling) or add a test that pins the current behavior so it's documented as known-weird.

```go
// Non-CONNECT request missing a path is unusual; fall back to authority
// so downstream URL parsing surfaces the data rather than dropping it.
if authority := tnetBytesOrString(reqMap["authority"]); authority != "" {
    return authority
}
```

---

### QUAL-004 — nit — QUAL

**File:** `test/run-live-tests.sh:1280`

**Issue:** `test_import_mitmproxy_native` uses the same `expected-mitmproxy-capture.json` as `test_import_mitmproxy`. This works because both paths produce identical `ObservedRequest` output, but the sharing is implicit — a reader might not realize changes to either could break the other.

**Recommendation:** Add a one-line comment above the `expected_capture` assignment noting the deliberate sharing, or split into two fixture files.

---

### QUAL-005 — nit — QUAL

**File:** `pkg/importer/testhelpers_test.go:43-56`

**Issue:** `tnetListElement` and `tnetDictElement` build tnetstrings via `fmt.Sprintf("%d:%s]", len(body), body)` where `body` is `[]byte` — the exact pattern the production encoder deliberately avoided (bytes containing `%` break silently). Current call sites pass only ASCII-safe bytes so no test fails, but the helper is a latent bug for future callers composing payloads that include binary bodies or percent-encoded paths.

**Recommendation:** Switch the container helpers to explicit byte concatenation, mirroring `tnetenc.encodeWithMarker`.

```go
return []byte(fmt.Sprintf("%d:%s]", len(body), body))
return []byte(fmt.Sprintf("%d:%s}", len(body), body))
```

---

### QUAL-006 — low — QUAL

**File:** `pkg/importer/tnetstring.go:57-60`

**Issue:** `maxTnetstringElements` is a package-level `var` mutated by `withTempCap` in tests. "Production callers MUST treat this as read-only" is in `tnetstring.go`; "NOT PARALLEL-SAFE" is in `testhelpers_test.go`. The warnings are easy to miss because they live in separate files, and the pattern would race if any future test accidentally calls `t.Parallel()`. No current test uses `t.Parallel()`, so this is safe today.

**Recommendation:** Add a same-file comment on `maxTnetstringElements` (and `maxNativeFlows` in `mitmproxy.go`) pointing at the parallel-safety constraint.

---

### SEC-BE-002 — low — SEC-BE

**File:** `pkg/importer/mitmproxy.go:259`
**CVSS:** `CVSS:3.1/AV:L/AC:H/PR:N/UI:R/S:U/C:N/I:L/A:N` (base 2.5, normalized: `C:L → I:L`)

**Issue:** `reqMap["host"]` (attacker-controlled up to the 64MB per-element cap) is written directly into `url.URL.Host`. No length cap (RFC 1035 says 253), no hostname validation, no rejection of userinfo (`user:pass@attacker.example`), embedded ports, IPv6 bracket confusion, or control bytes. Risk is low because `ObservedRequest` is not auto-refetched today, but the field becomes a foot-gun for any future consumer (rate limiter, scope filter, logger) that trusts the host-derived slice. CVSS pass normalized the impact to `I:L` (integrity of URL data) rather than `C:L`, since the described harm is downstream filter/rate-limiter confusion, not information disclosure.

**Recommendation:** Validate host at the importer boundary: reject empty, cap length to 253, reject `@`, reject whitespace/control bytes. Apply in both native and JSON paths for parity.

---

### SEC-BE-003 — low — SEC-BE

**File:** `pkg/importer/mitmproxy.go:335-353`
**CVSS:** `CVSS:3.1/AV:L/AC:H/PR:N/UI:R/S:U/C:N/I:N/A:L` (base 2.5)

**Issue:** `nativeContent` base64-encodes the raw body only so `parseFlow`'s shared `decodeContent` can base64-decode it back. For a 64MB body, transient memory peaks at raw(64MB) + base64(~85MB) + decoded(64MB) ~= 213MB. Flows are processed sequentially so the spike does not accumulate, but peak per-flow is unnecessarily high on memory-constrained hosts.

**Recommendation:** Pass raw `[]byte` directly to `parseFlow`. Either change `mitmproxyRequest.Content`/`mitmproxyResponse.Content` from `*string` to `[]byte` for both paths (JSON path would decode once up front), or add a native-only parse path that skips the base64 round-trip.

---

### SEC-BE-004 — low — SEC-BE

**File:** `pkg/importer/tnetstring.go:99-102`
**CVSS:** `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:L` (base 3.3)

**Issue:** `decodeTnetstringStream` allocates `payload := make([]byte, length)` where `length` is bounded only by `maxTnetstringElement` (64MB), not by bytes actually remaining in the stream. An ~11-byte prefix like `67108864:` followed by immediate EOF forces a 64MB transient allocation before `io.ReadFull` fails. Single-import nuisance; becomes a memory-amplification primitive if the importer is ever wrapped behind a server endpoint accepting concurrent uploads. The inner decoder (`decodeTnetstringInner`) bounds length by `r.Len()` so it is not affected.

**Recommendation:** Either bound the stream-path length by remaining bytes where known, or grow the payload buffer in 1MB chunks so a truncated claim aborts before full allocation.

```go
length, err := readLengthPrefix(r, maxTnetstringElement)
if err != nil {
    return nil, err
}
payload := make([]byte, length)
if _, err := io.ReadFull(r, payload); err != nil {
    return nil, fmt.Errorf("tnetstring: read payload (len=%d): %w", length, err)
```

---

### SEC-BE-005 — nit — SEC-BE

**File:** `pkg/importer/tnetstring.go:191-194`
**CVSS:** `CVSS:3.1/AV:L/AC:H/PR:N/UI:R/S:U/C:N/I:N/A:N` (base 0.0)

**Issue:** `parseTnetPayload`'s `tnetBytes` case does `buf := make([]byte, len(payload)); copy(buf, payload)` even though both decoder paths allocate `payload` via fresh `make([]byte, length)`. The copy doubles transient memory for no aliasing benefit.

**Recommendation:** Return `payload` directly. Add a comment (or test assertion) confirming the decoder paths never share the payload slice with a parent buffer.

---

### TEST-003 — low — TEST

**File:** `pkg/importer/mitmproxy_test.go:806-847`

**Issue:** `TestMitmproxyImporter_Native_RegressionLAB2309` asserts the fixture starts with a digit and decodes cleanly. The comment mentions the pre-fix error message (`"expected JSON array or object, got \"2\""`) but the test doesn't pin it (via `NotContains` or equivalent). A future dispatch rewrite that accidentally succeeds on digit-prefixed input via an unrelated path would not trip this guard.

**Recommendation:** Either add `assert.NotContains(t, err.Error(), "expected JSON array or object")` on the error-free success path as defense-in-depth, or document the manual red-green verification in a comment (commit SHA of the reverted dispatch, date).

---

### DOC-001 — nit — DOC

**File:** `CLAUDE.md:75`

**Issue:** Phase 6 heuristic: new exported symbols (`internal/tnetenc.Encode`, `pkg/importer.ErrTnetstringDepth`) were added, and `CLAUDE.md`'s `pkg/importer` bullet still describes caps as just "size limits (500MB)". The PR adds per-element (64MB), list/dict cardinality (1M), and flow (500K) caps. `pkg/importer/doc.go` was correctly updated — only the project-overview doc is stale.

**Recommendation:** Update the `pkg/importer` bullet in `CLAUDE.md` (line 75) to mention per-element and cardinality caps alongside the 500MB file cap.

---

## Ticket alignment

_None._ All three LAB-2309 acceptance criteria are implemented — see the summary above.

---

## Tests & documentation

(TEST and DOC findings are enumerated above, interleaved with their severity and blocker status.)

---

5 CodeRabbit bot review comments (dict-pair counting, error-payload truncation, key-existence ordering, CONNECT URL shape, vendored-fixture provenance) are all confirmed fixed in HEAD SHA `094972f` and are filtered from this roll-up.

---

## Notes for the engineer

- The PR is substantively strong: 92.9% pkg coverage, 100% on `internal/tnetenc`, a real vendored fixture, a drift check between generator and committed fixture, and `make check` green. The blockers below are mostly hardening + at-cap boundary tests — none indicate the feature is broken.
- **SEC-BE-001** (scheme whitelist) is the one finding I would not merge without addressing. It is a genuine new attack surface introduced by accepting untrusted `.mitm` bytes without constraining the URL scheme the rest of the codebase will see.
- Several TEST blockers (TEST-005, TEST-006) are nit-severity but the skill's blocker formula treats any missing-test finding as a blocker. The engineer should use judgment — these are cheap additions but not strictly required.
- Non-blocker security findings (SEC-BE-002 through SEC-BE-005) are consistent cleanup opportunities; worth a short follow-up ticket so they aren't lost.

---

_Review generated by `capability-pr-review` (with agent substitutions noted above). Rerun against the same PR head SHA + Linear ticket state to reproduce this finding set._

eli-wald:
# Praetorian Capability PR Review (round 10)

**PR:** https://github.com/praetorian-inc/vespasian/pull/89
**Linear:** https://linear.app/praetorianlabs/issue/LAB-2309/external-issue-mitmproxy-import-doesnt-work
**Prior review SHA:** `094972f935b4bce79146021125c9192e10109708` (submitted 2026-04-22T15:08Z)
**This review SHA:** `6cbb8a5c6e1c5d12e45f18c0f35133e2b17a871b`
**Delta:** +907 / −147 across 9 files (rounds 6, 7, 8, 9 commits)

Looks good, the only thing that didn't get surfaced before was a URL encoding bug.

### Progress since the prior review

Between `094972f` and `6cbb8a5` the author ran 4 dedicated fix-round commits. **18 prior-round findings are confirmed addressed** — the `.review/addressed.json` entries plus a manual pass against each item in my prior review:

| Category | Finding | Status at 6cbb8a5 |
|---|---|---|
| QUAL-002 | importJSON re-peeks first byte | FIXED — firstByte passed as parameter + default case added |
| QUAL-003 | buildRequestPath non-CONNECT empty-path | FIXED — now returns "/" with doc comment + test pins it |
| QUAL-004 | live test shared fixture undocumented | FIXED — sharing documented at shell line 1280 |
| QUAL-005 | testhelpers %s on []byte | FIXED — `buildElement` uses explicit byte concatenation with a note on why %s is unsafe |
| QUAL-006 | maxTnetstringElements parallel-safety warning | FIXED — all four mutable caps carry prominent NOT PARALLEL-SAFE doc comments |
| SEC-BE-001 | unvalidated URL scheme | FIXED — `allowedSchemes` map + enforcement at `flowFromNativeState` |
| SEC-BE-002 | host validation | FIXED — `validateHost` checks empty, 253-byte cap, @-userinfo, control bytes |
| SEC-BE-003 | body 3x memory amplification | FIXED — `mitmproxyRequest.Content` is now []byte; `nativeContent` passes raw bytes with no base64 round-trip |
| SEC-BE-004 | stream pre-allocation | FIXED — `readStreamPayload` uses `bytes.Buffer.Grow(streamInitialCap=1MB)` + `io.CopyN`, so truncated claims max at 1MB instead of 64MB |
| SEC-BE-005 | tnetBytes double-copy | FIXED — `parseTnetPayload` case tnetBytes returns payload directly |
| TEST-001 | sample fixture never imported E2E | FIXED — `_GeneratedFixtureImports` |
| TEST-002 | BOM / leading-minus dispatch coverage | FIXED — `_BOMPrefixRejected`, `_LeadingMinusRejected` |
| TEST-003 | regression test pre-fix string pin | Acknowledged with documented limitation — accepted as addressed |
| TEST-004 | maxNativeFlows at-cap | FIXED — `_MaxFlowsCap_AtCap` |
| TEST-005 | negative-length row | FIXED — added to `InvalidInputs` table |
| TEST-006 | digit-cap boundary | FIXED — `_LengthPrefixAtDigitCap` pair |
| TEST-007 | 64MB / depth=64 at-cap pairs | FIXED — `_ElementAtCapAccepted`/`+1Rejected`, `_DepthAtCapAccepted`/`+1Rejected` |
| DOC-001 | CLAUDE.md stale importer caps | FIXED — CLAUDE.md updated (in the diff) |

Plus the 5 CodeRabbit prior-round comments remain filtered (all fixed).

### Ticket alignment (LAB-2309)

All 3 acceptance criteria met — same as prior round, reinforced by the expanded regression/boundary tests.

---

## Blockers

### QUAL-002 — medium — QUAL

**File:** `pkg/importer/mitmproxy.go:531-540`

**Issue:** `constructURL` assigns the wire request target to `u.Path`. `url.URL.String()` calls `EscapedPath()` which percent-encodes `Path` — including any `%` characters already present in an RFC 3986 encoded sequence. A real mitmproxy capture with path `/api/hello%20world` therefore produces `ObservedRequest.URL = "https://example.com/api/hello%2520world"` (double-encoded). mitmproxy stores the verbatim wire request target in the `path` field per its own docs, so pre-encoded paths are the expected input. The capability-reviewer verified this with a standalone Go test. Affects capture.json and every downstream consumer (classifier, probes, spec generator).

**Recommendation:** Set `u.RawPath = pathPart` — per `url.URL` semantics, when `RawPath` is non-empty and well-formed, `String()` uses it directly. If keeping `Path` populated, assign the decoded form. Add a test: import a native flow whose `path` is `/api/hello%20world` and assert `ObservedRequest.URL` contains a single-encoded `%20`.

```go
func constructURL(scheme, host string, port int, path string) string {
    pathPart, query, _ := strings.Cut(path, "?")
    u := &url.URL{
        Path:     pathPart,   // ← should be RawPath for pre-encoded input
        RawQuery: query,
    }
```

---

### TEST-001 — nit — TEST *(missing tests)*

**File:** `pkg/importer/helpers.go:27`

**Issue:** Round 8/9 added bounded-error tests (`_LargeMethodBoundedInError`, `_LargeSchemeBoundedInError`, `_InvalidIntHugePayloadBounded`, `_DictHugeKeyErrorBounded`) but they only assert `len(msg) < 256` — a generous upper bound that would still pass if `maxPreviewLen` silently dropped to 32 or grew to 128. Neither the exact length of the preview nor the at-cap vs above-cap branch of `previewString` / `payloadPreview` is pinned.

**Recommendation:** Add focused unit tests on both helpers: (a) input of exactly `maxPreviewLen` bytes returns the full content with NO truncation suffix; (b) `maxPreviewLen+1` bytes yields exactly `maxPreviewLen` bytes of quoted body + the `... (N bytes total)` annotation with the correct total. Parameterize on the constant, not the literal 64.

---

### TEST-002 — low — TEST *(missing tests)*

**File:** `pkg/importer/tnetstring_test.go:501-525`

**Issue:** `parseTnetList` and `parseTnetDict` use `if len(result) >= maxTnetstringElements`, so the maximum accepted cardinality is `cap-1`. `TestTnetstring_ListElementCountCap_BelowCapAccepted` uses n=100 against the default 1M cap — 10,000× below the boundary — and does not pin the exact at-cap behavior. `_AboveCapRejected` lowers the cap to 10 and emits `lowered+5`, proving rejection above the cap but not that `cap-1` is accepted and `cap` is rejected. A `>=` → `>` refactor (off-by-one loosening) would increase capacity by 1 and still pass every existing test. Both the list and dict cap sites need paired coverage.

**Recommendation:** Add paired tests (one per container kind):

```go
withTempCap(t, &maxTnetstringElements, 10)
// emit exactly 9 entries → assert NoError
// emit exactly 10 entries → assert "list exceeds"/"dict exceeds"
```

Pins the exact `>=` semantics and catches `>` / `>=` refactors.

```go
if len(result) >= maxTnetstringElements {
    return nil, fmt.Errorf("tnetstring: list exceeds %d elements", maxTnetstringElements)
```

---

## Non-blocking

### QUAL-001 — nit — QUAL

**File:** `pkg/importer/mitmproxy.go:113`

**Issue:** `importJSON` still carries `//nolint:gocyclo // format parsing` at complexity 14. The `[` and `{` branches are independent ~15-line sections that could extract into `parseJSONArray` / `parseJSONObject`, reducing `importJSON` to a ~10-line dispatcher. The annotation is correct for the current linter config; decomposition is simply the cleaner alternative. The author has chosen the suppression path — note this only as a style option, not a correctness issue.

**Recommendation:** Optional — extract the two branches and drop the nolint.

---

### QUAL-003 — nit — QUAL

**File:** `pkg/importer/tnetstring.go:197`

**Issue:** `parseTnetPayload` (complexity 14) and `readLengthPrefix` (complexity 11) both exceed the typical threshold of 10, yet neither has a `//nolint:gocyclo` annotation. `importJSON` in mitmproxy.go already has one at the same complexity level. The complexity here is intrinsic (type dispatch, length-prefix parsing) so decomposition is harder to justify than for `importJSON`. Passes linting today; would trip if the project's `golangci-lint` threshold tightens.

**Recommendation:** For consistency with `importJSON`, either add `//nolint:gocyclo // tnetstring type dispatch` and `//nolint:gocyclo // length prefix validation`, or extract the scalar-type cases in `parseTnetPayload` into a `parseScalarPayload` helper.

---

### SEC-BE-001 — low — SEC-BE

**File:** `pkg/importer/helpers.go:27` (constant) + `pkg/importer/mitmproxy.go:381-386` (consumer)
**CVSS:** `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:L/A:N` (base 3.5)

**Issue:** `maxPreviewLen = 64` caps the LENGTH of attacker bytes in error messages, but the two consumers diverge in rendering. `payloadPreview` (tnetstring.go:343) uses `%q`, which Go-escapes control bytes (`\x1b`, `\x00`, etc.). `previewString` (mitmproxy.go:381) uses `%s`, rendering raw bytes verbatim into error strings that flow through `fmt.Errorf` at mitmproxy.go:489 (invalid method) and :493 (unsupported scheme). A crafted `.mitm` flow with method or scheme containing up to 63 bytes of ANSI escape sequences — e.g. `method="FOO\x1b[2J\x1b[H"` — gets echoed to the operator's terminal / CI log and can clear the screen, recolor output, move the cursor, or poison log parsers. The PR's stated threat class is "bound attacker-controlled payload in all tnetstring error paths" (094972f); bounding length without escaping content leaves the content-shaping half of that class open. `payloadPreview` already demonstrates the correct pattern.

**Recommendation:** Make `previewString` %q-quote its output to match `payloadPreview`:

```go
func previewString(s string) string {
    if len(s) <= maxPreviewLen {
        return fmt.Sprintf("%q", s)
    }
    return fmt.Sprintf("%q... (%d bytes total)", s[:maxPreviewLen], len(s))
}
```

Update the helpers.go doc comment (which already acknowledges "separate signatures") to document the now-aligned quoting discipline. Alternative: collapse both into a single `maxPreviewLen`-aware formatter in helpers.go that always `%q`-quotes.

---

## Ticket alignment

_None._ All three LAB-2309 acceptance criteria are implemented and covered by regression tests. See the "Progress since prior review" table.

---

## Tests & documentation

(TEST-001 and TEST-002 are enumerated in the Blockers section. No new DOC findings — `CLAUDE.md` was updated in this round.)

---

**23 total** — 18 from my prior review + 5 from the earlier CodeRabbit round (all confirmed fixed at `6cbb8a5`). See the table above for per-finding status. Raw evidence in `.review/addressed.json`.

---

## Notes for the engineer

- **The rounds 6–9 delta is notably high-quality.** Every single one of my prior blockers (SEC-BE-001 scheme whitelist, TEST-007 boundary tests, TEST-001/002/004 coverage gaps, TEST-005/006 test pinning) was addressed with a direct, labeled regression test. The `maxPreviewLen = 64` shared constant, the `allowedSchemes` map, the `validateHost` helper, the `readStreamPayload` streaming path, and the `pairsParsed` counter are all thoughtful fixes.
- The one new **medium** finding (QUAL-002 double-encoding) is the only thing I'd not merge without addressing — it corrupts URLs in `capture.json` for any mitmproxy input whose path contains percent-encoded characters, and the fix is a one-line `u.RawPath = pathPart`.
- SEC-BE-001 (`%s` vs `%q`) is aesthetically the same class of finding that motivated 094972f; fixing it closes the content-shaping half of the attacker-bytes-in-errors threat class. Not a blocker at base score 3.5, but a natural round-10 follow-up.
- The two TEST blockers are both of the "pin the exact boundary" variety — cheap to add and they complete the "paired at-cap accept / cap+1 reject" pattern that the rest of the test suite already follows.

---

_Review generated by `capability-pr-review` (with agent substitutions noted above). Rerun against the same PR head SHA + Linear ticket state to reproduce this finding set._

eli-wald (approved):

**PR:** https://github.com/praetorian-inc/vespasian/pull/89
**Linear:** https://linear.app/praetorianlabs/issue/LAB-2309/external-issue-mitmproxy-import-doesnt-work
**Head SHA:** 5fc1e4bb5be0e84874924d38c4798f5498ccb918
**Reviewed at:** 2026-04-24T19:00:35Z

**Recommendation:** COMMENT

This is round 17 against a PR that has already absorbed 16 rounds of review. The serious attack surface — tnetstring parser bounds, URL construction, host/scheme validation, error-disclosure quoting — is solid at HEAD. All three LAB-2309 acceptance criteria are met. The findings below are non-blocking polish: two `nit`s (name-visibility + test-script comment) and two `low`s (a latent test-infrastructure concurrency foot-gun, and a tautological regression test).

| Totals   | critical | high | medium | low | nit |
|----------|----------|------|--------|-----|-----|
| Count    | 0        | 0    | 0      | 2   | 2   |
| Blockers | 0        | 0    | 0      | 0   | 0   |

---

## Blockers

_None._

---

## Non-blocking

### QUAL-001 — nit — QUAL

**File:** `pkg/importer/tnetstring.go:82`

**Issue:** `ErrTnetstringDepth` is declared as an exported var. Both tests that use it reside in package `importer` (not `importer_test`), so they can access the unexported form. The export widens the package surface for an internal sentinel; callers of `Import()` never need `errors.Is` against depth overflow because it is wrapped generically.

**Recommendation:** Lowercase to `errTnetstringDepth`. No external call site references it.

```
var ErrTnetstringDepth = errors.New("tnetstring: nesting depth exceeds limit")
```

---

### QUAL-002 — nit — QUAL

**File:** `test/run-live-tests.sh:1372-1380`

**Issue:** Regex `^3[1-9]$` accepts hex 31..39 (`'1'..'9'`) and rejects 0x30 (`'0'`). The comment describes the guard as excluding `'0'` because mitmproxy never starts a flow stream with `'0:'`, but the error text says "expected an ASCII digit 1-9" which conflates mitmproxy convention with tnetstring validity. A future fixture that begins with `'0'` would fail with a misleading message.

**Recommendation:** Either extend the regex to accept 30..39 and add a comment noting `'0'` is theoretically valid but never produced by mitmproxy, or tighten the comment to explain a real mitmproxy flow dict is always >9 bytes so the leading length digit is always ≥1.

```
if ! [[ "$first_byte" =~ ^3[1-9]$ ]]; then
    log_fail "Fixture first byte is 0x${first_byte:-<empty>}, expected an ASCII digit 1-9 (tnetstring length prefix)"
```

---

### SEC-BE-001 — low — SEC-BE

**File:** `pkg/importer/tnetstring.go:71-140`

**CVSS:** `CVSS:3.1/AV:L/AC:H/PR:H/UI:N/S:U/C:N/I:L/A:L` (2.5)

**Issue:** `maxTnetstringElement`, `maxTnetstringElements`, and `streamInitialCap` are package-level `var` (not `const`) so `withTempCap` can lower them during tests. Each carries a NOT PARALLEL-SAFE comment, and no `t.Parallel()` is used in `pkg/importer` today. This discipline is correct but brittle: a future contributor adding parallel subtests would silently race on these scalars, and a silently-passing race would mask the at-cap behavior the caps exist to enforce. The risk surface is test-infrastructure integrity, not runtime exploit — production code only reads these values.

**Recommendation:** Either (a) change `withTempCap` to mutate a `*sync/atomic.Int64` and teach the parser to load via atomic, making parallel-safety a property of the type rather than a comment; or (b) add a build-tag gate so the cap variables are writeable only in `_test.go` translation units.

```
var maxTnetstringElement = 64 * 1024 * 1024 // 64 MB per-element allocation bound
var maxTnetstringElements = 1 << 20 // 1M entries per list or dict
var streamInitialCap = 1 << 20 // 1 MB
```

---

### TEST-001 — low — TEST

**File:** `pkg/importer/mitmproxy_test.go:1547-1561`

**Issue:** `TestMitmproxyImporter_Native_RegressionLAB2309_ErrorMessagePin` is advertised as a companion that pins the pre-fix error text never appears on the success path, but its body only calls `Import` on a well-formed native flow and asserts `require.NoError`. That is a strict subset of `TestMitmproxyImporter_Native_RegressionLAB2309` at line 817, which already covers this path with richer assertions (method, URL, headers, status, response body, source). The author's own comment acknowledges the body is tautological: "err is nil, so there is no message to inspect."

**Recommendation:** Either delete the test (the line-817 version covers this) or replace the body with an assertion that actually pins failure-path diagnostic text. Concrete replacement: feed a digit-prefixed but structurally-invalid input, `require.Error`, and `assert.NotContains(err.Error(), "expected JSON array or object")`. That would make the "pre-fix text never appears" invariant load-bearing against a dispatch regression.

```
func TestMitmproxyImporter_Native_RegressionLAB2309_ErrorMessagePin(t *testing.T) {
	require.NoError(t, err, "native flow must not produce the pre-fix JSON error")
	// err is nil, so there is no message to inspect.
```

---

## Ticket alignment

_All acceptance criteria for LAB-2309 are met._

- **AC1 — native `.mitm` import succeeds:** MET. `importNative` at `mitmproxy.go:195` is dispatched via first-byte check in `Import`, `real-mitmproxy.mitm` + `sample-mitmproxy.mitm` fixtures are wired into the live-test harness, and `TestMitmproxyImporter_Native_RegressionLAB2309` at `mitmproxy_test.go:817` pins the digit-prefixed dispatch with full field propagation.
- **AC2 — JSON export path preserved:** MET. The `importJSON` path is unchanged; first-byte value simply threads through as an extra parameter. No `ObservedRequest` schema changes.
- **AC3 — bounded against malicious/malformed tnetstring input:** MET. Element cap 64 MB, depth cap 64, list/dict cardinality cap 1 M (counts pairs, not unique keys), flow count cap 500 K, stream initial cap 1 MB, preview cap 64 bytes `%q`-quoted via `previewBytes`. `allowedSchemes` is a strict `http|https` map; `validateHost` rejects userinfo, `:`, `/?#\[]`, ASCII control/whitespace; `requirePort` + `validatePortRange` bound to 0..65535; `validHTTPMethods` strict map.

---

## Tests & documentation

### TEST-001 — low — TEST

**File:** `pkg/importer/mitmproxy_test.go:1547-1561`

See the Non-blocking section above — same finding.

---

This PR has absorbed 16 prior review rounds before this one. The commit log (`cc30acc` … `5fc1e4b`) shows dedicated fix-round commits against rounds 10–16 plus the round-7 CodeRabbit pass. Specifically, the following classes of prior finding have been verified fixed at HEAD and are not re-reported:

- **Attacker bytes rendered with `%s` in error messages** (prior `SEC-BE-nit` from round 10-era review): fixed by routing both `previewString` and `payloadPreview` through `previewBytes` in `helpers.go`, which always uses `%q` quoting so ANSI escapes and control bytes cannot reach the operator's terminal verbatim.
- **URL double-encoding in `constructURL`** (prior `QUAL-medium` from round 10): fixed — `constructURL` now assigns `u.RawPath = pathPart` and `u.Path = decoded` when `url.PathUnescape` succeeds, so pre-encoded mitmproxy paths round-trip as single-encoded.
- **Port smuggling via `:` in host** (prior round 13): fixed — `validateHost` explicitly rejects `:` with documented IPv6 follow-up conditions.
- **Dict cardinality bypass via repeated keys** (CodeRabbit round 7): fixed — `parseTnetDict` counts `pairsParsed`, not `len(result)`.
- **Scheme allowlist, host length cap, empty-host rejection, method allowlist, port 0..65535 range** (round 10-era SEC-BE set): all fixed and now referenced by tests.
- **CLAUDE.md importer-cap staleness** (round 10 DOC): fixed — `CLAUDE.md` is in the canonical file list for this review.
- **Boundary pair tests (at-cap / at-cap+1)** for element size, depth, element count, digit cap, flow cap, preview length (round 10–16 TEST): fixed — corresponding `_AtCap` / `_AtCapPlusOne` tests exist in `tnetstring_test.go` and `mitmproxy_test.go`.
- **Bounded error previews for int/float/bool/dict-key** (CodeRabbit major round 7): fixed — all four helpers route through `previewBytes`.
- **Vendored fixture provenance + MIT license notice** (CodeRabbit round 7): fixed — `test/fixtures/LICENSE.mitmproxy` present and `test/fixtures/README.md` documents the commit-pinned source.

Per-file-line CodeRabbit advisories from the round-16 `4172069851` pass (preallocate slices in `testhelpers_test.go::tnetList/DictElement`; optional `streamInitialCap` → `var` consistency; optional error-message pin on `TestTnetstring_NonEmptyNull`) are either already addressed in `5fc1e4b` or explicit nitpicks the reviewer labelled optional. No re-report needed.

---

PR #84, opened 2026-04-20
fix(test): pass --dangerous-allow-private for localhost crawls (LAB-2308)

santiago-praetorian:
## Summary

- Adds `--dangerous-allow-private` to every `vespasian crawl` invocation in `test/run-live-tests.sh` so the SSRF guard (added in #66) lets localhost seeds through. Fixes the `rest-api`, `soap-service`, `edge-cases`, `crawl-depth`, and `crawl-unreachable` targets.
- Introduces a `TEST_HOST` override (default `localhost`) so the harness can run inside a devcontainer while target services run on the Docker host, e.g. `TEST_HOST=host.docker.internal ./test/run-live-tests.sh`.
- Documents the override and adds a troubleshooting note to `test/README.md`.
- No production crawl behavior changes — `generate`/`scan` sites all use `--probe=false` (or already have the flag), and the SSRF default stays safe.

Closes [LAB-2308](https://linear.app/praetorianlabs/issue/LAB-2308/test-live-rest-api-test-fails-due-to-missing-dangerous-allow-private).

## Test plan

- [x] `./test/setup-live-targets.sh --targets rest-api && ./test/run-live-tests.sh --targets rest-api` — PASS (8 endpoints, 8 expected, 8s) ✅
- [ ] `./test/run-live-tests.sh --targets soap-service,graphql-server` — requires SOAP + GraphQL targets; recommend CI or a follow-up local verify.
- [ ] `./test/run-live-tests.sh --targets edge-cases,crawl-depth,crawl-unreachable` — recommend local verify once CI picks up the branch.
- [ ] Devcontainer path: `TEST_HOST=host.docker.internal ./test/run-live-tests.sh --targets rest-api` — needs a Docker Desktop reviewer to confirm.

PR #92, opened 2026-04-23
fix(crawl): error loudly when seed URL is rejected by frontier (LAB-2438)

santiago-praetorian:
## Summary

Fixes [LAB-2438](https://linear.app/praetorianlabs/issue/LAB-2438/headless-crawl-silently-drops-private-host-seed-rest-apisoap-service).

Running `vespasian crawl http://localhost:<port>` without `--dangerous-allow-private` previously **silently** captured zero requests and exited 0 — the operator got `captured 0 requests` with no hint the seed was never visited. Root cause: `scopeChecker.ssrfCheck` rejects private hosts, `frontier.Push` applies the predicate to the seed, seed is dropped, all workers see an empty frontier in `Pop`, `engine.Crawl` returns `nil`.

This PR surfaces that condition as a loud, actionable error. `--dangerous-allow-private` remains required when the seed is a private host — the flag's existing contract is preserved.

## Scope decision

An earlier draft of this PR also exempted the seed host from the SSRF gate so localhost crawls worked without the flag. That was reverted after review: the flag's name already says "I am crawling a private host, let me through." An invisible seed exemption would make the flag harder to reason about, and future maintainers would need to remember a special case that isn't surfaced in the CLI. See the Linear ticket's Scope decision section for the full rationale.

The alternative semantic change (seed exemption, or renaming the flag) is tracked as possible future work.

## Changes

| File | Change |
|------|--------|
| `pkg/crawl/engine.go` | `Crawl` returns a descriptive error when `frontier.Push(seed)` adds zero entries. The message names the seed and points at `--dangerous-allow-private`. |
| `pkg/crawl/engine_test.go` (new) | Default-suite test that asserts the new error, with a `nil` browser in the engine struct so a future refactor that advances past the guard panics on nil-deref instead of silently passing. |
| `cmd/vespasian/main.go` | `CrawlCmd` and `ScanCmd` `--dangerous-allow-private` help text now states the flag is required when the seed URL is a private host. `GenerateCmd` help text clarifies the flag applies to the probe path only. |
| `README.md` | SSRF Protection feature row and `crawl` / `scan` / `generate` command help blocks updated to match. |
| `test/README.md` | Explains why the harness passes `--dangerous-allow-private` on every live crawl; adds a troubleshooting section for the new error message. |

## Explicitly not changed

- `pkg/crawl/scope.go` — unchanged. `ssrfCheck` still rejects any private host regardless of whether it matches the seed.
- `test/run-live-tests.sh` — unchanged. The harness continues to pass `--dangerous-allow-private` on all seven live-crawl invocations (added by #84 / LAB-2308). Consistent with the flag's contract; all live targets are localhost.
- Probe-side `--dangerous-allow-private` (used on `vespasian generate graphql` in the harness) — different code path in `pkg/probe`, out of scope.

## Test plan

- [x] `go test -race ./...` — green (default suite, includes new `TestRodEngine_Crawl_SeedRejectedByFrontierReturnsError`).
- [x] `gofmt -l .` — clean. `go vet ./...` — clean.
- [x] End-to-end reproducer (Linux + chromium):

  ```bash
  ./test/rest-api/rest-api &  # port 9300

  # WITHOUT flag — previously: captured 0 requests, exit 0; now:
  ./bin/vespasian crawl http://localhost:9300 -o /tmp/err.json --depth 2 --max-pages 50
  # → "vespasian: error: crawl failed: seed URL rejected by frontier
  #    (scope, SSRF, or parse): http://localhost:9300; if crawling a
  #    private host (localhost, 127.0.0.1, RFC1918, link-local),
  #    pass --dangerous-allow-private"
  # → exit 1

  # WITH flag — happy path unchanged:
  ./bin/vespasian crawl http://localhost:9300 --dangerous-allow-private \
      -o /tmp/ok.json --depth 2 --max-pages 50
  jq length /tmp/ok.json
  # → 18
  ```
- [ ] Reviewer to run `./test/setup-live-targets.sh && ./test/run-live-tests.sh` on macOS and confirm the harness still passes (no harness-side changes in this PR so expected to be green).

## Regression lineage

- #66 (`f62fbe0 feat: replace Katana headless crawl with concurrent go-rod engine`) introduced the silent-drop behavior by running the seed through the same scope predicate used for discovered links.
- #84 (LAB-2308) worked around the symptom in the harness by passing `--dangerous-allow-private` on every crawl. That remains the right posture for the harness; this PR keeps the harness as-is and fixes the underlying UX bug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Vespasian takes a different approach: it observes actual network traffic at the 
 | **Traffic Import** | Import existing captures from Burp Suite XML, HAR 1.2 files, and mitmproxy dumps |
 | **Active Probing** | OPTIONS discovery, JSON schema inference, WSDL document fetching, and GraphQL introspection |
 | **Path Normalization** | `/users/42` and `/users/87` become `/users/{id}` with known literal preservation (`/me`, `/self`) |
-| **SSRF Protection** | Blocks probing of private and loopback addresses by default. Use `--dangerous-allow-private` to test internal targets. |
+| **SSRF Protection** | Blocks crawling and probing of private and loopback addresses by default. Pass `--dangerous-allow-private` to test internal targets (localhost, 127.0.0.1, RFC1918, link-local); the flag is required when the seed URL is itself a private host. |
 | **Proxy Support** | Route headless browser traffic through Burp Suite or other intercepting proxies |
 | **Two-Stage Pipeline** | Capture once, generate many: separate capture and generation steps for maximum flexibility |
 
@@ -222,7 +222,9 @@ vespasian scan <url> [flags]
   --confidence       Min classification confidence (default: 0.5)
   --probe            Enable active probing (default: true)
   --deduplicate      Deduplicate endpoints before probing (default: true)
-  --dangerous-allow-private  Disable SSRF protection for private targets
+  --dangerous-allow-private  Disable SSRF protection for private/localhost targets.
+                     Required when the seed URL is a private host (localhost,
+                     127.0.0.1, RFC1918, link-local).
   --no-request-id    Disable auto X-Vespasian-Request-Id header
   -v, --verbose      Show requests in real-time
 ```
@@ -241,6 +243,9 @@ vespasian crawl <url> [flags]
   --scope            same-origin or same-domain (default: same-origin)
   --headless         Browser mode (default: true)
   --proxy            Proxy URL for headless browser (e.g., http://127.0.0.1:8080)
+  --dangerous-allow-private  Disable SSRF protection for private/localhost targets.
+                     Required when the seed URL is a private host (localhost,
+                     127.0.0.1, RFC1918, link-local).
   --no-request-id    Disable auto X-Vespasian-Request-Id header
   -v, --verbose      Show requests in real-time
 ```
@@ -267,7 +272,9 @@ vespasian generate <api-type> <capture-file> [flags]
   --confidence       Min classification confidence (default: 0.5)
   --probe            Enable active probing (default: true)
   --deduplicate      Deduplicate endpoints before probing (default: true)
-  --dangerous-allow-private  Disable SSRF protection for private targets
+  --dangerous-allow-private  Disable SSRF protection on the probe path
+                     (OPTIONS/schema/WSDL-fetch/GraphQL-introspection requests
+                     to private hosts).
   -v, --verbose      Show discovered endpoints
 ```
 
--- a/README.md
+++ b/README.md
@@ -223,9 +223,11 @@ vespasian scan <url> [flags]
   --probe            Enable active probing (default: true)
   --deduplicate      Deduplicate endpoints before probing (default: true)
   --dangerous-allow-private  Disable SSRF protection for crawling and probes,
-                     allowing private/localhost targets. Required when the seed
-                     URL is a private host (localhost, 127.0.0.1, RFC1918,
-                     link-local).
+                     allowing private/localhost targets (localhost, 127.0.0.1,
+                     RFC1918, link-local). Required when the seed URL is a
+                     private host, otherwise the crawl exits with an error and
+                     captures nothing. WARNING: Do not use on production
+                     systems.
   --no-request-id    Disable auto X-Vespasian-Request-Id header
   -v, --verbose      Show requests in real-time
 ```
@@ -244,9 +246,12 @@ vespasian crawl <url> [flags]
   --scope            same-origin or same-domain (default: same-origin)
   --headless         Browser mode (default: true)
   --proxy            Proxy URL for headless browser (e.g., http://127.0.0.1:8080)
-  --dangerous-allow-private  Disable SSRF protection for private/localhost targets.
-                     Required when the seed URL is a private host (localhost,
-                     127.0.0.1, RFC1918, link-local).
+  --dangerous-allow-private  Disable SSRF protection for crawling, allowing
+                     private/localhost targets (localhost, 127.0.0.1, RFC1918,
+                     link-local). Required when the seed URL is a private
+                     host, otherwise the crawl exits with an error and
+                     captures nothing. WARNING: Do not use on production
+                     systems.
   --no-request-id    Disable auto X-Vespasian-Request-Id header
   -v, --verbose      Show requests in real-time
 ```
@@ -274,8 +279,9 @@ vespasian generate <api-type> <capture-file> [flags]
   --probe            Enable active probing (default: true)
   --deduplicate      Deduplicate endpoints before probing (default: true)
   --dangerous-allow-private  Disable SSRF protection on the probe path
-                     (OPTIONS/schema/WSDL-fetch/GraphQL introspection requests
-                     to private hosts).
+                     (OPTIONS/schema/WSDL-fetch/GraphQL introspection) for
+                     private/localhost targets. WARNING: Do not use on
+                     production systems.
   -v, --verbose      Show discovered endpoints
 ```
 
--- a/README.md
+++ b/README.md
@@ -222,9 +222,10 @@ vespasian scan <url> [flags]
   --confidence       Min classification confidence (default: 0.5)
   --probe            Enable active probing (default: true)
   --deduplicate      Deduplicate endpoints before probing (default: true)
-  --dangerous-allow-private  Disable SSRF protection for private/localhost targets.
-                     Required when the seed URL is a private host (localhost,
-                     127.0.0.1, RFC1918, link-local).
+  --dangerous-allow-private  Disable SSRF protection for crawling and probes,
+                     allowing private/localhost targets. Required when the seed
+                     URL is a private host (localhost, 127.0.0.1, RFC1918,
+                     link-local).
   --no-request-id    Disable auto X-Vespasian-Request-Id header
   -v, --verbose      Show requests in real-time
 ```
@@ -273,7 +274,7 @@ vespasian generate <api-type> <capture-file> [flags]
   --probe            Enable active probing (default: true)
   --deduplicate      Deduplicate endpoints before probing (default: true)
   --dangerous-allow-private  Disable SSRF protection on the probe path
-                     (OPTIONS/schema/WSDL-fetch/GraphQL-introspection requests
+                     (OPTIONS/schema/WSDL-fetch/GraphQL introspection requests
                      to private hosts).
   -v, --verbose      Show discovered endpoints
 ```
--- a/cmd/vespasian/main.go
+++ b/cmd/vespasian/main.go
@@ -480,7 +480,7 @@ type GenerateCmd struct {
 	Confidence            float64 `default:"0.5" help:"Minimum confidence threshold"`
 	Probe                 bool    `default:"true" help:"Enable endpoint probing"`
 	Deduplicate           bool    `default:"true" help:"Deduplicate classified endpoints before probing"`
-	DangerousAllowPrivate bool    `help:"Disable SSRF protection for probes, allowing private/localhost targets. WARNING: Do not use on production systems." name:"dangerous-allow-private"`
+	DangerousAllowPrivate bool    `help:"Disable SSRF protection on the probe path (OPTIONS/schema/WSDL-fetch/GraphQL introspection) for private/localhost targets. WARNING: Do not use on production systems." name:"dangerous-allow-private"`
 	Verbose               bool    `short:"v" help:"Enable verbose logging"`
 }
 
--- a/cmd/vespasian/main.go
+++ b/cmd/vespasian/main.go
@@ -386,7 +386,7 @@ func setupBrowserAndSignals(rawHeaders []string, crawlOpts CrawlOptions, extraOp
 // CrawlCmd crawls a web application to capture HTTP traffic.
 type CrawlCmd struct {
 	URL                   string `arg:"" help:"Target URL to crawl"`
-	DangerousAllowPrivate bool   `help:"Disable SSRF protection for crawling, allowing private/localhost targets. WARNING: Do not use on production systems." name:"dangerous-allow-private"`
+	DangerousAllowPrivate bool   `help:"Disable SSRF protection for crawling, allowing private/localhost targets (localhost, 127.0.0.1, RFC1918, link-local). Required when the seed URL is a private host, otherwise the crawl exits with an error and captures nothing. WARNING: Do not use on production systems." name:"dangerous-allow-private"`
 	CrawlOptions
 }
 
@@ -553,7 +553,7 @@ type ScanCmd struct {
 	Confidence            float64 `default:"0.5" help:"Minimum confidence threshold"`
 	Probe                 bool    `default:"true" help:"Enable endpoint probing"`
 	Deduplicate           bool    `default:"true" help:"Deduplicate classified endpoints before probing"`
-	DangerousAllowPrivate bool    `help:"Disable SSRF protection for probes, allowing private/localhost targets. WARNING: Do not use on production systems." name:"dangerous-allow-private"`
+	DangerousAllowPrivate bool    `help:"Disable SSRF protection for crawling and probes, allowing private/localhost targets (localhost, 127.0.0.1, RFC1918, link-local). Required when the seed URL is a private host, otherwise the crawl exits with an error and captures nothing. WARNING: Do not use on production systems." name:"dangerous-allow-private"`
 
 	CrawlOptions
 }
--- a/pkg/crawl/browser.go
+++ b/pkg/crawl/browser.go
@@ -123,6 +123,7 @@ func (b *BrowserManager) cleanup() {
 // for use with defer in the normal (non-signal) path.
 func (b *BrowserManager) Close() {
 	if b.browser != nil {
+		// #nosec G104 -- best-effort close on a process that may already be dead; any error is unactionable and the subsequent Kill() + cleanup() still execute.
 		b.browser.Close() //nolint:errcheck,gosec // best-effort; process may already be dead
 	}
 	b.Kill()
--- a/pkg/crawl/browser.go
+++ b/pkg/crawl/browser.go
@@ -123,7 +123,7 @@ func (b *BrowserManager) cleanup() {
 // for use with defer in the normal (non-signal) path.
 func (b *BrowserManager) Close() {
 	if b.browser != nil {
-		_ = b.browser.Close() // best-effort; process may already be dead
+		b.browser.Close() //nolint:errcheck,gosec // best-effort; process may already be dead
 	}
 	b.Kill()
 	b.cleanup()
--- a/pkg/crawl/engine.go
+++ b/pkg/crawl/engine.go
@@ -47,19 +47,27 @@ const DefaultStableWait = 3 * time.Second
 // in cmd/vespasian/main.go.
 const flagDangerousAllowPrivate = "--dangerous-allow-private"
 
-// unparseableURLPlaceholder is substituted for a URL that failed url.Parse AND
-// contains an "@" (meaning we cannot safely strip userinfo without parsing).
-// Emitting the raw string in that case would leak credentials the whole point
-// of redactSeedURL is to hide.
+// unparseableURLPlaceholder is substituted for a URL that cannot be safely
+// stripped of userinfo (either url.Parse failed and "@" is present, or
+// url.Parse succeeded into an opaque form where u.User is not populated).
+// Emitting the raw string in either case would leak credentials — the whole
+// point of redactSeedURL is to hide them.
 const unparseableURLPlaceholder = "<unparseable URL with userinfo redacted>"
 
 // redactSeedURL returns raw with any userinfo (user[:password]) removed so the
 // URL can be echoed to stderr / logs without leaking credentials. Behavior:
-//   - url.Parse succeeds: userinfo is stripped and the URL is re-serialized.
-//   - url.Parse fails AND raw contains "@": fail closed — emit a generic
-//     placeholder instead of the raw string, since a parse failure on a URL
-//     with "@" (e.g. "http://admin:se%zz@host/path" — invalid percent escape
-//     in userinfo) would otherwise echo the credentials verbatim.
+//   - url.Parse succeeds AND the re-serialized URL has no "@": userinfo is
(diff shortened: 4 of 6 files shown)


eli-wald (changes requested):

**PR:** https://github.com/praetorian-inc/vespasian/pull/92
**Linear:** https://linear.app/praetorianlabs/issue/LAB-2438/headless-crawl-silently-drops-private-host-seed-rest-apisoap-service
**Head SHA:** `7e5bcb9b54ec97242351a523bb8f485883749497`
**Reviewed at:** 2026-04-23
**Worktree:** `/Users/eliwald/tools/caesars/vespasian/.worktrees/capability-pr-review/pr-92`

| Totals   | critical | high | medium | low | nit |
|----------|----------|------|--------|-----|-----|
| Count    | 0        | 0    | 3      | 2   | 2   |
| Blockers | 0        | 0    | 3      | 1   | 0   |

## Live-test results (`./test/run-live-tests.sh`)

After building this branch (`v1.0.0-51-g7e5bcb9`) and running the full live-test harness on macOS:

```
make build                     → OK (40 MB binary at bin/vespasian)
./test/setup-live-targets.sh   → rest-api:8990, soap-service:8991, graphql-server:8992 all listening
./test/run-live-tests.sh       → exit 0
```

**All 20 targets PASS, 0 fail, 0 skipped:**

| classifier-edge | crawl-depth | crawl-unreachable | edge-cases | generate-graphql | generate-graphql-imports |
|---|---|---|---|---|---|
| generate-rest | generate-wsdl | graphql-server | import-base64 | import-burp | import-duplicates |
| import-empty | import-har | import-malformed | import-mitmproxy | import-unicode | rest-api |
| soap-service | spec-edge | | | | |

This confirms LAB-2438 AC #4 ("the harness continues to pass `--dangerous-allow-private` on all seven crawl invocations"). rest-api completed in 12 s, soap-service in 8 s, crawl-depth in 30 s; every expected endpoint count matched exactly. No harness changes were made in this PR, and none were needed — the fix is purely on the error-path UX side. Results artifacts saved to `worktree/test/.results/`.

### Ticket alignment (LAB-2438)

| AC | Status |
|---|---|
| AC1: error exits non-zero, names seed URL, points at `--dangerous-allow-private` | **Met** — verified by code read of `engine.go:Crawl` and by `TestRodEngine_Crawl_SeedRejectedByFrontierReturnsError` |
| AC2: with the flag, traffic is still captured normally | **Met by harness (live tests green), but NOT pinned in default suite — see TEST-001** |
| AC3: unit test uses nil browser so future refactor panics on nil-deref | **Met** — rodEngine built with explicit `browser: nil`; Push-rejection returns before any worker goroutine launches |
| AC4: live-test harness still green | **Met — 20/20 PASS locally** |
| AC5: docs updated (main.go help, test/README.md, README.md) | **Partial — see QUAL-001; README CLI reference omits two key clauses** |
| AC6: `--dangerous-allow-private` semantics unchanged; `scope.go` untouched | **Met** — verified by `git diff main...HEAD -- pkg/crawl/scope.go` (empty) |

---

## Blockers

### QUAL-001 — medium — QUAL

**File:** `README.md:247-251` (and `:225-229` for scan)

**Issue:** The kong help text in `main.go` reads *"Required when the seed URL is a private host, otherwise the crawl exits with an error and captures nothing. WARNING: Do not use on production systems."* The README CLI reference sections for `vespasian crawl` (lines 247-251) and `vespasian scan` (lines 225-229) are abbreviated and **omit both clauses**. An operator who reads the README but not `--help` will not know the flag is *required* (not optional) for private seeds, nor see the production warning — undoing the operator-facing clarity this PR was built to deliver.

**Recommendation:** Update both README flag descriptions to mirror the kong help text verbatim. Apply the same update to the `generate` section at lower priority.

---

**File:** `pkg/crawl/engine.go:43-50`
**CVSS:** `CVSS:3.1/AV:L/AC:H/PR:L/UI:R/S:U/C:L/I:N/A:N` (base 2.2)

**Issue:** `redactSeedURL` returns `raw` unchanged whenever `url.Parse` errors. `url.Parse` is very permissive, so the attack surface is narrow — but malformed escape sequences in userinfo (e.g. `http://admin:se%zz@host/path`) produce `net/url: invalid URL escape "%zz"`, Parse fails, and the credential-bearing seed is echoed verbatim to stderr. The whole point of `redactSeedURL` is to redact credentials before logging; the fail-open path defeats that in precisely the case where the operator has an unusual (typoed) credentialed URL.

**Recommendation:** On parse failure, fail **closed**: if `raw` contains `@` before the first `/`, emit a generic placeholder like `<unparseable URL with userinfo>` instead of the raw string. Well-formed URLs without userinfo remain unchanged.

```go
func redactSeedURL(raw string) string {
    u, err := url.Parse(raw)
    if err != nil {
        return raw
    }
    u.User = nil
    return u.String()
}
```

---

### TEST-001 — medium — TEST *(missing tests)*

**File:** `pkg/crawl/engine_test.go`

**Issue:** LAB-2438 AC #2 says "with `--dangerous-allow-private` the crawl still captures traffic normally (no happy-path regression)". The PR has the negative test (reject-all predicate; seed rejected; error returned) but no symmetric **unit-level** positive test that wires `Crawl` with an accept-all predicate and asserts the rejection error is NOT emitted. Today AC #2 is pinned only by the live-test harness (passing locally as noted above) and the integration-tagged tests. If the default suite is the tripwire for regressions, AC #2 currently has no tripwire.

**Recommendation:** Add a unit test that constructs `rodEngine` with an accept-all predicate, calls `Push` with a seed (`Push` returns >0), and asserts the function does NOT return the `seed URL rejected by frontier` error. Use a fake frontier or return before any worker touches the `nil` browser.

---

### TEST-003 — low — TEST *(missing tests)*

**File:** `pkg/crawl/engine_test.go:115-116`

**Issue:** `TestRedactSeedURL` covers no-userinfo, user+password, user-only, empty password, and malformed. It does **not** cover IPv6 literal with userinfo (e.g. `http://u:p@[::1]:8080/x`) — bracket notation interacts with `url.Parse` differently than hostname/IP. Empty string is also missing. These are common edge cases where `url.Parse` has subtle behavior and `redactSeedURL`'s output could drift without detection.

**Recommendation:** Add two table cases:

```go
{"ipv6 with userinfo", "http://u:p@[::1]:8080/x", "http://[::1]:8080/x"},
{"empty string", "", ""},
```

If IPv6 round-trips through `url.Parse` with altered bracketing, document the normalization in the expected column rather than papering over the discrepancy.

---

## Non-blocking

### QUAL-002 — nit — QUAL

**File:** `pkg/crawl/engine.go:30 + :55`

**Issue:** `DefaultConcurrency` (line 30) and `MaxConcurrency` (line 55) are semantically paired and validated together in `newRodEngine`, but they are separated by `flagDangerousAllowPrivate` (line 37) and `redactSeedURL` (lines 43-50), breaking the Go idiom of grouping related consts.

**Recommendation:** Group concurrency-related consts in a single block before `engineOptions`. Move `flagDangerousAllowPrivate` and `redactSeedURL` to a dedicated "operator-message helpers" section.

---

### QUAL-003 — nit — QUAL

**File:** `pkg/crawl/engine_test.go:42`

**Issue:** The nil-browser comment explains the panic-on-refactor safety net correctly, but the actual nil-deref site is `visitPage`'s `e.browser.Page()` at `engine.go:241`, not the worker loop. Minor readability gap — the design itself is correct (confirmed: `rodEngine` is built with explicit `browser: nil`; `Push`-rejection returns before any worker goroutine launches, satisfying AC #3 as specified).

**Recommendation:** Extend the comment: *"...nil-browser deref (at `e.browser.Page()` in `visitPage`) rather than silently passing."* Optionally add an inline `// AC #3: nil browser is deliberate — see LAB-2438` so the intent survives code archaeology.

---

### TEST-002 — low — TEST

**File:** `pkg/crawl/engine_test.go:62`

**Issue:** The existing test asserts the error contains "rejected", the seed URL, and `flagDangerousAllowPrivate` — good. But the richer operator-facing message is *"seed URL rejected by frontier (scope, SSRF, or parse): <seed>; if crawling a private host (localhost, 127.0.0.1, RFC1918, link-local), pass --dangerous-allow-private"*. Two load-bearing clauses are not pinned: the `(scope, SSRF, or parse)` diagnostic breakdown and the private-host enumeration. A refactor dropping either would pass every existing assertion.

**Recommendation:** Add `strings.Contains(err.Error(), "scope, SSRF, or parse")` and `strings.Contains(err.Error(), "private host")` to lock in the operator-facing diagnostic.

---

## Ticket alignment

All six LAB-2438 ACs addressed as above. AC #5 (docs) is partially met — see QUAL-001.

## Tests & documentation

Blocker entries TEST-001 and TEST-003 and DOC-adjacent QUAL-001 are enumerated above.

**1 prior finding confirmed resolved:**
- CodeRabbit nit on `README.md` (inconsistent "GraphQL-introspection" vs "GraphQL introspection") — **FIXED at HEAD**; README now uses "GraphQL introspection" consistently across all 5 occurrences (lines 47, 277, 291, 301, 332).

---

## Follow-up notes (pre-existing, out of the scope of this PR)

The `redactSeedURL` helper closes the credential-leak path at `engine.go:Crawl`'s rejection error — but several **other** operator-facing paths still log the raw seed URL (including any userinfo). These pre-existed on `main` and the PR body explicitly scopes them out. I'm noting them here because the engineer introduced the redaction pattern in this PR and **plumbing it through a few more call sites is a small, high-value follow-up** — not a blocker on this PR:

| Location | Leak | Recommendation |
|---|---|---|
| `cmd/vespasian/main.go:415` | `fmt.Fprintf(os.Stderr, "crawling %s ...", c.URL, ...)` runs on every verbose crawl, not just the error path | Export `crawl.RedactSeedURL` and call it here |
| `cmd/vespasian/main.go:612-625` | `probeWSDLDocument(c.URL, ...)` logs `c.URL` in verbose mode AND the WSDL URL derived from it is injected into `ObservedRequest.URL`, which is persisted to `capture.json` | Strip `u.User` inside `probeWSDLDocument` before computing `wsdlURL`; redact userinfo from verbose logs |
| `pkg/crawl/engine.go:207` | `fmt.Fprintf(e.opts.Stderr, "worker %d: error visiting %s: %v", ..., entry.URL, ...)` | Normalize URLs at `urlFrontier.Push` time: strip `u.User` before storing, which protects every downstream log/error |
| `cmd/vespasian/main.go:908` | `validateURL` wraps `url.Parse` failures with `%q` on the raw input | Textual redaction before the error message since `url.Parse` has failed |

A single-line change in `urlFrontier.Push` to strip `u.User` on insertion would close three of these four at once, since everything downstream reads URLs from the frontier.

---

## Summary for the engineer

- **Live tests 20/20 green** (`./test/run-live-tests.sh` exit 0). LAB-2438 AC #4 holds locally.
- The core fix is correct and well-scoped. The `redactSeedURL` helper, const anchoring, and nil-browser test design are all thoughtful.
- 4 blockers are all **small** to address: README text update (QUAL-001), fail-closed redaction edge case (SEC-BE-001), happy-path unit test (TEST-001), and an IPv6 table entry (TEST-003). Collectively maybe 40 lines of code.
- The CodeRabbit nit from the earlier round is confirmed fixed.
- The pre-existing credential-leak paths are the most interesting leftover work, but they predate this PR and the author explicitly scoped them out — worth a follow-up ticket rather than expanding this PR's blast radius.

---

santiago-praetorian:
## Review iteration complete — ready for re-review

All findings from @eli-wald's review ([4164461391](https://github.com/praetorian-inc/vespasian/pull/92#pullrequestreview-4164461391)) are addressed across the 3 new commits on this branch (`ee8445d`, `8569e95`, `6e1a737`). I then ran three rounds of `/praetorian-capability-pr-review` against the local branch tip, each time addressing the findings it surfaced. The fourth run (against current head `6e1a737`) returned **zero findings across all three agents**.

### Review convergence

| Head       | Cycle                           | Findings | Blockers | Recommendation      |
|------------|---------------------------------|---------:|---------:|---------------------|
| `7e5bcb9`  | eli-wald 4164461391             | 7        | 4        | REQUEST_CHANGES     |
| `ee8445d`  | `/praetorian-capability-pr-review` #1 | 3  | 0        | COMMENT             |
| `8569e95`  | `/praetorian-capability-pr-review` #2 | 2  | 0        | COMMENT             |
| `6e1a737`  | `/praetorian-capability-pr-review` #3 | **0** | **0** | **APPROVE**     |

### Blockers from the original review — all resolved

- **QUAL-001** — README `--dangerous-allow-private` help text for `crawl`/`scan`/`generate` now mirrors the kong help text verbatim, including the required/captures-nothing clause and the production-systems warning.
- **SEC-BE-001** — `redactSeedURL` now fails closed in two additional cases:
  - `url.Parse` fails on a string containing `@` (e.g. invalid percent-escape in userinfo).
  - `url.Parse` succeeds on an opaque-form URL (e.g. `http:user:pass@host/path` — credentials in `u.Opaque`, `u.User=nil`) — a post-`u.String()` residual-`@` check forces the placeholder.
  - New constant `redactedURLPlaceholder = "<URL with userinfo redacted>"` (renamed from `unparseableURLPlaceholder` since it now also fires on parseable-but-opaque URLs).
- **TEST-001** — New `TestRodEngine_Crawl_SeedAcceptedDoesNotReturnRejectionError` pins LAB-2438 AC #2 in the default suite (previously only covered by the live-test harness).
- **TEST-003** — Added `ipv6 with userinfo`, `empty string`, `malformed with userinfo`, and `opaque with userinfo` table cases to `TestRedactSeedURL` (9 cases total).

### Non-blockers from the original review — all resolved

- **QUAL-002** — `DefaultConcurrency` / `MaxConcurrency` / `DefaultStableWait` grouped together; flag and helper moved under an explicit "operator-message helpers" section.
- **QUAL-003** — Nil-browser test comment extended to name the deref site (`e.browser.Page()` in `visitPage`) and anchor LAB-2438 AC #3.
- **TEST-002** — Pinned the load-bearing `"scope, SSRF, or parse"` and `"private host"` clauses in the rejection test.

### End-to-end live-test harness — 20/20 PASS

```
  Test Summary
  --------------------------  --------  ----------  ---------  --------
  crawl-depth                 PASS      -           -          50s
  generate-graphql            PASS      8           8          0s
  generate-rest               PASS      8           8          0s
  generate-wsdl               PASS      3           3          1s
  graphql-server              PASS      8           8          1s
  import-base64               PASS      2           2          1s
  import-burp                 PASS      5           5          0s
  rest-api                    PASS      8           8          17s
  spec-edge                   PASS      -           -          1s

  Total: 20 passed, 0 failed, 0 skipped
```

This confirms LAB-2438 AC #4 (harness continues to pass `--dangerous-allow-private` on all `vespasian crawl` invocations) — no harness-side changes were needed or made.

### Verification on the pushed branch

- `gofmt -s` clean
- `go vet ./...` clean
- `golangci-lint run` → 0 issues
- `go test -race ./...` → all packages PASS
- CLI `--help` output verified for `crawl` / `scan` / `generate` against the kong source
- End-to-end redaction verified against a credentialed private seed

Ready for re-review.

eli-wald (approved):
# Praetorian Capability PR Review — round 2

**PR:** https://github.com/praetorian-inc/vespasian/pull/92
**Linear:** https://linear.app/praetorianlabs/issue/LAB-2438/headless-crawl-silently-drops-private-host-seed-rest-apisoap-service
**Prior review SHA:** `7e5bcb9b54ec97242351a523bb8f485883749497` (round 1: 4 blockers + 3 non-blocking)
**This review SHA:** `81ef83f0bd36b44927589029c3ff47fd2a23b376`
**LAB-2438 delta:** `engine.go +26/-10`, `engine_test.go +70 LOC (new tests + table additions)`, `README.md +22`, `browser.go +3/-1` (plus unrelated main merge bringing in LAB-2222 work)

**Recommendation:** COMMENT (no blockers; 2 nit-level follow-ups)


| Totals   | critical | high | medium | low | nit |
|----------|----------|------|--------|-----|-----|
| Count    | 0        | 0    | 0      | 1   | 1   |
| Blockers | 0        | 0    | 0      | 0   | 0   |

## Round-1 blockers — all confirmed fixed at HEAD

| Round-1 blocker | Status | Fixed by |
|---|---|---|
| **QUAL-001** (medium) README CLI reference omits "captures nothing" + "WARNING" clauses | ✅ FIXED | `ee8445d` — README `crawl` and `scan` sections now carry the full kong help text clauses (README.md:247-255 and :225-231) |
| **SEC-BE-001** (medium) `redactSeedURL` fails-open on malformed credentialed URLs | ✅ FIXED | `ee8445d` + `8569e95` — now fails closed to `redactedURLPlaceholder` across all four branches (success/clean, success-with-residual-@, parse-fail-with-@, parse-fail-without-@); also covers the opaque URL form `http:user:pass@host/path` where `url.Parse` populates `u.Opaque` instead of `u.User` |
| **TEST-001** (medium, missing_tests) No default-suite happy-path test for AllowPrivate / AC #2 | ✅ FIXED | `ee8445d` + `6e1a737` — `TestRodEngine_Crawl_SeedAcceptedDoesNotReturnRejectionError` added; assertion tightened from substring check to `err != nil` fail, with nil-browser panic as a secondary tripwire |
| **TEST-003** (low, missing_tests) `TestRedactSeedURL` missing IPv6 + empty cases | ✅ FIXED | `ee8445d` — table now covers IPv6 with userinfo, empty string, opaque form, `@`-in-path / `@`-in-query (deliberate false positives), and malformed-with-userinfo. Placeholder is asserted directly via `want: redactedURLPlaceholder` |

Also addressed:
- **CodeRabbit nit** (review 4165767099): `redactSeedURL` docstring doesn't document that the residual-`@` check can be a false positive on `@` in path/query. ✅ FIXED by `d538196` — both the function-level comment and the inline comment at the residual-`@` check are now precise about the deliberate conservative behavior.
- **GitHub Advanced Security** alert #31 (gosec G104): unhandled error on `b.browser.Close()` in `BrowserManager.Close`. ✅ FIXED by `81ef83f` — dual annotation (`// #nosec G104` + `//nolint:errcheck,gosec`) covering both scanners, narrow line-scoped, with commit-message rationale. Safe to swallow because `Kill()` on the next line unconditionally terminates Chrome via `killOnce.Do`, so the swallowed error cannot produce a dangling Chrome process.
- **Post-merge errcheck** (`14e15b9`): matching nolint convention on `browser.Close()` after merging main brought in the LAB-2222 cookie work that changed Close's shape.

## Ticket alignment (LAB-2438)

| AC | Status |
|---|---|
| AC1: error exits non-zero, names seed URL, points at `--dangerous-allow-private` | **Met** — error-message shape now pinned in tests (clauses `scope, SSRF, or parse` and `private host`) |
| AC2: with the flag, traffic is still captured normally | **Met AND pinned in default suite** (was integration-only in round 1) |
| AC3: unit test uses nil browser so future refactor panics on nil-deref | **Met** (unchanged) |
| AC4: live-test harness still green | **Verified in round 1** — 20/20 PASS |
| AC5: docs updated | **Met** — README now mirrors kong help text |
| AC6: `--dangerous-allow-private` semantics unchanged; `scope.go` untouched | **Met** (unchanged) |

---

## Non-blocking

### QUAL-001 — low — QUAL

**File:** `pkg/crawl/browser.go:137-141`

**Issue:** `SetCookies` guards with `if b == nil || b.browser == nil { return fmt.Errorf("browser not connected") }`. The `b == nil` check is legal on pointer receivers but is defensive dead-weight for real call sites — `BrowserManager` is always constructed via `NewBrowserManager` which returns a non-nil pointer on success. The `b.browser == nil` check IS load-bearing (partial construction during error unwinding, tests). A reviewer encountering the combined guard might conclude nil `*BrowserManager` is an intended usage pattern.

**Important context:** this code is **unchanged by PR #92** — it came in via the main merge from the already-merged LAB-2222 PR #68. Flagging for awareness, not action on this PR.

**Recommendation:** Follow-up cleanup — either (a) keep the combined guard with a same-line comment noting `b.browser == nil` is the realistic production path, or (b) drop the `b == nil` guard since callers should already be checking `NewBrowserManager`'s error. Open a separate ticket if you want to track.

---

### TEST-001 — nit — TEST

**File:** `pkg/crawl/engine_test.go:175-208`

**Issue:** My round-1 review asked for a whitespace/control-byte userinfo case. Behaviorally equivalent to the already-covered "malformed with userinfo" case (url.Parse rejects the control byte, residual `@` forces the placeholder), so coverage of the fail-closed branch is adequate. But the `redactSeedURL` docstring lists percent-escape parse failures as *one* example of the parse-fail-plus-`@` branch, not the only trigger. A future `url.Parse` relaxation that accepts control bytes in userinfo could silently route the case into the `u.User=nil` round-trip path instead of the placeholder — a control-byte test would catch that drift.

**Recommendation:** Append one more case to the `TestRedactSeedURL` table:

```go
{"control byte in userinfo redacts to placeholder", "http://u\x00:p@host/path", redactedURLPlaceholder}, //nolint:gosec // G101: intentional malformed test credential used to verify fail-closed redaction
```

No production-code change.

---

## Ticket alignment

All six LAB-2438 ACs met (table above).

## Tests & documentation

All round-1 TEST findings addressed; one new nit enumerated above. Docs (`README.md`, `test/README.md`, kong help) now consistent with the kong-help operator-facing strings.

**8 total** — the 4 round-1 blockers (QUAL-001, SEC-BE-001, TEST-001, TEST-003), the 3 round-1 non-blockers (QUAL-002 const ordering, QUAL-003 nil-browser comment, TEST-002 error-string pin), and the CodeRabbit `@`-heuristic docstring nit. Plus the GHAS alert #31 (G104) resolution is confirmed narrow and correctly suppressed.

---

## Summary for the engineer

- **Round-1 blockers all confirmed resolved.** The `redactSeedURL` hardening is excellent — four distinct branches are enumerated and each fails closed correctly. The opaque-URL escape (`http:u:p@h/p` → `u.Opaque` → `u.User = nil` is a no-op) was a subtle attack shape the round-1 fix caught on the second pass via `8569e95`.
- **Docstring + test discipline post-review is high.** The docstring enumerates each branch's behavior with explicit note of the deliberate `@`-in-path/query false positive; the test table covers every enumerated branch with named cases.
- **Zero new blockers, two nits.** The `SetCookies` guard nit is unchanged code from the main merge (not really this PR), and the control-byte test case is drift-protection.
- **No remaining action required to ship this PR** beyond the two nit-level follow-ups (neither is blocking).

---

PR #101, opened 2026-04-25
feat(rest): parse form-encoded and multipart request bodies (LAB-2106)

maneelmodi:
## Summary

Implements form-encoded (`application/x-www-form-urlencoded`) and multipart (`multipart/form-data`) request body parsing in the OpenAPI generator, completing LAB-2106. Form-based endpoints (logins, uploads, legacy APIs) now get complete `requestBody` schemas instead of appearing parameterless in generated specs.

Linear: https://linear.app/praetorianlabs/issue/LAB-2106

## Commit timeline

| Commit | What |
|---|---|
| `fb1a8bf` | Core feature (form parsing) |
| `7445244` | Tier-2 fixes (body-fingerprint dedup + sort paths) |
| `8fc0079` | Inline doc on body-fingerprint design |
| `8608ddd` | Pre-review polish (SOAP body test + benchmarks) |
| `2a3f15c` | 9 reviewer findings + 3 Go 1.24+ idiom modernizations |
| **`b9c0113`** | **Santiago + CodeRabbit review findings (this commit)** |

### Core feature surface
- `pkg/generate/rest/form.go` (new): `ParseURLEncodedForm`, `ParseMultipartForm`, `extractBoundary`, case-insensitive `getHeader`, merge helpers
- `pkg/generate/rest/openapi.go`: content-type-partitioned `buildOperation`, all-media-types `extractComponents` with sorted iteration, sanitized `resourceNameFromPath`
- `pkg/classify/classifier.go`: content-type + body-fingerprint aware `Deduplicate`, multipart boundary normalization
- `test/rest-api/main.go`: `/api/login` (urlencoded) and `/api/upload` (multipart) endpoints

### Hardening (b9c0113)
- **`maxMultipartParts = 1000`** cap in `ParseMultipartForm` to prevent OOM under adversarial input (SEC-BE-002)
- **Scope-separated fingerprint maps** in `extractComponents` (`fingerprintToReqName` / `fingerprintToRespName`) so echo-style endpoints don't tag responses as `…Request` (CodeRabbit cross-scope $ref)
- **Deterministic golden** — `reference-capture.json` + `expected-spec.yaml` extended with urlencoded + multipart entries so `test_generate_rest` pins form-body output (TEST-006)
- **Tightened test assertions** — body-content `ElementsMatch` instead of count-only; multipart partial-content-types subtest; repeated-file-fields E2E test; `toCamelCase` ASCII-only doc

## Testing matrix

| Layer | Result |
|---|---|
| Unit tests | **482 passing**, 89.1% coverage `pkg/classify`, **93.2%** `pkg/generate/rest` |
| `-count=10 -race` heavy stress | No flakiness |
| Full repo `go test -race` | 9 packages clean, 0 regressions |
| `./test/run-live-tests.sh` | **20/20 targets pass** (including the regenerated `generate-rest` golden) |
| `vespasian scan` against real vulnerable apps | **7 apps**: DVWA, Mutillidae, Juice Shop, WebGoat, bWAPP, VAmPI, NodeGoat |
| Real mitmproxy capture → import → spec | Multipart `file: binary` correctly produced |
| `vespasian crawl --proxy + -H Cookie` | Auth + proxy routing works end-to-end |
| Burp XML / HAR / mitmproxy importers | All preserve form bodies correctly |
| Spectral OpenAPI 3.0 lint | 0 errors across all generated specs |
| openapi-generator codegen (Python + TypeScript) | Clients compile with form-aware signatures |
| PUT + PATCH with form bodies | E2E validated; `Update` component prefix correctly selected |
| `Format: json` output | Valid OpenAPI 3.0 JSON |
| `--probe=true` with form endpoints | Probe stage doesn't break form pipeline |
| 10,000-endpoint scale | 5.15s, 577 MB peak RSS, no crashes |
| 50-iteration memory leak test | Heap stable, +86 KB total |
| Adversarial fuzz | 16/16 multipart + 13/13 urlencoded malformed cases handled |
| Idempotency (5 runs of new golden) | Byte-identical output |
| `golangci-lint govet+shadow` | 0 issues |
| `make fmt` | No-op (all code gofmt-compliant) |

## Performance baselines (linux/arm64)

| Benchmark | ns/op |
|---|---|
| `BenchmarkDeduplicate/n=100` | 69,310 |
| `BenchmarkDeduplicate/n=1000` | 698,116 |
| `BenchmarkDeduplicate/n=5000` | 3,304,505 (linear scaling) |
| `BenchmarkParseURLEncodedForm` | 2,503 |
| `BenchmarkParseMultipartForm` | 13,355 |

## Design decisions made deliberately

1. **SHA-256 [:8] for body fingerprint** — stdlib, 64-bit collision probability ~7×10⁻¹³ at realistic scale, documented inline
2. **Allowlist-based extension stripping** — paths can legitimately contain dots; only known web-framework extensions are stripped
3. **CamelCase via rune-walking, ASCII-only** — handles `-`, `_`, `.` uniformly; non-ASCII intentionally falls through to `Resource` fallback
4. **mitmproxy-as-proxy for real multipart capture** — bypasses pre-existing CDP `hasPostData` limitation; matches real security-assessment workflow
5. **Body fingerprint added only for non-empty bodies** — preserves bodyless dedup (GETs etc.); minimizes regression risk
6. **Multipart boundary normalized before hashing** — clients generate random boundaries; without normalization identical logical forms would never dedup
7. **Multipart parts capped at 1000** — bounds memory under adversarial input; realistic forms are well below this
8. **Scope-separated fingerprint maps in extractComponents** — `fingerprintToReqName` and `fingerprintToRespName` are independent so echo-style endpoints don't cross-pollinate names; `nameCounter` stays shared for global collision avoidance
9. **`mergeJSONBodies` delegates to `mergeObjectSchemas`** — unified type-conflict resolution across JSON, urlencoded, and multipart (all promote to string on conflict)
10. **`io.LimitReader(part, 4096)` for multipart text type inference** — type inference only needs a small prefix; avoids materializing huge text fields
11. **`bytes.NewReader(body)` instead of `strings.NewReader(string(body))`** — eliminates a byte→string copy in hot path

## Items deliberately deferred (with rationale)

| Item | Reason | Recommended follow-up |
|---|---|---|
| **Crawler dropping multipart bodies (CDP `hasPostData`)** | Pre-existing in `pkg/crawl/network.go`; out of LAB-2106 scope; workaround validated (mitmproxy as proxy) | Separate ticket for `pkg/crawl` |
| **Super-linear time at 10k+ endpoints** | Performance, not correctness; benchmarks show `Deduplicate` is linear, so bottleneck is in `buildOperation`/`extractComponents`; requires profiling | Separate ticket |
| **`getContentType` cross-package duplication (QUAL-001)** | Identical helper exists in `pkg/generate/graphql/infer.go`; resolving requires a new shared utility package (e.g., `pkg/httputil`); architectural decision out of LAB-2106 scope | Separate refactor ticket |
| **Content-type normalization duplication (QUAL-003)** | Same as QUAL-001 — would need shared package | Separate refactor ticket |
| **TEST-004 (Santiago): "first observation's type wins" assertion for `mergeJSONBodies`** | Conflicts with our intentional unification (commit `2a3f15c`) where `mergeJSONBodies` delegates to `mergeObjectSchemas` (promotes conflicts to `string` matching form merger). `TestMergeJSONBodies_TypeConflictPromotesToString` covers the unified behavior. Asked Santiago to confirm preference (unified vs first-wins) in PR review thread. | Awaiting reviewer decision |
| **Repeated urlencoded keys as arrays** | Schema-design decision (array type? items schema? minItems?) not in AC; documented in test comments | Revisit if user demand surfaces |
| **JSON output format CLI flag** | `OpenAPIGenerator.Format = "json"` exists in code; no CLI flag exposes it. Code path verified directly | Optional CLI ergonomics ticket |

## Test plan

- [x] `make check` passes
- [x] `go test -race ./...` 9 packages clean
- [x] `-count=10` stress no flakiness
- [x] `./test/run-live-tests.sh` 20/20 pass (including regenerated golden)
- [x] All LAB-2106 acceptance criteria satisfied end-to-end
- [x] 7 real vulnerable apps validated via `vespasian scan`
- [x] OpenAPI 3.0 conformance via spectral
- [x] Downstream codegen via openapi-generator (Python + TS)
- [x] PUT + PATCH form bodies E2E
- [x] JSON output format
- [x] `--probe=true` with form endpoints
- [x] Adversarial fuzz testing
- [x] Memory leak test
- [x] Idempotency verified
- [x] Performance benchmarks established
- [x] Multipart DoS cap enforced
- [x] Scope-separated $ref naming for echo-style endpoints

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -69,7 +69,7 @@ The CLI (`cmd/vespasian`) uses Kong for argument parsing. Each command (crawl, i
 - **pkg/classify**: Request classification engine with confidence-based heuristics; classifiers for REST, GraphQL, and WSDL; deduplication
 - **pkg/probe**: Active endpoint probing strategies (OPTIONS discovery, JSON schema inference, WSDL document fetching, GraphQL introspection with 3-tier WAF bypass); SSRF protection with DNS rebinding mitigation
 - **pkg/generate**: Spec generation interface and registry; delegates to sub-packages by API type
-- **pkg/generate/rest**: OpenAPI 3.0 generation, path normalization (UUID detection, context-aware parameter naming), JSON schema inference
+- **pkg/generate/rest**: OpenAPI 3.0 generation, path normalization (UUID detection, context-aware parameter naming), JSON schema inference, form-encoded and multipart request-body inference
 - **pkg/generate/graphql**: GraphQL SDL generation from introspection results or traffic-based inference
 - **pkg/generate/wsdl**: WSDL generation from SOAP traffic, WSDL document parsing, type inference from SOAP envelopes
 - **pkg/importer**: Traffic importers for Burp Suite XML, HAR 1.2, and mitmproxy dumps; format registry with size limits (500MB)
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ Vespasian classifies and generates specifications for three API types:
 
 | API Type | Classification Signals | Output Format | Probing |
 |----------|----------------------|---------------|---------|
-| **REST** | JSON/XML content-type, `/api/` `/v1/` path patterns, HTTP methods | OpenAPI 3.0 (YAML/JSON) | OPTIONS discovery, JSON schema inference |
+| **REST** | JSON/XML content-type, `/api/` `/v1/` path patterns, HTTP methods | OpenAPI 3.0 (YAML/JSON) | OPTIONS discovery, JSON, urlencoded, and multipart request-body inference |
 | **GraphQL** | `/graphql` path, query structure in POST body, `data`/`errors` response keys | GraphQL SDL | Tiered introspection queries (3 tiers for WAF bypass) |
 | **WSDL/SOAP** | SOAPAction header, SOAP envelope in body, `?wsdl` URL parameter | WSDL XML | Active `?wsdl` document fetching |
 
--- a/pkg/classify/classifier.go
+++ b/pkg/classify/classifier.go
@@ -128,6 +128,15 @@ func Deduplicate(classified []ClassifiedRequest) []ClassifiedRequest {
 			if ct := getContentType(req.Headers); ct != "" {
 				key += ":" + baseMediaType(ct)
 			}
+			// Append a short fingerprint of the body so distinct payload shapes
+			// on the same endpoint+method+CT survive deduplication. This is
+			// required for downstream form/JSON merge logic in buildOperation
+			// to see all observations and union their fields. 8 bytes (64 bits)
+			// is a deliberate balance: birthday-collision probability is ~7e-13
+			// at 500 distinct bodies per endpoint, well under realistic crawl
+			// scale (capped at MaxPages, default 100). A collision would
+			// silently merge two distinct bodies into one dedup bucket; this
+			// is no worse than pre-fix behavior and worth the simpler key.
 			h := sha256.Sum256(req.Body)
 			key += ":" + hex.EncodeToString(h[:8])
 		}
--- a/pkg/classify/classifier.go
+++ b/pkg/classify/classifier.go
@@ -114,6 +114,16 @@ func Deduplicate(classified []ClassifiedRequest) []ClassifiedRequest {
 			key += ":" + sa
 		}
 
+		// If this observation has a body, include the base content type in the
+		// key so that distinct request body shapes (JSON, urlencoded, multipart)
+		// survive deduplication on the same path. Empty bodies (e.g. GET) do
+		// not split.
+		if len(req.Body) > 0 {
+			if ct := getContentType(req.Headers); ct != "" {
+				key += ":" + baseMediaType(ct)
+			}
+		}
+
 		existing, found := seen[key]
 		if !found {
 			order = append(order, key)
@@ -155,4 +165,26 @@ func getSoapAction(headers map[string]string) string {
 		}
 	}
 	return ""
+}
+
+// getContentType returns the Content-Type header value, case-insensitively.
+func getContentType(headers map[string]string) string {
+	for k, v := range headers {
+		if strings.EqualFold(k, "content-type") {
+			return v
+		}
+	}
+	return ""
+}
+
+// baseMediaType returns the lowercased media type from a Content-Type value,
+// stripped of any parameters (e.g. "; boundary=..."). Returns "" on empty input.
+func baseMediaType(ct string) string {
+	if ct == "" {
+		return ""
+	}
+	if i := strings.Index(ct, ";"); i >= 0 {
+		ct = ct[:i]
+	}
+	return strings.ToLower(strings.TrimSpace(ct))
 }
--- a/pkg/classify/classifier.go
+++ b/pkg/classify/classifier.go
@@ -15,6 +15,8 @@
 package classify
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"net/url"
 	"strings"
 
@@ -114,14 +116,20 @@ func Deduplicate(classified []ClassifiedRequest) []ClassifiedRequest {
 			key += ":" + sa
 		}
 
-		// If this observation has a body, include the base content type in the
-		// key so that distinct request body shapes (JSON, urlencoded, multipart)
-		// survive deduplication on the same path. Empty bodies (e.g. GET) do
-		// not split.
+		// If this observation has a body, include the base content type and a
+		// short hash of the body bytes in the key so that:
+		//   - Distinct body shapes on the same path+method survive as separate
+		//     entries (required for downstream form/JSON field-merge logic in
+		//     buildOperation in pkg/generate/rest/openapi.go, which unions fields
+		//     across observations).
+		//   - Identical bodies still collapse correctly (true duplicates).
+		//   - Empty-body requests (GET, DELETE, HEAD, OPTIONS) are unaffected.
 		if len(req.Body) > 0 {
 			if ct := getContentType(req.Headers); ct != "" {
 				key += ":" + baseMediaType(ct)
 			}
+			h := sha256.Sum256(req.Body)
+			key += ":" + hex.EncodeToString(h[:8])
 		}
 
 		existing, found := seen[key]
--- a/pkg/classify/classifier.go
+++ b/pkg/classify/classifier.go
@@ -15,8 +15,10 @@
 package classify
 
 import (
+	"bytes"
 	"crypto/sha256"
 	"encoding/hex"
+	"mime"
 	"net/url"
 	"strings"
 
@@ -85,7 +87,7 @@ func RunClassifiers(classifiers []APIClassifier, requests []crawl.ObservedReques
 // In practice this is bounded by the upstream crawl layer's MaxPages setting
 // (default 100). The import path (ReadCapture) does not enforce size limits,
 // so callers importing from untrusted capture files should validate input size.
-func Deduplicate(classified []ClassifiedRequest) []ClassifiedRequest {
+func Deduplicate(classified []ClassifiedRequest) []ClassifiedRequest { //nolint:gocyclo // boundary normalization for multipart adds necessary branches
 	type entry struct {
 		req ClassifiedRequest
 	}
@@ -137,7 +139,18 @@ func Deduplicate(classified []ClassifiedRequest) []ClassifiedRequest {
 			// scale (capped at MaxPages, default 100). A collision would
 			// silently merge two distinct bodies into one dedup bucket; this
 			// is no worse than pre-fix behavior and worth the simpler key.
-			h := sha256.Sum256(req.Body)
+			fingerprintBody := req.Body
+			if ct := getContentType(req.Headers); ct != "" {
+				if mt, params, err := mime.ParseMediaType(ct); err == nil && mt == "multipart/form-data" {
+					if boundary := params["boundary"]; boundary != "" {
+						// Multipart bodies contain a random boundary token (per-request) that
+						// would otherwise make every observation unique. Normalize it to a
+						// sentinel so identical logical forms with different boundaries dedup.
+						fingerprintBody = bytes.ReplaceAll(req.Body, []byte(boundary), []byte("BOUNDARY"))
+					}
+				}
+			}
+			h := sha256.Sum256(fingerprintBody)
 			key += ":" + hex.EncodeToString(h[:8])
 		}
 
--- a/pkg/classify/classifier.go
+++ b/pkg/classify/classifier.go
@@ -134,18 +134,23 @@ func Deduplicate(classified []ClassifiedRequest) []ClassifiedRequest { //nolint:
 			// on the same endpoint+method+CT survive deduplication. This is
 			// required for downstream form/JSON merge logic in buildOperation
 			// to see all observations and union their fields. 8 bytes (64 bits)
-			// is a deliberate balance: birthday-collision probability is ~7e-13
+			// is a deliberate balance: birthday-collision probability is ~1e-14
 			// at 500 distinct bodies per endpoint, well under realistic crawl
-			// scale (capped at MaxPages, default 100). A collision would
-			// silently merge two distinct bodies into one dedup bucket; this
-			// is no worse than pre-fix behavior and worth the simpler key.
+			// scale (capped at MaxPages, default 100). With 1M distinct bodies the
+			// probability rises to ~5e-8, still negligible in practice.
+			// A collision would silently merge two distinct bodies into one dedup
+			// bucket; this is no worse than pre-fix behavior and worth the simpler key.
 			fingerprintBody := req.Body
 			if ct := getContentType(req.Headers); ct != "" {
 				if mt, params, err := mime.ParseMediaType(ct); err == nil && mt == "multipart/form-data" {
-					if boundary := params["boundary"]; boundary != "" {
+					if boundary := params["boundary"]; len(boundary) >= 4 {
 						// Multipart bodies contain a random boundary token (per-request) that
 						// would otherwise make every observation unique. Normalize it to a
 						// sentinel so identical logical forms with different boundaries dedup.
+						// boundary < 4 chars: skip normalization, fall through to raw body hash.
+						// This is defensive — RFC 2046 allows 1-70 char boundaries but real-world
+						// browsers/libraries use 30+ chars. A pathologically short boundary would
+						// corrupt the body more than it'd help dedup.
 						fingerprintBody = bytes.ReplaceAll(req.Body, []byte(boundary), []byte("BOUNDARY"))
 					}
 				}
--- a/pkg/classify/classifier.go
+++ b/pkg/classify/classifier.go
@@ -127,7 +127,8 @@ func Deduplicate(classified []ClassifiedRequest) []ClassifiedRequest { //nolint:
 		//   - Identical bodies still collapse correctly (true duplicates).
 		//   - Empty-body requests (GET, DELETE, HEAD, OPTIONS) are unaffected.
 		if len(req.Body) > 0 {
-			if ct := getContentType(req.Headers); ct != "" {
+			ct := getContentType(req.Headers)
+			if ct != "" {
 				key += ":" + baseMediaType(ct)
 			}
 			// Append a short fingerprint of the body so distinct payload shapes
@@ -141,7 +142,7 @@ func Deduplicate(classified []ClassifiedRequest) []ClassifiedRequest { //nolint:
 			// A collision would silently merge two distinct bodies into one dedup
 			// bucket; this is no worse than pre-fix behavior and worth the simpler key.
 			fingerprintBody := req.Body
-			if ct := getContentType(req.Headers); ct != "" {
+			if ct != "" {
 				if mt, params, err := mime.ParseMediaType(ct); err == nil && mt == "multipart/form-data" {
 					if boundary := params["boundary"]; len(boundary) >= 4 {
 						// Multipart bodies contain a random boundary token (per-request) that
--- a/pkg/classify/classifier_test.go
+++ b/pkg/classify/classifier_test.go
@@ -425,13 +425,15 @@ func TestDeduplicate_KeepsDistinctBodiesByContentType(t *testing.T) {
 }
 
 func TestDeduplicate_MergesSameContentType(t *testing.T) {
+	// Byte-identical bodies with the same CT must collapse — they are true duplicates.
+	body := []byte(`{"name":"alice"}`)
 	classified := []ClassifiedRequest{
 		{
 			ObservedRequest: crawl.ObservedRequest{
 				Method:  "POST",
 				URL:     "https://example.com/api/create",
 				Headers: map[string]string{"Content-Type": "application/json"},
-				Body:    []byte(`{"name":"first"}`),
+				Body:    body,
 			},
 			IsAPI: true, Confidence: 0.8, APIType: "rest",
 		},
@@ -440,14 +442,14 @@ func TestDeduplicate_MergesSameContentType(t *testing.T) {
 				Method:  "POST",
 				URL:     "https://example.com/api/create",
 				Headers: map[string]string{"Content-Type": "application/json"},
-				Body:    []byte(`{"name":"second"}`),
+				Body:    body,
 			},
 			IsAPI: true, Confidence: 0.9, APIType: "rest",
 		},
 	}
 
 	result := Deduplicate(classified)
-	require.Len(t, result, 1, "same content type on same path should collapse to one entry")
+	require.Len(t, result, 1, "byte-identical bodies on same path+CT should collapse to one entry")
 	assert.InDelta(t, 0.9, result[0].Confidence, 0.001, "highest confidence kept")
 }
 
@@ -497,4 +499,55 @@ func TestRunClassifiers_WSDLWinsOverREST(t *testing.T) {
 	require.Len(t, results, 1)
 	assert.Equal(t, "wsdl", results[0].APIType, "WSDL should win for SOAP traffic")
 	assert.GreaterOrEqual(t, results[0].Confidence, 0.90)
+}
+
+// TestDeduplicate_DistinctBodiesSurvive verifies that form-encoded POST observations
+// with distinct body bytes survive deduplication as separate entries, enabling
+// downstream buildOperation to union fields across observations (LAB-2106).
+func TestDeduplicate_DistinctBodiesSurvive(t *testing.T) {
+	ct := "application/x-www-form-urlencoded"
+	classified := []ClassifiedRequest{
+		{ObservedRequest: crawl.ObservedRequest{Method: "POST", URL: "https://example.com/api/checkout", Headers: map[string]string{"Content-Type": ct}, Body: []byte("product_id=1&qty=1")}, IsAPI: true, Confidence: 0.8},
+		{ObservedRequest: crawl.ObservedRequest{Method: "POST", URL: "https://example.com/api/checkout", Headers: map[string]string{"Content-Type": ct}, Body: []byte("product_id=2&coupon=SAVE10")}, IsAPI: true, Confidence: 0.8},
+		{ObservedRequest: crawl.ObservedRequest{Method: "POST", URL: "https://example.com/api/checkout", Headers: map[string]string{"Content-Type": ct}, Body: []byte("product_id=3&gift_wrap=true&note=hello")}, IsAPI: true, Confidence: 0.8},
+		{ObservedRequest: crawl.ObservedRequest{Method: "POST", URL: "https://example.com/api/checkout", Headers: map[string]string{"Content-Type": ct}, Body: []byte("product_id=4&address_id=99")}, IsAPI: true, Confidence: 0.8},
+		{ObservedRequest: crawl.ObservedRequest{Method: "POST", URL: "https://example.com/api/checkout", Headers: map[string]string{"Content-Type": ct}, Body: []byte("product_id=5&promo=FLASH")}, IsAPI: true, Confidence: 0.8},
+	}
+
+	result := Deduplicate(classified)
+	assert.Len(t, result, 5, "5 POST observations with distinct bodies must each survive dedup")
+}
+
+// TestDeduplicate_IdenticalBodiesCollapse verifies that POST observations with
+// byte-identical bodies collapse to a single entry (true duplicates).
+func TestDeduplicate_IdenticalBodiesCollapse(t *testing.T) {
+	body := []byte(`{"user":"alice","role":"admin"}`)
+	ct := "application/json"
+	classified := []ClassifiedRequest{
+		{ObservedRequest: crawl.ObservedRequest{Method: "POST", URL: "https://example.com/api/users", Headers: map[string]string{"Content-Type": ct}, Body: body}, IsAPI: true, Confidence: 0.8},
+		{ObservedRequest: crawl.ObservedRequest{Method: "POST", URL: "https://example.com/api/users", Headers: map[string]string{"Content-Type": ct}, Body: body}, IsAPI: true, Confidence: 0.85},
+		{ObservedRequest: crawl.ObservedRequest{Method: "POST", URL: "https://example.com/api/users", Headers: map[string]string{"Content-Type": ct}, Body: body}, IsAPI: true, Confidence: 0.9},
+		{ObservedRequest: crawl.ObservedRequest{Method: "POST", URL: "https://example.com/api/users", Headers: map[string]string{"Content-Type": ct}, Body: body}, IsAPI: true, Confidence: 0.7},
+		{ObservedRequest: crawl.ObservedRequest{Method: "POST", URL: "https://example.com/api/users", Headers: map[string]string{"Content-Type": ct}, Body: body}, IsAPI: true, Confidence: 0.75},
+	}
+
+	result := Deduplicate(classified)
+	require.Len(t, result, 1, "5 POST observations with byte-identical bodies must collapse to 1 entry")
+	assert.InDelta(t, 0.9, result[0].Confidence, 0.001, "highest confidence kept")
+}
+
+// TestDeduplicate_GETsStillCollapseByPath verifies that bodyless GETs continue
+// to deduplicate by path regardless of any header differences (unchanged behavior).
+func TestDeduplicate_GETsStillCollapseByPath(t *testing.T) {
+	classified := []ClassifiedRequest{
+		{ObservedRequest: crawl.ObservedRequest{Method: "GET", URL: "https://example.com/api/products?page=1", QueryParams: map[string]string{"page": "1"}}, IsAPI: true, Confidence: 0.8},
+		{ObservedRequest: crawl.ObservedRequest{Method: "GET", URL: "https://example.com/api/products?page=2", QueryParams: map[string]string{"page": "2"}}, IsAPI: true, Confidence: 0.85},
+		{ObservedRequest: crawl.ObservedRequest{Method: "GET", URL: "https://example.com/api/products?page=3", QueryParams: map[string]string{"page": "3"}}, IsAPI: true, Confidence: 0.75},
+	}
+
+	result := Deduplicate(classified)
+	require.Len(t, result, 1, "bodyless GETs must still collapse by path")
+	assert.InDelta(t, 0.85, result[0].Confidence, 0.001, "highest confidence kept")
+	// Query params from all 3 observations should be merged.
+	assert.Equal(t, "1", result[0].QueryParams["page"])
 }
--- a/pkg/classify/classifier_test.go
+++ b/pkg/classify/classifier_test.go
@@ -389,6 +389,94 @@ func TestDeduplicate_NonSOAPEndpointsStillDedupByPath(t *testing.T) {
 	assert.Len(t, result, 2, "REST deduplicates by path, WSDL separate")
 }
 
+func TestDeduplicate_KeepsDistinctBodiesByContentType(t *testing.T) {
+	classified := []ClassifiedRequest{
+		{
+			ObservedRequest: crawl.ObservedRequest{
+				Method:  "POST",
+				URL:     "https://example.com/api/submit",
+				Headers: map[string]string{"Content-Type": "application/json"},
+				Body:    []byte(`{"key":"value"}`),
+			},
+			IsAPI: true, Confidence: 0.9, APIType: "rest",
+		},
+		{
+			ObservedRequest: crawl.ObservedRequest{
+				Method:  "POST",
+				URL:     "https://example.com/api/submit",
+				Headers: map[string]string{"Content-Type": "application/x-www-form-urlencoded"},
+				Body:    []byte(`key=value`),
+			},
+			IsAPI: true, Confidence: 0.8, APIType: "rest",
+		},
+		{
+			ObservedRequest: crawl.ObservedRequest{
+				Method:  "POST",
+				URL:     "https://example.com/api/submit",
+				Headers: map[string]string{"Content-Type": "multipart/form-data; boundary=abc123"},
+				Body:    []byte(`--abc123\r\nContent-Disposition: form-data; name="key"\r\n\r\nvalue\r\n--abc123--`),
+			},
+			IsAPI: true, Confidence: 0.85, APIType: "rest",
+		},
+	}
+
+	result := Deduplicate(classified)
+	assert.Len(t, result, 3, "distinct content types on same path must not be merged")
+}
+
+func TestDeduplicate_MergesSameContentType(t *testing.T) {
+	classified := []ClassifiedRequest{
+		{
+			ObservedRequest: crawl.ObservedRequest{
+				Method:  "POST",
+				URL:     "https://example.com/api/create",
+				Headers: map[string]string{"Content-Type": "application/json"},
+				Body:    []byte(`{"name":"first"}`),
+			},
+			IsAPI: true, Confidence: 0.8, APIType: "rest",
+		},
+		{
+			ObservedRequest: crawl.ObservedRequest{
+				Method:  "POST",
+				URL:     "https://example.com/api/create",
+				Headers: map[string]string{"Content-Type": "application/json"},
+				Body:    []byte(`{"name":"second"}`),
+			},
+			IsAPI: true, Confidence: 0.9, APIType: "rest",
+		},
+	}
+
+	result := Deduplicate(classified)
+	require.Len(t, result, 1, "same content type on same path should collapse to one entry")
+	assert.InDelta(t, 0.9, result[0].Confidence, 0.001, "highest confidence kept")
+}
+
+func TestDeduplicate_GETsCollapseIgnoringContentType(t *testing.T) {
+	classified := []ClassifiedRequest{
+		{
+			ObservedRequest: crawl.ObservedRequest{
+				Method:  "GET",
+				URL:     "https://example.com/api/items",
+				Headers: map[string]string{"Content-Type": "application/json"},
+				// No body
+			},
+			IsAPI: true, Confidence: 0.8, APIType: "rest",
+		},
+		{
+			ObservedRequest: crawl.ObservedRequest{
+				Method:  "GET",
+				URL:     "https://example.com/api/items?page=2",
+				Headers: map[string]string{"Content-Type": "application/xml"},
+				// No body
+			},
+			IsAPI: true, Confidence: 0.85, APIType: "rest",
+		},
+	}
+
+	result := Deduplicate(classified)
+	require.Len(t, result, 1, "empty-body GETs should collapse by path regardless of Content-Type header")
+}
+
 func TestRunClassifiers_WSDLWinsOverREST(t *testing.T) {
 	classifiers := []APIClassifier{
 		&RESTClassifier{},
--- a/pkg/classify/classifier_test.go
+++ b/pkg/classify/classifier_test.go
@@ -15,6 +15,7 @@
 package classify
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -341,6 +342,45 @@ func TestDeduplicate_MergesSameSOAPAction(t *testing.T) {
 	assert.InDelta(t, 0.95, result[0].Confidence, 0.001, "highest confidence kept")
 }
 
+// TestDeduplicate_SOAPDistinctBodiesSurvive verifies that two SOAP requests
+// to the same endpoint and SOAPAction with DIFFERENT envelope bytes are
+// preserved as separate entries — the new body-fingerprint behavior. This
+// makes the contrast with TestDeduplicate_MergesSameSOAPAction (which uses
+// empty bodies) explicit, so future readers don't think the bodyless test
+// implies bodies are also collapsed.
+func TestDeduplicate_SOAPDistinctBodiesSurvive(t *testing.T) {
+	classified := []ClassifiedRequest{
+		{
+			ObservedRequest: crawl.ObservedRequest{
+				Method: "POST",
+				URL:    "https://example.com/service",
+				Headers: map[string]string{
+					"SOAPAction":   `"urn:GetUser"`,
+					"Content-Type": "text/xml; charset=utf-8",
+				},
+				Body: []byte(`<env:Envelope xmlns:env="..."><env:Body><GetUser><id>1</id></GetUser></env:Body></env:Envelope>`),
+			},
+			IsAPI: true, Confidence: 0.85, APIType: "wsdl",
+		},
+		{
+			ObservedRequest: crawl.ObservedRequest{
+				Method: "POST",
+				URL:    "https://example.com/service",
+				Headers: map[string]string{
+					"SOAPAction":   `"urn:GetUser"`,
+					"Content-Type": "text/xml; charset=utf-8",
+				},
+				Body: []byte(`<env:Envelope xmlns:env="..."><env:Body><GetUser><id>2</id></GetUser></env:Body></env:Envelope>`),
+			},
+			IsAPI: true, Confidence: 0.90, APIType: "wsdl",
+		},
+	}
+
+	result := Deduplicate(classified)
+	require.Len(t, result, 2,
+		"distinct SOAP envelope bodies on same path+SOAPAction should survive (allows downstream merge to see all observations)")
+}
+
 func TestDeduplicate_SOAPActionCaseInsensitive(t *testing.T) {
 	classified := []ClassifiedRequest{
 		{
@@ -550,4 +590,31 @@ func TestDeduplicate_GETsStillCollapseByPath(t *testing.T) {
 	assert.InDelta(t, 0.85, result[0].Confidence, 0.001, "highest confidence kept")
 	// Query params from all 3 observations should be merged.
 	assert.Equal(t, "1", result[0].QueryParams["page"])
+}
+
+// BenchmarkDeduplicate exercises the dedup hot path (key construction +
+// body fingerprint when applicable) at varying scales. Establishes a
+// baseline before stretch-goal performance work.
+func BenchmarkDeduplicate(b *testing.B) {
+	for _, n := range []int{100, 1000, 5000} {
+		b.Run(fmt.Sprintf("n=%d", n), func(b *testing.B) {
+			// Build a fixture: n distinct POST observations, urlencoded bodies
+			input := make([]ClassifiedRequest, n)
+			for i := 0; i < n; i++ {
+				input[i] = ClassifiedRequest{
+					ObservedRequest: crawl.ObservedRequest{
+						Method:  "POST",
+						URL:     fmt.Sprintf("https://example.com/api/endpoint%d", i%50),
+						Headers: map[string]string{"Content-Type": "application/x-www-form-urlencoded"},
+						Body:    []byte(fmt.Sprintf("name=u%d&value=%d", i, i)),
+					},
+					IsAPI: true, Confidence: 0.9, APIType: "rest",
+				}
+			}
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				_ = Deduplicate(input)
+			}
+		})
+	}
 }
--- a/pkg/classify/classifier_test.go
+++ b/pkg/classify/classifier_test.go
@@ -592,6 +592,42 @@ func TestDeduplicate_GETsStillCollapseByPath(t *testing.T) {
 	assert.Equal(t, "1", result[0].QueryParams["page"])
 }
 
+// TestDeduplicate_MultipartBoundaryNormalized verifies that two logically
+// identical multipart bodies with DIFFERENT boundary tokens dedup correctly.
+// Boundaries are random per-request from clients, so without normalization
+// every observation would be unique.
+func TestDeduplicate_MultipartBoundaryNormalized(t *testing.T) {
+	body1 := []byte("--BoundaryAAA\r\n" +
+		"Content-Disposition: form-data; name=\"x\"\r\n\r\n" +
+		"value\r\n" +
+		"--BoundaryAAA--\r\n")
+	body2 := []byte("--BoundaryZZZ\r\n" +
+		"Content-Disposition: form-data; name=\"x\"\r\n\r\n" +
+		"value\r\n" +
+		"--BoundaryZZZ--\r\n")
+	classified := []ClassifiedRequest{
+		{
+			ObservedRequest: crawl.ObservedRequest{
+				Method: "POST", URL: "https://example.com/api/upload",
+				Headers: map[string]string{"Content-Type": "multipart/form-data; boundary=BoundaryAAA"},
+				Body:    body1,
+			},
+			IsAPI: true, Confidence: 0.85, APIType: "rest",
+		},
+		{
+			ObservedRequest: crawl.ObservedRequest{
+				Method: "POST", URL: "https://example.com/api/upload",
+				Headers: map[string]string{"Content-Type": "multipart/form-data; boundary=BoundaryZZZ"},
+				Body:    body2,
+			},
+			IsAPI: true, Confidence: 0.95, APIType: "rest",
+		},
+	}
+	result := Deduplicate(classified)
+	require.Len(t, result, 1, "identical multipart content with different boundaries should dedup to one entry")
+	assert.InDelta(t, 0.95, result[0].Confidence, 0.001, "highest confidence kept")
+}
+
 // BenchmarkDeduplicate exercises the dedup hot path (key construction +
 // body fingerprint when applicable) at varying scales. Establishes a
 // baseline before stretch-goal performance work.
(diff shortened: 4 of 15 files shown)


santiago-praetorian (changes requested):
**PR:** https://github.com/praetorian-inc/vespasian/pull/101  
**Linear:** https://linear.app/praetorianlabs/issue/LAB-2106  
**Head SHA:** `8608ddd6477762af29d3b2fdd40854f53ade42b7`  
**Reviewed at:** 2026-04-27T15:01:59.221564+00:00

**Recommendation:** **REQUEST CHANGES**

|---|---|---|---|---|---|
| Count | 0 | 1 | 4 | 8 | 5 |
| Blockers | 0 | 1 | 4 | 0 | 0 |

---

## Blockers

### QUAL-005 — medium — QUAL

**File:** `pkg/generate/rest/openapi.go:646-677`

**Issue:** In extractComponents, the loop `for _, mediaType := range reqBody.Content` iterates over a plain Go map (openapi3.Content is map[string]*MediaType). When a single endpoint has both 'application/json' and 'multipart/form-data' (or urlencoded) request bodies — a reachable scenario after this PR's Deduplicate change preserves distinct-content-type observations — the iteration order is non-deterministic. Both schemas have different fingerprints, so both go through ensureUniqueName with baseName 'CreateXxxRequest'. The order they are encountered determines which gets 'CreateXxxRequest' and which gets 'CreateXxxRequest2'. This produces non-deterministic output for multi-content-type POST endpoints, the primary new use case introduced by this PR. The existing TestExtractComponents_Deterministic test does not cover this scenario (it only uses single JSON content type).

**Recommendation:** Sort the media type keys from reqBody.Content before iterating, the same way sortedPaths is sorted for outer path iteration. Add a determinism test that sends a single endpoint with both application/json and application/x-www-form-urlencoded observations to verify stable component names across repeated Generate calls.

```
for _, mediaType := range reqBody.Content {
	if mediaType == nil || mediaType.Schema == nil {
		continue
	}
	if schema := mediaType.Schema.Value; schema != nil && schema.Properties != nil {
		fingerprint := schemaFingerprint(schema)
		if fingerprint != "" {
			// ...
			baseName := methodPrefix + resourceName + "Request"
			componentName = ensureUniqueName(baseName)
```

---

### SEC-BE-002 — medium — SEC-BE

**File:** `pkg/generate/rest/form.go:72-93`

**Issue:** ParseMultipartForm calls reader.NextPart() in an unbounded for-loop. mime/multipart's CVE-2023-24536 mitigation only enforces a 10,000-header cap per part — there is no cap on the number of parts. An adversarially-crafted multipart body with millions of trivially-named small parts will allocate per-part state until the operator's process OOMs at spec-generation time.

**Recommendation:** 

_CVSS:_ `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:H` (score 5.5)

---

### SEC-BE-003 — medium — SEC-BE

**File:** `pkg/generate/rest/form.go:86-88`

**Issue:** Inside ParseMultipartForm, io.Copy reads each non-file part fully into a strings.Builder for type inference. There is no per-part byte cap. A hostile multipart body with one or two text parts of GB-scale will allocate the full body in memory. Type inference only needs a small prefix (a few KB), so the unbounded read serves no business purpose.

**Recommendation:** 

_CVSS:_ `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:H` (score 5.5)

---

### TEST-004 — medium — TEST

**File:** `pkg/generate/rest/openapi.go:125-150`

**Issue:** openapi.go:125 mergeJSONBodies is the JSON sibling of mergeURLEncodedBodies/mergeMultipartBodies (form.go:111, form.go:123) and is reachable from buildOperation at openapi.go:268 (default case). The form variants both have direct unit tests for conflicting types: TestMergeURLEncodedBodies_ConflictingTypes (form_test.go:631) asserts that integer-vs-string is promoted to `string`. mergeJSONBodies has no direct unit test, and its conflict semantics differ deliberately: it keeps the first observation's property and ignores the second (`if _, exists := merged.Value.Properties[propName]; !exists` at openapi.go:139), so a schema-shape conflict on the same key is silently dropped rather than promoted. TestOpenAPIGenerator_SchemaMerging at openapi_test.go:344 only exercises the union-of-disjoint-properties case and does not write the same property name with a conflicting type in two observations. The PR description explicitly calls out this divergence; without a test, future refactors cannot tell whether the divergence is intentional or accidental, and a regression that swaps the two strategies would not be caught.

**Recommendation:** Add TestMergeJSONBodies in pkg/generate/rest/openapi_test.go with at least three sub-cases mirroring the form merger tests: (a) union of disjoint properties across two observations; (b) same property with conflicting inferred types in observations 1 and 2 — assert the FIRST observation's type wins (documenting the intentional divergence from the form merger's string-promotion); (c) nil/empty body slices skipped without panicking. The third sub-case can be a single-line table entry. Place it adjacent to TestMergeMultipartBodies for discoverability.

```
// mergeJSONBodies infers and merges JSON schemas from multiple body observations.
func mergeJSONBodies(bodies [][]byte) *openapi3.SchemaRef {
		if merged.Value != nil && merged.Value.Properties != nil &&
			schema.Value != nil && schema.Value.Properties != nil {
			for propName, propSchema := range schema.Value.Properties {
				if _, exists := merged.Value.Properties[propName]; !exists {
```

---

### TEST-005 — high — TEST

**File:** `pkg/generate/rest/openapi_test.go:1244-1277`

**Issue:** The regression test for the determinism fix at extractComponents builds 26 endpoints, each with a single Content-Type (application/json) and a single status code (201). The fix only addresses the outer Paths.Map() iteration via sort.Strings(sortedPaths) at openapi.go:610. Two map iterations remain unsorted: (1) `for _, mediaType := range reqBody.Content` at openapi.go:646 and (2) `for statusCode := range op.operation.Responses.Map()` at openapi.go:682. I confirmed both are still non-deterministic by adding two probe tests in this worktree: a single POST endpoint with both application/json and application/x-www-form-urlencoded request bodies (distinct fingerprints) produced 23/99 differing component-name assignments across runs; a single GET path with 200 and 201 response schemas sharing a fingerprint produced 27/99 differing runs. The existing test cannot detect either gap because every endpoint has only one entry in reqBody.Content and Responses.Map(). Both Claude and Gemini bot reviewers flagged the request-body iteration; this rubric requires a regression test that actually exercises the unfixed branches.

**Recommendation:** Extend TestExtractComponents_Deterministic with two additional sub-cases (or add two new tests). Sub-case A: a single POST endpoint observed with two distinct content types (application/json + application/x-www-form-urlencoded) carrying distinct field shapes; assert that across 50+ Generate runs the $ref values bound to each media type are byte-identical. Sub-case B: a path observed with multiple response status codes whose schemas share a fingerprint (e.g., 200 and 201 with identical body shape); assert the Components.Schemas key chosen on first encounter is stable across runs. Both sub-cases will fail on the current openapi.go:646 and openapi.go:682 iterations, forcing those map walks to be sorted (sort the keys of reqBody.Content and Responses.Map() before iterating, mirroring the pattern already used for sortedPaths at openapi.go:609).

```
func TestExtractComponents_Deterministic(t *testing.T) {
	endpoints := make([]classify.ClassifiedRequest, 0, 26)
	for c := 'a'; c <= 'z'; c++ {
				Headers: map[string]string{"Content-Type": "application/json"},
				Response: crawl.ObservedResponse{
					StatusCode: 201,
```

---

## Non-blocking

### QUAL-001 — nit — QUAL

**File:** `pkg/classify/classifier.go:187-196`

**Issue:** getContentType (lines 187-195) is byte-for-byte identical to getContentType in pkg/generate/graphql/infer.go (line 2135). Both perform a case-insensitive scan for the Content-Type header. These exist in different packages and cannot directly share the implementation without creating a new shared package, so this is not a blocking DRY violation. However, a shared crawl-level or utility-level helper could avoid the pattern if it ever grows to three or more copies.

**Recommendation:** Document with a comment (or a TODO) that this duplicates the same helper in pkg/generate/graphql/infer.go. Consider extracting to a shared internal utility package if a third copy arises.

```
func getContentType(headers map[string]string) string {
	for k, v := range headers {
		if strings.EqualFold(k, "content-type") {
			return v
		}
	}
	return ""
}
```

---

### QUAL-002 — low — QUAL

**File:** `pkg/generate/rest/openapi.go:127-150`

**Issue:** mergeJSONBodies (lines 127-150) was extracted from the pre-existing body-merge loop in buildOperation during this refactor. The function is 24 lines of logic that is now called from a single site. The extraction is correct and improves readability, but mergeJSONBodies does NOT deduplicate property conflicts the same way mergeObjectSchemas does (it does a union-only merge with no conflict promotion to string). This creates an inconsistency: URL-encoded and multipart observations use mergeObjectSchemas (which promotes conflicting types to string), while JSON observations use mergeJSONBodies (which silently ignores a second observation of the same property even if the types differ). This is a subtle behavioral inconsistency, not a crash, but could result in incorrect type inference for JSON endpoints where the same field alternates types across observations.

**Recommendation:** For consistency, mergeJSONBodies should also detect type conflicts and promote to string, matching the behavior of mergeObjectSchemas. Alternatively, add a comment to mergeJSONBodies explicitly documenting that it does not promote conflicting types, and add a test that exercises this difference.

```
func mergeJSONBodies(bodies [][]byte) *openapi3.SchemaRef {
	if merged.Value != nil && merged.Value.Properties != nil &&
		schema.Value != nil && schema.Value.Properties != nil {
		for propName, propSchema := range schema.Value.Properties {
			if _, exists := merged.Value.Properties[propName]; !exists {
				merged.Value.Properties[propName] = propSchema
			}
		}
	}
```

---

### QUAL-003 — nit — QUAL

**File:** `pkg/generate/rest/openapi.go:243-249`

**Issue:** The inline content-type normalization in buildOperation (lines 243-249: `strings.ToLower(strings.TrimSpace(strings.SplitN(ct, ";", 2)[0]))`) duplicates the semantics of baseMediaType in pkg/classify/classifier.go (lines 199-207). Both strip parameters and lowercase. They are in different packages so cannot directly call each other, but the pattern is identical. This is the second new duplication of this logic pattern (baseMediaType itself was added in this PR).

**Recommendation:** Consider extracting a shared media type normalizer to pkg/crawl or a new pkg/httputil package. As a minimum, extract to a local helper inside the rest package so that openapi.go and form.go share one implementation.

```
trimmed := strings.ToLower(strings.TrimSpace(strings.SplitN(ct, ";", 2)[0]))
if trimmed != "" {
    baseType = trimmed
}
```

---

### QUAL-004 — low — QUAL

**File:** `pkg/generate/rest/openapi.go:498-518`

**Issue:** toCamelCase processes runes but only handles ASCII lowercase (r >= 'a' && r <= 'z') for capitalization. Non-ASCII lowercase letters (e.g., accented characters like 'é', 'ñ') fall through to the `default: capitalizeNext = true` branch, which causes them to be silently dropped from the output and treated as word-separator characters. A path like '/résumés' would produce component name 'R' (only the first ASCII char before the accent). This differs from capitalizeFirst (also in this file) which correctly uses unicode.ToUpper. The behavior is not documented.

**Recommendation:** Add a comment to toCamelCase noting that it is intentionally ASCII-only (OpenAPI component names are typically ASCII in practice). Alternatively, use unicode.IsLetter and unicode.IsUpper/unicode.ToUpper for proper Unicode handling, consistent with capitalizeFirst in the same file.

```
func toCamelCase(s string) string {
	for _, r := range s {
		switch {
		case r >= 'a' && r <= 'z':
			if capitalizeNext {
				r = r - 'a' + 'A'
			}
			b.WriteRune(r)
			capitalizeNext = false
		case (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9'):
			b.WriteRune(r)
			capitalizeNext = false
		default:
			capitalizeNext = true
		}
```

---

### SEC-BE-001 — low — SEC-BE

**File:** `pkg/generate/rest/form.go:70-70`

**Issue:** strings.NewReader(string(body)) duplicates the entire body in memory. multipart.NewReader accepts any io.Reader, so bytes.NewReader(body) avoids the allocation entirely. Not a security vulnerability on its own, but compounds the prior two findings under memory pressure.

**Recommendation:** 

_CVSS:_ `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:L` (score 3.3)

---

## Ticket alignment

_All Linear LAB-2106 acceptance criteria appear to be implemented in this PR._

---

## Tests & documentation

### TEST-004 — medium — TEST

**File:** `pkg/generate/rest/openapi.go:125`

**Issue:** openapi.go:125 mergeJSONBodies is the JSON sibling of mergeURLEncodedBodies/mergeMultipartBodies (form.go:111, form.go:123) and is reachable from buildOperation at openapi.go:268 (default case). The form variants both have direct unit tests for conflicting types: TestMergeURLEncodedBodies_ConflictingTypes (form_test.go:631) asserts that integer-vs-string is promoted to `string`. mergeJSONBodies has no direct unit test, and its conflict semantics differ deliberately: it keeps the first observation's property and ignores the second (`if _, exists := merged.Value.Properties[propName]; !exists` at openapi.go:139), so a schema-shape conflict on the same key is silently dropped rather than promoted. TestOpenAPIGenerator_SchemaMerging at openapi_test.go:344 only exercises the union-of-disjoint-properties case and does not write the same property name with a conflicting type in two observations. The PR description explicitly calls out this divergence; without a test, future refactors cannot tell whether the divergence is intentional or accidental, and a regression that swaps the two strategies would not be caught.

**Recommendation:** Add TestMergeJSONBodies in pkg/generate/rest/openapi_test.go with at least three sub-cases mirroring the form merger tests: (a) union of disjoint properties across two observations; (b) same property with conflicting inferred types in observations 1 and 2 — assert the FIRST observation's type wins (documenting the intentional divergence from the form merger's string-promotion); (c) nil/empty body slices skipped without panicking. The third sub-case can be a single-line table entry. Place it adjacent to TestMergeMultipartBodies for discoverability.

```
// mergeJSONBodies infers and merges JSON schemas from multiple body observations.
func mergeJSONBodies(bodies [][]byte) *openapi3.SchemaRef {
		if merged.Value != nil && merged.Value.Properties != nil &&
			schema.Value != nil && schema.Value.Properties != nil {
			for propName, propSchema := range schema.Value.Properties {
				if _, exists := merged.Value.Properties[propName]; !exists {
```

---

### TEST-005 — high — TEST

**File:** `pkg/generate/rest/openapi_test.go:1244`

**Issue:** The regression test for the determinism fix at extractComponents builds 26 endpoints, each with a single Content-Type (application/json) and a single status code (201). The fix only addresses the outer Paths.Map() iteration via sort.Strings(sortedPaths) at openapi.go:610. Two map iterations remain unsorted: (1) `for _, mediaType := range reqBody.Content` at openapi.go:646 and (2) `for statusCode := range op.operation.Responses.Map()` at openapi.go:682. I confirmed both are still non-deterministic by adding two probe tests in this worktree: a single POST endpoint with both application/json and application/x-www-form-urlencoded request bodies (distinct fingerprints) produced 23/99 differing component-name assignments across runs; a single GET path with 200 and 201 response schemas sharing a fingerprint produced 27/99 differing runs. The existing test cannot detect either gap because every endpoint has only one entry in reqBody.Content and Responses.Map(). Both Claude and Gemini bot reviewers flagged the request-body iteration; this rubric requires a regression test that actually exercises the unfixed branches.

**Recommendation:** Extend TestExtractComponents_Deterministic with two additional sub-cases (or add two new tests). Sub-case A: a single POST endpoint observed with two distinct content types (application/json + application/x-www-form-urlencoded) carrying distinct field shapes; assert that across 50+ Generate runs the $ref values bound to each media type are byte-identical. Sub-case B: a path observed with multiple response status codes whose schemas share a fingerprint (e.g., 200 and 201 with identical body shape); assert the Components.Schemas key chosen on first encounter is stable across runs. Both sub-cases will fail on the current openapi.go:646 and openapi.go:682 iterations, forcing those map walks to be sorted (sort the keys of reqBody.Content and Responses.Map() before iterating, mirroring the pattern already used for sortedPaths at openapi.go:609).

```
func TestExtractComponents_Deterministic(t *testing.T) {
	endpoints := make([]classify.ClassifiedRequest, 0, 26)
	for c := 'a'; c <= 'z'; c++ {
				Headers: map[string]string{"Content-Type": "application/json"},
				Response: crawl.ObservedResponse{
					StatusCode: 201,
```

---

### TEST-001 — low — TEST

**File:** `pkg/classify/classifier_test.go:354`

**Issue:** TestDeduplicate_SOAPDistinctBodiesSurvive (classifier_test.go:354) asserts only `require.Len(t, result, 2)` — that two entries survive — without asserting that the two surviving entries actually carry the two distinct body bytes. A future regression where Deduplicate accidentally returns two copies of the same body (e.g., a mutation bug in the body-fingerprint key construction at classifier.go:139) would still pass this assertion. The same loose-assertion shape appears in TestDeduplicate_DistinctBodiesSurvive (classifier_test.go:539) which only checks `assert.Len(t, result, 5)`. The body-fingerprint behavior is the headline tier-2 fix in this PR; the assertion should be tight enough to catch a payload-mixup regression.

**Recommendation:** Strengthen both tests to assert the set of body-bytes in the result equals the input set. For SOAPDistinctBodiesSurvive: collect `string(r.Body) for r in result` into a set and assert membership of both the `<id>1</id>` and `<id>2</id>` envelopes. For DistinctBodiesSurvive: build expected/actual body slices and assert.ElementsMatch. This is a 5-line change per test that converts existence-of-count assertions into specificity-of-content assertions, matching rubric item 4.

```
func TestDeduplicate_SOAPDistinctBodiesSurvive(t *testing.T) {
	result := Deduplicate(classified)
	require.Len(t, result, 2,
		"distinct SOAP envelope bodies on same path+SOAPAction should survive (allows downstream merge to see all observations)")
```

---

### TEST-002 — nit — TEST

**File:** `pkg/generate/rest/form_test.go:200`

**Issue:** The third sub-case `missing content-type entry skips that observation` at form_test.go:200 passes a single-element bodies slice and an empty contentTypes slice, then asserts the result is nil. mergeMultipartBodies at form.go:123 is implemented as `if i < len(contentTypes) { ct = contentTypes[i] }` — so the test only exercises the path where ALL observations lack a content type. The partial case (2 bodies, 1 content type) — where observation 0 succeeds and observation 1 silently uses an empty boundary — is untested. This is the realistic ‘some entries lost their headers during capture replay’ shape; rubric item 2 calls for at least one test per explicit branch.

**Recommendation:** Add a fourth sub-case to TestMergeMultipartBodies: pass `[][]byte{validBody1, validBody2}` and `[]string{ct1}` (length-1), and assert the merged schema contains only the properties from validBody1 (validBody2 is silently skipped because its boundary resolves to empty). This documents the bounds-check behavior and pins the silent-skip contract.

```
	t.Run("missing content-type entry skips that observation", func(t *testing.T) {
		// Provide body but no corresponding content-type entry.
		schema := mergeMultipartBodies([][]byte{buf.Bytes()}, []string{})
		// With no boundary the observation is skipped; result is nil.
		if schema != nil {
```

---

### TEST-003 — low — TEST

**File:** `pkg/generate/rest/form_test.go:436`

**Issue:** TestParseMultipartForm_MultipleFilesSameName at form_test.go:436 documents that two parts sharing `name="files"` collapse to a single `files` property (the second part overwrites the first). This is fine as a parser-level contract test, but the rubric calls out per-branch coverage of new exported behavior. The exported entry point ParseMultipartForm + the internal mergeMultipartBodies + buildOperation chain produces the schema that ships in the OpenAPI spec; if a real-world upload endpoint multi-attaches files the spec will silently drop one. There is no end-to-end test (Generate-level) that asserts the spec emitted for a `multipart/form-data` body with repeated-name file parts contains the documented single-property shape. TestOpenAPIGenerator_MultipartFormData_EndToEnd (openapi_test.go:1053) covers only the single-text-field-plus-single-file shape. Adding an E2E variant would lock in the behavior at the public boundary, so a future change that silently introduces array-of-files would be caught.

**Recommendation:** Add a sub-case to TestOpenAPIGenerator_MultipartFormData_EndToEnd (or a sibling test) that constructs a multipart body with two `name="files"` file parts, runs gen.Generate, and asserts the resulting requestBody.content["multipart/form-data"].schema.properties contains exactly one `files` property with format binary. Pair with a comment pointing at form.go:81 (the overwrite site) so reviewers know the test guards the documented lossy behavior, not a bug.

```
func TestParseMultipartForm_MultipleFilesSameName(t *testing.T) {
	prop, ok := schema.Value.Properties["files"]
	assert.True(t, ok, "expected property 'files' to exist")
		// Current behavior: second file overwrites first; format is still binary.
		assert.Equal(t, "binary", prop.Value.Format)
```

---

### TEST-006 — low — TEST

**File:** `test/rest-api/expected-spec.yaml`

**Issue:** expected-paths.json was updated to include /api/login (POST) and /api/upload (POST), and main.go gained handlers for both — but reference-capture.json contains no urlencoded or multipart traffic and expected-spec.yaml has no /api/login or /api/upload entries. The live `test_generate_rest` (which compares the generated YAML to expected-spec.yaml) therefore does not enforce that the new form-body codepath produces stable, correct OpenAPI requestBody output end-to-end. Unit tests cover the parsers and TestOpenAPIGenerator_MultipartFormData_EndToEnd covers Generate-level multipart, but the deterministic golden-file integration test is missing form coverage.

**Recommendation:** Either (a) extend reference-capture.json with one urlencoded and one multipart capture and add the corresponding paths/components to expected-spec.yaml, or (b) document explicitly in test/run-live-tests.sh and the PR description that the live integration golden does not pin form output and is intentionally left to unit-level coverage.

---

### DOC-001 — nit — DOC

**File:** `CLAUDE.md`

**Issue:** CLAUDE.md describes pkg/generate/rest as 'OpenAPI 3.0 generation, path normalization (UUID detection, context-aware parameter naming), JSON schema inference.' That bullet is now incomplete — form/multipart request-body inference is also part of the package. The pkg/classify bullet ('deduplication') is general enough to be still correct. Per CLAUDE.md's own guidance ('After adding or changing features, review README.md and CLAUDE.md for accuracy and update them if needed'), this should be reflected.

**Recommendation:** Append 'form-encoded and multipart request-body inference' to the pkg/generate/rest bullet in the Key Packages section.

```
- **pkg/generate/rest**: OpenAPI 3.0 generation, path normalization (UUID detection, context-aware parameter naming), JSON schema inference
```

---

### DOC-002 — nit — DOC

**File:** `README.md:177`

**Issue:** The README capabilities table lists 'JSON schema inference' as the schema-inference offering for REST. The PR adds inference for application/x-www-form-urlencoded and multipart/form-data bodies as well; advertising 'JSON' specifically in the user-facing README will mislead users into thinking only JSON request bodies are documented in generated specs.

**Recommendation:** Update the row (or the Active Probing summary on line 47) to mention form-encoded and multipart request-body inference alongside JSON.

```
| **REST** | JSON/XML content-type, `/api/` `/v1/` path patterns, HTTP methods | OpenAPI 3.0 (YAML/JSON) | OPTIONS discovery, JSON schema inference |
```

---

### DOC-003 — low — DOC

**File:** `pkg/classify/doc.go`

**Issue:** doc.go states '[Deduplicate] removes duplicate endpoints based on method and normalized URL.' The PR changed Deduplicate to also factor content-type and an 8-byte SHA-256 body fingerprint (when body is non-empty) into the dedup key — bodyless GETs still collapse, but distinct bodies on the same (method, path, content-type) now survive. Reading the current doc string would mislead a future maintainer about Deduplicate's contract.

**Recommendation:** Update the Deduplicate sentence to reflect: (method, normalized URL, content-type, body fingerprint) keying with bodyless requests collapsing as before.

```
// [Deduplicate] removes duplicate endpoints based on method and normalized URL.
```

---

### DOC-004 — low — DOC

**File:** `pkg/generate/rest/doc.go`

**Issue:** The PR adds two new exported functions (ParseURLEncodedForm, ParseMultipartForm) plus form merge helpers in pkg/generate/rest/form.go, but the package-level doc.go was not updated. The current text says the package handles 'OpenAPI 3.0 specifications', 'path normalization', 'UUID detection', and 'JSON schema inference from response bodies' — request-body inference for form-encoded and multipart traffic is now part of the public surface and is missing. CLAUDE.md (line 137) explicitly requires updating doc.go when the package's public API or purpose changes.

**Recommendation:** Add a bullet (or sentence) under 'Key components' describing form-body parsing: ParseURLEncodedForm and ParseMultipartForm, distinguishing file-upload fields (format: binary) from text fields, plus the per-content-type merge behavior used by buildOperation.

```
//   - Schema inference examines response JSON to generate OpenAPI schema
//     objects with depth and property guards.
package rest
```

---

0 findings were filtered (no PR threads with explicit resolution markers detected).

---

maneelmodi:
@santiago-praetorian — thanks for the thorough review. Worth flagging that your review was based on commit `8608ddd6`, which preceded the fix commit `2a3f15c` (pushed ~26 min before your review submitted). Of your 5 blockers, 4 were already addressed there, plus we just shipped `b9c0113` (this commit) addressing the remaining one and your non-blocking findings. Status of every finding against current HEAD `b9c0113`:

## Blockers

| ID | Status | Notes |
|---|---|---|
| **QUAL-005** (medium): Inner-map non-determinism in `extractComponents` | ✅ Already fixed in `2a3f15c` (preceded your review) | Sorted `reqBody.Content`, `response.Content`, AND status codes. `TestExtractComponents_DeterministicMultiContentType` (5 paths × 2 content types × 5 runs byte-identical) covers it. |
| **SEC-BE-002** (medium): Multipart parts unbounded loop | ✅ Fixed in `b9c0113` | `maxMultipartParts = 1000` cap; `TestParseMultipartForm_PartCountCapped` writes 1001 parts and asserts ≤1000 are processed. |
| **SEC-BE-003** (medium): Text part unbounded read | ✅ Already fixed in `2a3f15c` | `io.LimitReader(part, 4096)` in `ParseMultipartForm`. |
| **TEST-004** (medium): `mergeJSONBodies` test missing | ⚠️ See below — design discussion | We intentionally unified `mergeJSONBodies` to delegate to `mergeObjectSchemas` (commit `2a3f15c`), so it now promotes conflicting types to `string` like the form merger does. Your suggested test asserts "FIRST observation's type wins" — that contradicts the unification. `TestMergeJSONBodies_TypeConflictPromotesToString` covers the unified behavior. **Open question for you: prefer unified semantics (current) or keep JSON divergent? Happy to revert the unification if you prefer "first wins" — please advise.** |
| **TEST-005** (high): Multi-CT determinism test | ✅ Already fixed in `2a3f15c` | `TestExtractComponents_DeterministicMultiContentType` exercises the previously-uncovered inner-map paths. |

## Non-blocking findings

| ID | Severity | Status |
|---|---|---|
| QUAL-001 | nit | **Deferred** — `getContentType` cross-package duplication with `pkg/generate/graphql/infer.go` requires creating a new shared utility package. Out of LAB-2106 scope; tracked for a separate refactor PR. |
| QUAL-002 | low | ✅ Fixed in `2a3f15c` (`mergeJSONBodies` now delegates to `mergeObjectSchemas`). |
| QUAL-003 | nit | **Deferred** — same reason as QUAL-001 (would need shared package). |
| QUAL-004 | low | ✅ Fixed in `b9c0113` — added doc comment to `toCamelCase` noting the ASCII-only behavior is intentional. |
| SEC-BE-001 | low | ✅ Already fixed in `2a3f15c` (`bytes.NewReader(body)`). |
| TEST-001 | low | ✅ Fixed in `b9c0113` — `TestDeduplicate_DistinctBodiesSurvive` and `TestDeduplicate_SOAPDistinctBodiesSurvive` now use `assert.ElementsMatch` over actual body bytes. |
| TEST-002 | nit | ✅ Fixed in `b9c0113` — added "partial content-types — second observation skipped" subtest to `TestMergeMultipartBodies`. |
| TEST-003 | low | ✅ Fixed in `b9c0113` — `TestOpenAPIGenerator_MultipartRepeatedFileFields_E2E` pins the last-wins behavior end-to-end. |
| TEST-006 | low | ✅ Fixed in `b9c0113` — `reference-capture.json` and `expected-spec.yaml` now contain one urlencoded entry (`/api/login`) and one multipart entry (`/api/upload` with `format: binary`), so the deterministic golden `test_generate_rest` pins form-body output. |
| DOC-001 | nit | ✅ Fixed in `b9c0113` (CLAUDE.md). |
| DOC-002 | nit | ✅ Fixed in `b9c0113` (README.md REST row). |
| DOC-003 | low | ✅ Fixed in `b9c0113` (`pkg/classify/doc.go` Deduplicate description). |
| DOC-004 | low | ✅ Fixed in `b9c0113` (`pkg/generate/rest/doc.go`). |

## CodeRabbit's outside-diff finding

`fingerprintToName` was shared between request- and response-body extraction in `extractComponents` — echo-style endpoints could have responses tagged as `…Request`. ✅ Fixed in `b9c0113` — `fingerprintToReqName` and `fingerprintToRespName` are now separate maps; `nameCounter` stays shared for global collision avoidance. `TestExtractComponents_RequestResponseScopedRefs` covers it.

## Verification on `b9c0113`

- 482 tests pass (full repo, `go test -race`)
- `-count=10` stress: no flakiness
- 20/20 canonical live tests pass (including the regenerated `expected-spec.yaml`)
- 5-run idempotency on the new golden: byte-identical
- Coverage: `pkg/classify` 89.1%, `pkg/generate/rest` 93.2% (both up)

Could you re-run the review against `b9c0113`? And let me know your preference on TEST-004 (unified-string vs first-wins) so we can either confirm the current behavior or revert the unification.

santiago-praetorian (changes requested):
# Praetorian Capability PR Review — request changes

**Linear:** [LAB-2106](https://linear.app/praetorianlabs/issue/LAB-2106)  
**Head SHA:** `b9c0113f0c06ca802087a1333b2b71d0aa3a6681`  
**Reviewed at:** 2026-05-03T19:22:25.567206+00:00

Thanks for the thorough fixes on the prior round — all 5 prior blockers and the non-blocking findings were addressed in `2a3f15c` + `b9c0113` and verified against this head. Acceptance criteria for LAB-2106 are met. Live test suite passes 20/20.

Requesting changes on a small set of low/nit issues plus one stale doc surfaced by the live run. None are individually serious; bundled here so they ship together rather than as a separate cleanup PR.

|---|---|---|---|---|---|
| Count | 0 | 0 | 0 | 7 | 3 |

## Documentation (must update)

### DOC-001 — low

**File:** `test/README.md:218`

**Issue:** test/README.md line 218 documents the sample live-test run output as `rest-api PASS 8 8 79s`. This PR adds two new rest-api endpoints (/api/login urlencoded and /api/upload multipart) that bump the expected count to 10, confirmed by the live run: `rest-api PASS 10 10 22s`. The prior review's Phase 6 doc check covered CLAUDE.md, README.md, pkg/classify/doc.go, and pkg/generate/rest/doc.go — test/README.md was missed because it sits under the test harness directory and was not in the canonical PR file list (the PR did not touch it).

**Recommendation:** Update test/README.md line 218 from `rest-api PASS 8 8 79s` to `rest-api PASS 10 10 <duration>` to reflect the new endpoint count. Optionally re-time the sample with a fresh run.

```
  rest-api                    PASS      8           8          79s
```


**How to validate end-to-end:**

```bash
# 1. Build + start the test target services (rest-api, soap-service, graphql-server)
./test/setup-live-targets.sh

# 2. Run the full live suite (or scope to just rest-api)
./test/run-live-tests.sh
# or:
./test/run-live-tests.sh --targets rest-api

# 3. Tear down when done
./test/setup-live-targets.sh --teardown
```

The `Test Summary` table at the end of `run-live-tests.sh` is what `test/README.md:218` is meant to mirror. After this PR's `/api/login` and `/api/upload` additions, the actual run reports:

```
  rest-api                    PASS      10          10         22s
```

`test/README.md:218` still shows `rest-api PASS 8 8 79s` from before the form-body endpoints existed. Updating that line to `rest-api PASS 10 10 <duration>` aligns the docs with what the harness actually emits at HEAD `b9c0113`.

The ENDPOINTS/EXPECTED columns are sourced from `test/rest-api/expected-paths.json` (already updated in this PR — that's the determinism contract) and the live count from the crawl. The README sample is the only stale reference.

Prereq notes (from running this myself): if `setup-live-targets.sh` exits silently after the "Starting Services" header, you have stale orphan processes from prior runs occupying ports 8990–9010. Sweep with `pkill -x rest-api && pkill -x soap-service && lsof -nP -i :8990-9010 -t -c node | xargs -r kill` and retry. Tracked separately in [LAB-2893](https://linear.app/praetorianlabs/issue/LAB-2893/harden-testsetup-live-targetssh-orphan-pid-accumulation-and-silent).

---

## Quality (low / nit)

### QUAL-001 — low

**File:** `pkg/classify/classifier.go:130`

**Issue:** Within the `if len(req.Body) > 0` block, `getContentType(req.Headers)` is called twice: once at line 130 (to append the base media type to the key) and again at line 143 (to check for multipart and extract the boundary). The header map is iterated separately on each call. While not incorrect, this is unnecessarily redundant and wastes two linear header scans per body-bearing request.

**Recommendation:** Cache the result of `getContentType` in a local variable before the `if len(req.Body) > 0` block (or at least before line 130) and reuse it for both the base-type append and the multipart-boundary extraction. Example:
```go
ct := getContentType(req.Headers)
if len(req.Body) > 0 {
    if ct != "" {
        key += ":" + baseMediaType(ct)
    }
    fingerprintBody := req.Body
    if ct != "" {
        if mt, params, err := mime.ParseMediaType(ct); err == nil && mt == "multipart/form-data" {
            ...
        }
    }
    ...
}
```

```
			if ct := getContentType(req.Headers); ct != "" {
				key += ":" + baseMediaType(ct)
			}
			// Append a short fingerprint of the body ...
			fingerprintBody := req.Body
			if ct := getContentType(req.Headers); ct != "" {
```

---

### QUAL-002 — low

**File:** `pkg/generate/rest/openapi.go:246`

**Issue:** In `buildOperation`, the `ctGroups` map is iterated with `for mediaType, obs := range ctGroups`. Because Go map iteration order is non-deterministic, the `content` map is populated in a random order per run. For a single content type per endpoint this has no visible effect, but for an endpoint that produces multiple content-type entries (e.g. both `application/json` and `application/x-www-form-urlencoded`) the property-merge order within each group's schema is deterministic (slice-ordered), however if a future change makes the schema depend on insertion order or if `mergeObjectSchemas` is made non-commutative, this hidden non-determinism becomes a bug. The existing `TestExtractComponents_DeterministicMultiContentType` test covers output-level determinism via `extractComponents` sorting, but that sorting is downstream — `buildOperation`'s merge phase itself is non-deterministic in traversal order. Consistency with the rest of the deterministic-sort discipline in this file argues for sorting here too.

**Recommendation:** Collect content-type keys from `ctGroups` into a sorted slice and iterate in order, matching the pattern used elsewhere in `extractComponents` and `buildOperation`'s query-param and path-param sections:
```go
mediaTypes := make([]string, 0, len(ctGroups))
for mt := range ctGroups {
    mediaTypes = append(mediaTypes, mt)
}
sort.Strings(mediaTypes)
for _, mediaType := range mediaTypes {
    obs := ctGroups[mediaType]
    ...
}
```

```
			for mediaType, obs := range ctGroups {
```

---

### QUAL-003 — low

**File:** `pkg/generate/rest/openapi.go:286`

**Issue:** The response-body merge block (lines 286–304) still uses a hand-rolled property union loop that silently keeps the first type when two observations for the same status code have conflicting types for a property. This is inconsistent with the request-body merge path, which was updated by this PR to delegate to `mergeObjectSchemas` (which promotes conflicting types to `string`). The inconsistency is a correctness risk: a response property that is `integer` in one observation and `string` in another will silently keep the first-seen type rather than being promoted to `string`.

**Recommendation:** Replace the inline property-union loop in the existing-status-code branch with a call to `mergeObjectSchemas`, consistent with the request-body path:
```go
if len(ep.Response.Body) > 0 && existing.Value != nil && existing.Value.Content != nil {
    ct := strings.ToLower(ep.Response.ContentType)
    if ct == "" || strings.Contains(ct, "json") {
        newSchema := InferSchema(ep.Response.Body)
        if mt := existing.Value.Content["application/json"]; mt != nil {
            mt.Schema = mergeObjectSchemas(mt.Schema, newSchema)
        }
    }
}
```

```
			// Merge response body schema if both are objects
			if len(ep.Response.Body) > 0 && existing.Value != nil && existing.Value.Content != nil {
				// Only infer JSON schema for JSON-compatible content types
				ct := strings.ToLower(ep.Response.ContentType)
				if ct == "" || strings.Contains(ct, "json") {
					newSchema := InferSchema(ep.Response.Body)
					if newSchema != nil && newSchema.Value != nil && newSchema.Value.Properties != nil {
						if mt := existing.Value.Content["application/json"]; mt != nil && mt.Schema != nil &&
							mt.Schema.Value != nil && mt.Schema.Value.Properties != nil {
							for propName, propSchema := range newSchema.Value.Properties {
								if _, exists := mt.Schema.Value.Properties[propName]; !exists {
									mt.Schema.Value.Properties[propName] = propSchema
```

---

### QUAL-004 — nit

**File:** `pkg/generate/rest/openapi.go:521`

**Issue:** In `sanitizeResourceName`, the `for ext := range commonPathExtensions` loop breaks on the first matching extension found. Because Go map iteration is non-deterministic, if a segment's lower-case form happened to end with two listed extensions simultaneously (e.g., a contrived `.json.xml`), the extension stripped would vary across runs. In practice this cannot occur because no listed extension is a suffix of another listed extension, so the current behavior is correct. However, ranging over a map with a `break` makes the intent opaque — a reader might wonder about the non-determinism. The guard is benign today but fragile if extensions are extended. A sorted slice would be clearer and future-safe.

**Recommendation:** Replace the map-range-with-break with a check against a sorted slice of extensions or use `strings.Cut`/`strings.TrimSuffix` in a loop over a slice literal. This is a clarity/future-safety nit, not a correctness issue with the current extension set.

```
	for ext := range commonPathExtensions {
		if strings.HasSuffix(lower, ext) {
			seg = seg[:len(seg)-len(ext)]
			break
		}
```

---

### QUAL-005 — low

**File:** `pkg/generate/rest/openapi.go:649`

**Issue:** In `extractComponents`, `fingerprintToReqName` is a single map shared across ALL content-type variants of request bodies (the outer loop now iterates sorted `ctKeys` for all content types). If two different content-type schemas — e.g., a JSON body and a url-encoded body — happen to produce the same schema fingerprint (identical property names and types), the second one will reuse the component name derived from the first content-type's processing, without reflecting the content-type context in the name. This means `CreateLoginRequest` (derived from urlencoded) might end up with a `$ref` pointing to `CreateLoginRequest` even though it was created for JSON, or vice versa. This is a latent accuracy issue for endpoints that accept both JSON and form-encoded bodies with the same field shapes.

**Recommendation:** Consider scoping `fingerprintToReqName` per content-type key to prevent cross-CT schema reuse, or include the content-type in the fingerprint string so identical property shapes with different media types get distinct component names. Alternatively, document that cross-CT fingerprint reuse is intentional (if the author prefers a single schema component for identical shapes regardless of media type).

```
				sort.Strings(ctKeys)
				for _, ctKey := range ctKeys {
					mediaType := reqBody.Content[ctKey]
					if mediaType == nil || mediaType.Schema == nil {
						continue
					}
					if schema := mediaType.Schema.Value; schema != nil && schema.Properties != nil {
						fingerprint := schemaFingerprint(schema)
						if fingerprint != "" {
							var componentName string
							if existingName, exists := fingerprintToReqName[fingerprint]; exists {
```

---

## Tests (low / nit)

### TEST-001 — nit

**File:** `pkg/classify/classifier_test.go:471`

**Issue:** TestDeduplicate_KeepsDistinctBodiesByContentType passes a Body containing `\r\n` inside a Go raw string literal (backticks), so the bytes are literal backslash-r-backslash-n rather than actual CR/LF separators. The test still passes because Deduplicate only fingerprints the byte sequence and the assertion is on the result count (3 distinct CTs ⇒ 3 entries) — the bytes themselves are never parsed as multipart. However the body is misleading to future readers and is the only multipart body in this file that is not built with mime/multipart.NewWriter (compare to TestDeduplicate_MultipartBoundaryNormalized which uses real CRLF via interpreted string literals).

**Recommendation:** Switch to an interpreted string literal (double-quotes) so `\r\n` becomes real CRLF: `Body: []byte("--abc123\r\nContent-Disposition: form-data; name=\"key\"\r\n\r\nvalue\r\n--abc123--")`. Or build the body with multipart.NewWriter for consistency with neighbouring tests.

```
Body:    []byte(`--abc123\r\nContent-Disposition: form-data; name="key"\r\n\r\nvalue\r\n--abc123--`),
```

---

### TEST-002 — nit

**File:** `pkg/generate/rest/form_test.go:519`

**Issue:** The cap-enforcement test builds maxMultipartParts+1 parts and asserts `len(properties) <= maxMultipartParts`. Both sides reference the same constant, so raising maxMultipartParts to e.g. 100000 (silently widening the OOM blast radius this finding was created to prevent) leaves the test green: the loop creates 100001 parts, all fit under the new cap, the assertion holds. The test correctly catches removing the cap (len would exceed maxMultipartParts) but does not pin the policy value `1000` chosen for SEC-BE-002.

**Recommendation:** Add a single assertion that maxMultipartParts itself is `<= 1000` (or whatever the documented policy ceiling is): `assert.LessOrEqual(t, maxMultipartParts, 1000, "part-count cap policy ceiling for SEC-BE-002")`. This makes future widenings observable in code review.

```
for i := 0; i <= maxMultipartParts; i++ {
assert.LessOrEqual(t, len(schema.Value.Properties), maxMultipartParts,
"schema must contain at most maxMultipartParts properties; got %d", len(schema.Value.Properties))
```

---

### TEST-003 — low

**File:** `pkg/generate/rest/openapi_test.go:691`

**Issue:** The hyphen/underscore subtests express the expected value as `singularize("StoredXss")`, `singularize("CrossSiteScripting")`, `singularize("UserProfile")`. resourceNameFromPath internally calls sanitizeResourceName → singularize, so the assertion is `singularize(toCamelCase(seg)) == singularize("StoredXss")`. If singularize were modified to drop a trailing 's' for Xss/Scripting/Profile, both the production code and the expected value would change in lockstep and the test would still pass. The test still verifies the camelCase transformation correctly (the LHS depends on toCamelCase, the RHS does not), but it does not pin the singularization shape this PR's new path-cleanup pipeline produces.

**Recommendation:** Replace the singularize() expression in `expected` with the literal final string (e.g. `expected: "StoredXss"` if singularization should NOT strip the trailing 's', or whichever literal the current behavior produces). Compute it once with t.Logf during local verification, then hard-code the literal so future changes to singularize are observable.

```
{name: "hyphenated segment", path: "/stored-xss", expected: singularize("StoredXss")},
{name: "multi-hyphenated", path: "/cross-site-scripting", expected: singularize("CrossSiteScripting")},
{name: "underscore segment", path: "/user_profile", expected: singularize("UserProfile")},
```

---

### TEST-004 — low

**File:** `pkg/generate/rest/openapi_test.go:968`

**Issue:** TestBuildOperation_FormBody/'url-encoded form body' and TestOpenAPIGenerator_MultipartFormData_EndToEnd verify only that the expected content-type key (application/x-www-form-urlencoded or multipart/form-data) appears in requestBody.content (and the multipart test additionally substring-matches 'avatar' / 'binary' on the serialized YAML). Neither test descends into the SchemaRef.Value.Properties to assert that the actual field names from the body (`username`, `password`, `description`, etc.) are preserved with the expected types. A regression that produced an empty schema, dropped properties, mis-typed file fields as object, or swapped `format: binary` for `format: byte` would not flip these assertions. Property-level checks already exist in form_test.go for the parsers in isolation — the integration layer (Generate → buildOperation → mergeURLEncoded/mergeMultipart → schema) has weaker coverage.

**Recommendation:** After the content-type-key assertion, dive into content[mediaType].schema.properties and assert: (1) all expected property names are present, (2) text fields are type=string, (3) numeric fields are type=integer, (4) file fields have type=string AND format=binary. In the multipart test, replace the strings.Contains(specStr, ...) checks with structural lookups via the loaded openapi3 doc (the test already loads the spec; just walk the schema properties).

```
_, hasFormEncoded := content["application/x-www-form-urlencoded"]
assert.True(t, hasFormEncoded, "expected application/x-www-form-urlencoded in content, got keys: %v", content)
assert.Contains(t, specStr, "avatar", "spec missing 'avatar' field from multipart body")
assert.Contains(t, specStr, "binary", "spec missing format: binary for file upload field")
```

---

## Out of scope (filed separately)

Filed [LAB-2893](https://linear.app/praetorianlabs/issue/LAB-2893/harden-testsetup-live-targetssh-orphan-pid-accumulation-and-silent) for two unrelated bugs in `test/setup-live-targets.sh` (orphan-PID accumulation across runs + silent exit on port exhaustion) discovered while validating this PR. Not blocking this merge.

---

_Review generated by `capability-pr-review`. Rerun against the same PR head SHA + Linear ticket state to reproduce._

maneelmodi:
**Round 5 — internal pre-review pass (commit `7b21a5b`)**

Ran the in-house `/review-pr` (backend + security reviewers) on `b9c0113` before asking for human re-review. All real findings are now addressed:

| Reviewer | Severity | Finding | Status in `7b21a5b` |
|---|---|---|---|
| Security | Medium | `ParseURLEncodedForm` lacked body-size + field-count caps (JSON had 10MB, multipart had 1000 parts; importer accepts up to 500MB → DoS gap) | ✅ `maxURLEncodedBodySize = 10MB` + `maxFormFields = 1000`; tests added |
| Backend | High | `buildOperation` iterated `ctGroups` without sorting | ✅ Sorted before iterating |
| Security | Low | `< 4`-char boundary would corrupt body during normalization | ✅ `len(boundary) >= 4` guard; test added |
| Security | Nit | Collision-probability comment overstated (~7e-13 vs actual ~1e-14) | ✅ Corrected |
| Backend | Medium | Orphaned `// Compile-time interface compliance check.` comment | ✅ Removed |
| Backend | Medium | `TestExtractComponents`, `TestCapitalizeFirst`, `TestInferQueryParamType` mixed raw `t.*` and testify | ✅ Converted to testify |
| Backend | Low | `BenchmarkDeduplicate` mixed `b.N` style vs peers' `b.Loop()` | ✅ Standardized |
| (sync) | — | `pkg/crawl/browser.go` was stale vs `origin/main` (lacked `// #nosec G104` + nolint comment that had been added to main after we branched) | ✅ Synced from `origin/main` — not a LAB-2106 modification, just lets `make check` pass on this branch as on main |

**Verification:** 482+ tests pass; `-count=10 -race` no flakiness; 20/20 live tests pass; 5-run golden idempotency; `make check` clean for the first time on this branch.

**Deliberately deferred:** cross-package `getContentType`/`baseMediaType` duplication (needs new shared utility package); `buildOperation`/`extractComponents` complexity refactor (separate ticket); `mergeJSONBodies` first-wins-vs-unified (awaiting your reply on TEST-004).

@santiago-praetorian — would you mind re-reviewing against `7b21a5b`?

maneelmodi:
**Round 6 — `/capability-pr-review` internal pass (commit `caf4133`)**

Ran the in-house `/capability-pr-review` skill (capability-reviewer + backend-security + frontend-security + test-lead, normalized via cvss-scoring-expert) against `7b21a5b` before another human re-review. Found 12 non-critical findings; 11 fixed, 1 deferred with rationale.

| ID | Severity | Status |
|---|---|---|
| QUAL-001 | nit | ✅ Hoisted `getContentType` lookup in `Deduplicate` |
| **QUAL-002** | low | ⏸ **Deferred** — cross-package duplication; needs new shared utility package; out of LAB-2106 scope (already documented in PR description) |
| QUAL-003 | nit | ✅ Extracted local `baseMediaType` helper in pkg/generate/rest |
| SEC-BE-001 | low | ✅ Silent cap drops now emit `slog.Warn` with reason + cap |
| SEC-BE-002 | low | ✅ `ParseMultipartForm` gains 10MB body-size cap (parity with urlencoded+JSON) |
| SEC-FE-001 | nit | ✅ Inline test-fixture `<script>` documented as test-only |
| SEC-FE-002 | nit | ✅ `secret123` → `testpassword`; reference-capture.json updated |
| TEST-001 | nit | ✅ `TestDeduplicate_BodyWithoutContentType` |
| TEST-002 | low | ✅ `TestGetContentType` + `TestBaseMediaType` direct table tests |
| TEST-003 | low | ✅ Raw-string `\r\n` fixture converted to interpreted CRLF |
| TEST-004 | low | ✅ Empty-slice/all-empty-bodies subtests for merge helpers |
| TEST-005 | low | ✅ `TestMergeJSONBodies_SkipBranches` |

**Verification on `caf4133`**: 491 tests pass; `-count=10` no flakiness; 20/20 live tests; 5-run golden idempotency; coverage `pkg/classify` 89.9% / `pkg/generate/rest` 93.8% (both up); `make check` clean.

**On QUAL-002**: cross-package `getContentType`/`getHeader`/`baseMediaType` duplication exists across `pkg/classify`, `pkg/generate/rest`, and `pkg/generate/graphql/infer.go`. Fixing requires a new `pkg/httputil` (or similar) utility package — a non-trivial architectural decision affecting three callers. LAB-2106 is the form-data feature, not a cross-package refactor. Recommend a separate ticket.

PR is now in solid shape — all `/capability-pr-review` findings either resolved or deferred with rationale. Full final state in PR description and Linear comment.

santiago-praetorian:
## `/capability-pr-review` — APPROVE (head `caf4133`)

Re-ran `/capability-pr-review` against the current head plus a live-test pass and a real-traffic mitmproxy → import → generate run.

### `/capability-pr-review` summary

**Recommendation:** **APPROVE** — zero blockers, zero critical/high.

|---|---|---|---|---|---|
| Count | 0 | 0 | 0 | 7 | 1 |

Agents dispatched: `capability-reviewer` (QUAL), `backend-security` (SEC-BE), `core:testing:test-lead` (TEST), plus the ticket-alignment + Phase-6 docs/tests passes. Frontend-security did not fire (no `.ts/.tsx/.js/.jsx/.html/.css/.vue/.svelte` files in the PR). `cvss-scoring-expert` normalization was a no-op (SEC-BE returned zero findings).

#### All six LAB-2106 acceptance criteria met

| AC | Implementing code | Tests |
|---|---|---|
| AC1 urlencoded → typed params | `pkg/generate/rest/form.go::ParseURLEncodedForm` | `TestParseURLEncodedForm*` |
| AC2 multipart, file vs text | `pkg/generate/rest/form.go::ParseMultipartForm` | `TestParseMultipartForm*` |
| AC3 merge across observations | `mergeURLEncodedBodies`, `mergeMultipartBodies`, `mergeObjectSchemas`, `classify.Deduplicate` body-fingerprint key | `TestMergeURLEncodedBodies_*`, `TestMergeMultipartBodies/*`, `TestDeduplicate_*` |
| AC4 correct `requestBody` media types | `buildOperation` partitions by base media type | `expected-spec.yaml` golden |
| AC5 JSON unaffected | default arm in `buildOperation` still routes to `mergeJSONBodies` | `TestBuildOperation_FormBody/mixed`, `generate-rest` golden |
| AC6 tests cover all categories | URL-encoded + multipart + file uploads + mixed | full `pkg/generate/rest/form_test.go` + `openapi_test.go` |

#### Backend-security: zero findings

Walked the full SEC-BE rubric (injection, AuthN/Z, input validation, path traversal, deserialization, crypto, secrets, HTTP client, error disclosure, concurrency, deps). Every prior DoS concern (body-size caps, field-count caps, multipart-parts cap, boundary length guard, silent cap warnings) is resolved across rounds 4–6 — `addressed.json` filter confirmed.

#### Non-blocking findings (8 total — all `low`/`nit`)

| ID | Sev | File:Line | Summary |
|---|---|---|---|
| QUAL-001 | low | `pkg/generate/rest/form.go:104` | `ParseMultipartForm` cap-guard off-by-one — reads exactly 1001 parts before the `> maxMultipartParts` break fires; schema truncation is still correct, just the log message is one part ahead. |
| QUAL-002 | low | `pkg/generate/rest/openapi.go:241` | `buildOperation` defaults bodies with no Content-Type header to `application/json` silently; non-JSON bytes are dropped by `InferSchema` returning nil with no `slog.Warn`. |
| QUAL-003 | nit | `pkg/generate/rest/openapi.go:513` | `toCamelCase` digit handling not documented; doc comment doesn't explain that a leading digit triggers `sanitizeResourceName`'s `Resource` prefix. |
| QUAL-004 | low | `pkg/generate/rest/openapi.go:540` | `sanitizeResourceName` iterates `commonPathExtensions` (a Go map) — current 11 extensions are not mutual suffixes so output is deterministic today, but fragile if e.g. `.do`/`.doc` or `.js`/`.mjs` are ever added. Suggest ordered slice. |
| TEST-001 | low | `pkg/generate/rest/form_test.go:558` | `TestParseMultipartForm_PartCountCapped` uses `LessOrEqual` where input shape guarantees exact equality — matches CodeRabbit's open thread. |
| TEST-002 | low | `pkg/generate/rest/openapi_test.go:693` | `TestResourceNameFromPath_HandlesHyphens` uses `singularize()` in `expected` (tautological) — matches CodeRabbit's open thread. |
| TEST-003 | low | `pkg/generate/rest/openapi_test.go:1031` | `TestBuildOperation_FormBody/mixed` asserts content-type keys but not property shapes. |
| TEST-004 | low | `pkg/generate/rest/openapi_test.go:1082` | `TestOpenAPIGenerator_MultipartFormData_EndToEnd` relies on substring matches on the serialized spec; sibling `_E2E` test demonstrates the right `openapi3.NewLoader` approach. |

None gate this PR. A single follow-up ticket for the 4 QUAL + 4 TEST nits (plus the previously-deferred QUAL-002 cross-package `getContentType`/`getHeader`/`baseMediaType` extraction into a new `pkg/httputil`) would be the clean way to track them.

---

### Live test suite — 20/20 PASS

```
crawl-depth                 PASS      -           -          50s
generate-graphql            PASS      8           8          1s
generate-rest               PASS      10          10         0s
generate-wsdl               PASS      3           3          0s
graphql-server              PASS      8           8          1s
import-base64               PASS      2           2          1s
import-burp                 PASS      5           5          0s
rest-api                    PASS      10          10         20s
spec-edge                   PASS      -           -          1s

Total: 20 passed, 0 failed, 0 skipped
```

The `rest-api`, `generate-rest`, and `import-mitmproxy` targets all exercise the new form-body code path end-to-end.

---

### Manual end-to-end — real mitmproxy capture → import → generate

Drove real urlencoded + multipart traffic through `mitmdump` against the `test/rest-api` server, dumped to JSON, then ran `vespasian import mitmproxy` and `vespasian generate rest`. Resulting `forms-spec.yaml`:

```yaml
components:
    schemas:
        CreateLoginRequest:
            properties:
                password:    {type: string}
                remember_me: {type: boolean}
                username:    {type: string}
            type: object
        CreateUploadRequest:
            properties:
                description: {type: string}
                file:        {format: binary, type: string}
            type: object
paths:
    /api/login:
        post:
            requestBody:
                content:
                    application/x-www-form-urlencoded:
                        schema: {$ref: '#/components/schemas/CreateLoginRequest'}
    /api/upload:
        post:
            requestBody:
                content:
                    multipart/form-data:
                        schema: {$ref: '#/components/schemas/CreateUploadRequest'}
```

Confirmed working in this flow:

- Type inference firing on real bytes (`remember_me=true` → `boolean`, not `string`).
- File-vs-text discrimination in multipart (`file` → `type: string, format: binary`; `description` → `type: string`).
- Both schemas lifted to `#/components/schemas/$ref` with no inline duplication.
- Correct `requestBody.content` media-type keys for each endpoint (AC4).

---

### Conclusion

LAB-2106 acceptance criteria met, live suite green, manual real-traffic round-trip produces a correct OpenAPI 3.0 document. **Approving.**

Worktree retained at `vespasian/.worktrees/capability-pr-review/pr-101/` for the next reviewer to re-inspect.

santiago-praetorian (approved):
Approving per the detailed review comment above. Zero blockers, zero critical/high, all six LAB-2106 acceptance criteria met, live suite 20/20, manual mitmproxy round-trip confirms end-to-end form-body inference.

PR #121, opened 2026-05-09
feat(query-params): preserve multi-value query parameters (LAB-2110)

santiago-praetorian:
## Summary

Closes [LAB-2110](https://linear.app/praetorianlabs/issue/LAB-2110/support-multi-value-query-parameters-in-capture-and-openapi-generation).

Vespasian was discarding all but the first value when a query parameter appeared more than once (e.g. `?tag=a&tag=b`). This PR widens `ObservedRequest.QueryParams` from `map[string]string` to `map[string][]string` and propagates the change through capture, import, classification, and OpenAPI generation.

## What changed

**Producers** now assign `u.Query()` directly (since `url.Values` is already `map[string][]string`):

- `pkg/crawl/crawler.go` (`MapResult`)
- `pkg/crawl/network.go` (`mapNetworkToObservedRequest`)
- `pkg/crawl/forms.go` (POST + GET form discovery)
- `pkg/importer/helpers.go` (`extractQueryParams`)

**Classification** — `pkg/classify/classifier.go`:
- New exported `MergeUniqueOrdered(a, b []string) []string` helper.
- `Deduplicate` now merges multi-value params with **order-preserving union**: when two observations of the same endpoint share a key, all distinct values are kept in first-seen order, deduped.

**OpenAPI generation** — `pkg/generate/rest/openapi.go`:
- `buildOperation` detects multi-value via a `multiValueSeen` flag set if any observation of a parameter has `len(values) > 1`.
- Multi-value params emit:
  ```yaml
  schema:
    type: array
    items: { type: <inferred> }
  style: form
  explode: true
  ```
- Single-value params keep the existing scalar emission (no `style`/`explode`).
- New `inferQueryParamItemsType(values []string) string` mirrors `inferQueryParamType`'s precedence: integer → number → boolean → string.

**Test fixtures** — `test/fixtures/expected-har-duplicates-capture.json` previously documented the broken behavior (`sort=price` only); the fixture now correctly preserves both `sort=price` and `sort=name`.

## Breaking change

The on-disk shape of `capture.json`'s `query_params` field changes:

```diff
- "query_params": {"page": "1", "tag": "a"}
+ "query_params": {"page": ["1"], "tag": ["a","b"]}
```

Capture files generated by previous versions of Vespasian **must be regenerated**. README updated with an operator note. No custom `UnmarshalJSON` was added — `capture.json` is an intermediate artifact between capture and generation, not a long-lived asset.

## New tests

- `TestMapResult_MultiValueQueryParam`
- `TestMapNetworkToObservedRequest_MultiValueQueryParam`
- `TestBurpImporter_QueryParams_MultiValue`, `TestHARImporter_QueryParams_MultiValue`, `TestMitmproxyImporter_QueryParams_MultiValue`
- `TestDeduplicate_MergesMultiValueQueryParams`
- `TestMergeUniqueOrdered` (5 cases: nil/nil, a-nil, b-nil, overlap, b-with-self-dupes)
- `TestBuildOperation_ScalarQueryParam` (regression: scalar path unchanged)
- `TestBuildOperation_MultiValueQueryParam_AllInts`, `_Mixed`, `_AllBool`
- `TestInferQueryParamItemsType` (7 cases including empty, all-int, all-float, all-bool, mixed)

Existing single-value tests updated to `[]string{"v"}` form.

## Test plan

- [x] `go build ./...` exits 0
- [x] `go test -race ./...` all packages pass
- [x] `make lint` reports `0 issues.`
- [x] `make fmt` produces no diff
- [x] `make check` exits 0
- [x] Coverage: 77.1% combined (up from 76.9% on main; new functions are 100% covered)

## Acceptance criteria from LAB-2110

- [x] Multi-value query parameters are preserved through capture, import, and classification stages
- [x] OpenAPI generation produces `type: array` parameters with correct `items` schema
- [x] Single-value parameters continue to work as before (no regression — `TestBuildOperation_ScalarQueryParam`)
- [x] Deduplication correctly merges multi-value parameters across observations
- [x] Tests pass with single, multi, mixed, and array type-inference cases


santiago-praetorian:
## Review feedback addressed in 7826b22

Follow-up commit responding to the orchestrator-level PR review (`PR-REVIEW-COMMENT.md`) plus the Gemini and Claude bot reviews. Codex and CodeRabbit posted only rate-limit/credit-exhausted stubs — no review content to address from those.

### Resolved

| ID | Issue | Resolution |
| -- | ----- | ---------- |
| QUAL-001 / Gemini "MergeUniqueOrdered mutation" | `MergeUniqueOrdered` mutated caller's backing array | **D1 — copy-on-write rewrite.** Allocates fresh slice; dedups both inputs (also fixes the latent "duplicates in `a` survive" case Gemini noted). |
| QUAL-004 / SEC-BE-001 / TEST-005 / Gemini "Critical panic" / Claude "Minor note" | `info.values[0]` panic on `query_params: {"foo": []}` | **D2 — skip the param.** `if len(info.values) == 0 { continue }` in `buildOperation` emission loop. Regression test red-green verified. |
| QUAL-002 | GET form branch parsed query string twice | Cleanup: assign `obs.QueryParams = q` (already `url.Values`) instead of `u.Query()` re-parse. |
| QUAL-005 | Deleted absolute-URL validation comment | Restored rationale comment in `extractQueryParams`. |
| TEST-001 / Gemini "Deduplicate aliasing" | `TestDeduplicate` didn't assert non-mutation | Added input-not-mutated assertion. Combined with D1 (copy-on-write), the input observation slices are no longer aliased to merged output — the aliasing concern is moot. |
| TEST-002 | `TestMergeUniqueOrdered` didn't pin contract | New subtests asserting input slice is unchanged + duplicates in `a` are deduped. |
| TEST-003 | Missing capture round-trip + legacy-rejection regression | Two new tests in `pkg/crawl/capture_test.go`. The legacy-rejection test pins architecture §2 hard-break decision via `*json.UnmarshalTypeError`. |
| TEST-004 | Missing multi-value form tests | New `TestDiscoverForms_MultiValueQueryParams_POST` and `_GET` tests, parity with crawler/network/importer coverage. |
| DOC-001 | `CLAUDE.md` not updated for breaking change | Added one-line note in Capture Format section. |

### Declined (with rationale)

| ID | Issue | Decision |
| -- | ----- | -------- |
| QUAL-003 | `inferQueryParamItemsType` duplicates `inferQueryParamType` precedence | **Kept as-is.** 4-line precedence at 2 call sites — below the Rule of Three threshold. Consolidation would either require an awkward dual-purpose API or a reduce-style helper, neither of which improves readability. Rationale documented in `.capability-development/followup-architecture.md` D3. |

### Verification

```
make check
# fmt clean, vet clean, lint 0 issues, all tests pass with race detection
```

### Diff stats

```
9 files changed, 229 insertions(+), 9 deletions(-)
```

santiago-praetorian:
## Round 3 — `caa04de`

Addressed the second `/offsec:capability-pr-review` re-run findings + the 4 fresh CodeRabbit inline comments on `7826b22`.

### Resolved

| Source | Issue | Resolution |
| ------ | ----- | ---------- |
| QUAL-001 / CodeRabbit #2 (`classifier.go:173`) | `MergeUniqueOrdered` doc says "nil only when both inputs are nil" but guard returns nil for empty-but-non-nil too | Aligned doc to implementation. Broader contract (nil ≡ empty) is more useful for callers ranging over the result. |
| CodeRabbit #3 (`capture_test.go:78`) | `TestReadCapture_RejectsLegacyShape` used bare `json.Unmarshal`, not `ReadCapture` | Now calls `ReadCapture(strings.NewReader(legacy))`. `errors.As(err, &typeErr)` still matches `*json.UnmarshalTypeError`. |
| CodeRabbit #4 (`openapi.go:248`) | Scalar branch used only `info.values[0]` — order-dependent (`?limit=1` before `?limit=1.5` emitted `integer`) | Replaced with `inferQueryParamItemsType(info.values)`. Both branches now share the single source of truth for type inference (Rule of Three satisfied — late refactor of QUAL-003 from round 1, which had been declined as YAGNI). Added `TestBuildOperation_ScalarQueryParam_OrderIndependence` as regression test (red-green verified). |
| QUAL-002 (`forms.go:150`) | GET form branch assigns `url.Values` map directly without explicit safety annotation | Added `// q is a fresh url.Values from u.Query() above; safe to assign without copy.` comment. |

### Declined (with rationale)

| Source | Issue | Decline reason |
| ------ | ----- | ------------- |
| CodeRabbit #1 critical (`classifier.go:121-141`) | Claimed `req.QueryParams` fails typecheck because `QueryParams` is on `ObservedRequest` not `ClassifiedRequest` | **Bogus.** `ClassifiedRequest` embeds `crawl.ObservedRequest` (`pkg/classify/types.go:22-23`); Go field promotion makes `req.QueryParams` valid. Verified empirically: `make lint` exits with `0 issues.` and `go build ./...` succeeds. CodeRabbit's `golangci-lint v2.12.1` is running typecheck in a mode that doesn't follow embeddings — a CodeRabbit configuration issue, not a code defect. |

### Verification

```
make check
# fmt clean, vet clean, golangci-lint 0 issues., go test -race all pass
```

### Diff stats

```
5 files changed, 36 insertions(+), 7 deletions(-)
```

santiago-praetorian:
## Round 4 — `60f00b8`

Addressed the round-3 finding (QUAL-001, low): removed dead production code `inferQueryParamType` and its now-redundant test `TestInferQueryParamType`. The precedence cases were already covered by `TestInferQueryParamItemsType`.

`make check` exits 0. -46 lines net.

santiago-praetorian:
## Round 5 — `b5577a2`

Addressed the round-4 finding (TEST-001, low). Added 7 single-element-slice subtests to `TestInferQueryParamItemsType` covering the edge cases that were unique to the deleted `TestInferQueryParamType` (negative int, zero, negative float, scientific notation, empty string, alphanumeric, uppercase boolean-like).

`TestInferQueryParamItemsType` now has 14 subtests (was 7). `make check` exits 0.

santiago-praetorian:
## Round 6 — `a77b999`

Fixes a regression I discovered during manual end-to-end testing of the PR:

**Bug:** With the default `--deduplicate=true`, scalar query parameters that appeared with different values across observations (`?page=1` then `?page=2`) were over-widened to `type: array` in the generated OpenAPI spec. Tag/id were correctly arrays, but page should have been scalar.

**Root cause:** `classify.Deduplicate` merges `QueryParams` via `MergeUniqueOrdered` (union across observations), so the merged slice has `len > 1` even when each contributing request was scalar. `buildOperation`'s `multiValueSeen` flag was checking the merged length, which can't tell "was multi-value in a single request" from "differed across requests".

**Fix:** Track per-observation multi-value-ness on `ClassifiedRequest` BEFORE dedup can erase it.

- New field `MultiValueQueryKeys map[string]bool` on `ClassifiedRequest`.
- `RunClassifiers` populates the map from raw `QueryParams` where `len > 1`. Always non-nil so consumers can distinguish pipeline-produced from direct construction.
- `Deduplicate` unions the maps via a new `mergeMultiValueKeys` helper (also keeps cyclomatic complexity under the lint threshold).
- `buildOperation` prefers the map when non-nil; falls back to `len(vals) > 1` only when nil, preserving existing test behavior.

**Tests added (4):**
- `TestRunClassifiers_PopulatesMultiValueQueryKeys`
- `TestDeduplicate_UnionsMultiValueQueryKeys`
- `TestBuildOperation_PostDedupScalarNotOverWidened` — direct regression for the bug
- `TestBuildOperation_PostDedupArrayStillDetected` — companion: multi-value-in-one-obs still emits array

**Verification:**
- Manual end-to-end (the test that originally surfaced the bug): page now emits as scalar `integer`, tag/id remain arrays.
- `make lint` clean.
- All tests on touched packages pass.

santiago-praetorian:
## Round 8 — `c3597bf`

Addressed 5 of 6 round-7 review findings; declined 1 with rationale.

### Resolved (5)

| ID | File | Resolution |
| --- | --- | --- |
| QUAL-002 | `pkg/classify/classifier.go` | Doc-comment update on `mergeMultiValueKeys` explaining the false-entry-omission contract |
| QUAL-003 | `pkg/generate/rest/openapi.go` | Strengthened the cross-package-duplication comment on `baseMediaType` (deferral rationale + byte-identical invariant) |
| QUAL-004 | `pkg/generate/rest/openapi.go` | Refactored `inferQueryParamType` to delegate to `inferQueryParamItemsType([]string{value})` — precedence ladder now lives in one place |
| TEST-001 | `pkg/analyze/forms_test.go` | Tightened 8 slice assertions from `len == 0` to `len != 1` — now catches silent multi-value over-widening from a hypothetical form-merge regression |
| TEST-002 | `pkg/classify/classifier_test.go` | `TestDeduplicate_GETsStillCollapseByPath` now carries explicit `MultiValueQueryKeys: map[string]bool{}` on inputs and asserts non-nil + empty post-dedup, locking in the round-6 scalar-not-over-widened invariant |

### Declined (1)

| ID | Decline reason |
| --- | --- |
| QUAL-001 | The reviewer's own recommendation was "no change required" — the entryCopy deep-copy pattern is correct and adequately covered by the existing non-mutation assertion in `TestDeduplicate_MergesMultiValueQueryParams` |

### Verification

```
make check
# 0 issues. (lint), all packages pass with race
```

4 files changed, +33 / −23.

santiago-praetorian:
## Manual testing results — `c3597bf`

Ran four targeted end-to-end tests against the built binary. All pass.

### Test A — HAR import preserves multi-value query params

Input HAR with `?tag=a&tag=b&tag=c&page=1` and `?tag=b&tag=d&page=2&id=1&id=2`:

```bash
/tmp/vespasian import har /tmp/multi.har -o /tmp/multi-capture.json --no-banner
jq '.[].query_params' /tmp/multi-capture.json
```

Output:
```json
{"page": ["1"], "tag": ["a","b","c"]}
{"id": ["1","2"], "page": ["2"], "tag": ["b","d"]}
```

✅ Every value preserved end-to-end through the HAR importer.

### Test B — OpenAPI emits array vs scalar correctly (round-6 fix)

```bash
/tmp/vespasian generate rest /tmp/multi-capture.json -o /tmp/spec.yaml --no-banner
```

Emission per parameter:

| Param | Emitted | Why |
|---|---|---|
| `tag` | `array, items: string, style: form, explode: true` | Multi-value within each request |
| `id` | `array, items: integer, style: form, explode: true` | Multi-value in the request that had it |
| `page` | scalar `integer` (no `style`/`explode`) | Scalar in each request; values differ across requests |

✅ The round-6 fix works: `page` is correctly emitted as scalar even though `Deduplicate` unioned `["1","2"]`. The `MultiValueQueryKeys` tracking preserved per-observation truth past the dedup merge.

### Test D — Empty-values guard (round-1 panic fix)

Hand-crafted capture with `"query_params": {"foo": [], "bar": ["1"]}`:

```bash
/tmp/vespasian generate rest /tmp/empty.json -o /tmp/empty-spec.yaml --confidence=0 --no-banner
grep -E "name:|type:" /tmp/empty-spec.yaml
```

Output:
```
                  name: bar
                    type: integer
```

✅ Exit 0 (no panic), `bar` emitted as scalar `integer`, `foo` correctly omitted. Pre-LAB-2110 this scenario panicked with `index out of range` on `info.values[0]`.

### Test E — Order-independent scalar type inference (round-3 fix)

Two scalar observations of `?limit=1` then `?limit=1.5`:

```bash
/tmp/vespasian generate rest /tmp/order.json -o /tmp/order-spec.yaml --confidence=0 --no-banner
grep -A3 "name: limit" /tmp/order-spec.yaml
```

Output:
```
                  name: limit
                  required: true
                  schema:
                    type: number
```

✅ `limit` emitted as scalar `number` (not `integer`). The scalar branch now walks all observed values via `inferQueryParamItemsType` instead of taking `values[0]`. Pre-round-3 this would have incorrectly emitted `integer` because the first observed value was `"1"`.

### Summary

| Test | Result | Verifies |
|---|---|---|
| A — HAR import | ✅ | Producer-side multi-value preservation |
| B — OpenAPI emission | ✅ | Array vs scalar; round-6 over-widening fix |
| D — Empty-values guard | ✅ | Round-1 panic fix |
| E — Order-independent scalar | ✅ | Round-3 type-inference fix |

Combined with the unit tests + `make check` + the live end-to-end suite, the PR is verified working as designed at every layer.

santiago-praetorian:
## Live end-to-end test suite — `c3597bf`

Ran `test/run-live-tests.sh` against the built binary. All 21 targets pass.

```
  Test Summary
  --------------------------  --------  ----------  ---------  --------
  crawl-depth                 PASS      -           -          51s
  generate-graphql            PASS      8           8          0s
  generate-rest               PASS      10          10         0s
  generate-wsdl               PASS      3           3          1s
  graphql-server              PASS      8           8          0s
  import-base64               PASS      2           2          0s
  import-burp                 PASS      5           5          1s
  import-mitmproxy-native     PASS      3           3          1s
  rest-api                    PASS      10          10         21s
  spec-edge                   PASS      -           -          0s

  Total: 21 passed, 0 failed, 0 skipped
```

Notable for LAB-2110:

- **`import-duplicates` PASS** — exercises the HAR fixture (`test/fixtures/sample-har-duplicates.json` → `expected-har-duplicates-capture.json`) that this PR updated to reflect the new multi-value `query_params` shape (`"sort": ["price","name"]` instead of the previously-broken `"sort": "price"`).
- **`import-unicode` PASS** — exercises the Burp unicode fixture this PR also updated to `["世界"]` slice form.
- **`import-har`, `import-burp`, `import-mitmproxy`, `import-mitmproxy-native` PASS** — all three importer paths still correctly produce captures under the new `map[string][]string` type.
- **`generate-rest` PASS** — OpenAPI generator end-to-end against the live REST test server matches `expected-spec.yaml` byte-for-byte.
- **`rest-api` PASS (21s)** — full crawl → classify → generate cycle against the live Go REST server produces the expected spec.

Combined with `make check` (unit + race) and the 4 targeted manual tests in the previous comment, the PR is verified at all three layers: unit, integration, and live end-to-end.

santiago-praetorian:
## Clarification on `generate-rest` and `rest-api` endpoint counts

Both targets show **10 endpoints (was 8)**. The +2 comes from the merge with `main`, not directly from LAB-2110:

```
+/api/login    POST   (LAB-2106 / LAB-2109)
+/api/upload   POST   (LAB-2106 / LAB-2109)
```

These were added in `main` commit `fb1a8bf` (`feat(rest): parse form-encoded and multipart request bodies (LAB-2106)`) when the merged-in feature taught the crawler to follow JS-driven `fetch()` calls to form-POST endpoints (`/api/login`, `/api/upload`).

For LAB-2110 specifically, both targets pass against the new expected count, confirming that the type widening to `map[string][]string` does not regress the crawl/classify/generate pipeline on the live REST test server's full set of endpoints — including the two new form-POST endpoints that exercise the LAB-2106 form-data parsing path.

gitoso (approved):
I reviewed the PR and tested it with Probus. Everything worked well 😉 

I am approving the changes as is, but Claude's review flagged some security concerns that I think are worth noting. Is a "malicious traffic capture file" part of our Threat Model? 

If it is, then I think the two issues should be fixed before merging, as a maliciously-crafted capture file could cause resource consumption if they are eventually parsed by Vespasian. Other than that, the PR looks good to me.

Here are the security concerns raised by the AI-review:

---

### SEC-BE-001 — low — SEC-BE

**File:** `pkg/classify/classifier.go:192-201`

**Issue:** When two or more observations on the same METHOD:path:body-fingerprint share a query-parameter key, Deduplicate calls MergeUniqueOrdered(existing.req.QueryParams[k], vs) for each. MergeUniqueOrdered allocates a fresh seen-map sized to len(a)+len(b) and a fresh out slice of the same capacity on every call (classifier.go:240-256), so merging K observations each with V values yields O(K*V) allocations of ever-growing slices, i.e. O((K*V)^2) total work and peak heap. The crawl path is bounded by DefaultMaxPages=1000, but the import path (Burp/HAR/mitmproxy → extractQueryParams) is bounded only by the 100 MB capture file cap. A malicious capture with many requests collapsing to the same dedup key and each carrying a long ?k=... list can therefore stall an offline `vespasian generate` invocation for an unbounded amount of CPU/memory relative to file size. The same amplification then happens a second time inside buildOperation (openapi.go:235) when it folds info.values across grouped endpoints.

**Recommendation:** Cap the per-key merged-values length inside Deduplicate (e.g., stop merging once existing.req.QueryParams[k] reaches a reasonable threshold like 1024 distinct values) and short-circuit MergeUniqueOrdered when its inputs are already at the cap. Alternatively, change MergeUniqueOrdered to mutate a caller-owned set/seen map so the seen-map isn't reallocated on every call. The fix in extractQueryParams (SEC-BE finding above) would also remove the upstream amplification, but defending Deduplicate independently is worthwhile because ClassifiedRequest values can also be constructed directly in tests / by other future callers.

```
			if req.QueryParams != nil {
				if existing.req.QueryParams == nil {
					existing.req.QueryParams = make(map[string][]string)
				}
				for k, vs := range req.QueryParams {
					existing.req.QueryParams[k] = MergeUniqueOrdered(existing.req.QueryParams[k], vs)
				}
			}
```

---

### SEC-BE-002 — low — SEC-BE

**File:** `pkg/importer/helpers.go:54-67`

**Issue:** extractQueryParams parses every URL string with url.Parse(...).Query() and returns the resulting map[string][]string directly, without any cap on the number of distinct keys or values per key. The capture file as a whole is read with a 100 MB io.LimitReader (pkg/crawl/capture.go:23), but that limit is over the entire JSON document, not over a single URL field. An adversarial Burp/HAR/mitmproxy import can therefore contain a single request whose URL string is on the order of tens of megabytes and consists of `?k=v1&k=v2&k=v3...` repeated, which url.Values.Query() expands into a slice with millions of entries inside a single map cell. That slice then propagates through RunClassifiers, gets deep-copied in Deduplicate (classifier.go:182-188), and accumulated again by MergeUniqueOrdered in buildOperation (openapi.go:235). MergeUniqueOrdered builds a seen-map sized to len(a)+len(b) on every call (classifier.go:240-241), so the merge chain is O(N^2) in CPU and allocates O(N) auxiliary memory per call — running on a 10M-entry slice will burn gigabytes of heap and seconds-to-minutes of CPU before the operator can interrupt. Risk is bounded by the 100 MB capture cap (so this is not unbounded) but a single 100 MB malicious file can easily exhaust the operator's RAM during an offline `vespasian import` or `vespasian generate` run.

**Recommendation:** Add a per-key value-count cap inside extractQueryParams (e.g., cap each map value at 256 distinct values; drop the excess with a debug log), or cap len(parsed.Query()) total entries. Mirroring the spirit of analyze/forms.go's maxFieldsPerForm=500 / maxFormsPerBody=1000 caps is the obvious shape. Same cap should be applied symmetrically in pkg/crawl/network.go:177 and pkg/crawl/crawler.go:356 (MapResult) since both also assign u.Query() directly without bounds.

```
func extractQueryParams(urlStr string) map[string][]string {
	parsed, err := url.Parse(urlStr)
	queryValues := parsed.Query()
	if len(queryValues) == 0 {
		return nil
	}
	return queryValues
```

gitoso (approved):
LGTM

PR #127, opened 2026-05-27
feat(grpc): protocol foundation — classifier, reflection probe, wire-format parser (LAB-2779)

saullocarvalho:
## Summary

Implements [LAB-2779](https://linear.app/praetorianlabs/issue/LAB-2779) — the gRPC protocol foundation (1/5 of the gRPC support epic, LAB-2778). This is the protocol plumbing that later gRPC work (generator, CLI wiring, traffic inference) builds on.

- **Classifier** (`pkg/classify/grpc.go`): detects `application/grpc` and `application/grpc-web*` content-types, `grpc-status`/`grpc-message` trailer headers, and the `/<pkg.Service>/<Method>` POST path shape. Content-type + trailer together get an HTTP/2-style confidence boost.
- **Reflection probe** (`pkg/probe/grpc.go`): enumerates services, methods, and message types via the gRPC Server Reflection Protocol using `grpcreflect.NewClientAuto` (v1 with v1alpha fallback), walking the transitive `FileDescriptorProto` closure. SSRF preflight via the existing URL validator plus a dial-time TOCTOU re-check. When a server is reachable but reflection is unavailable it returns a structured `GRPCReflectionResult{ReflectionEnabled: false}` with the gRPC status reason (`Unimplemented`/`Unauthenticated`/`PermissionDenied`) rather than a nil result.
- **Wire-format parser** (`internal/grpcwire/`): length-prefix framing + protobuf wire-format primitives (`ParseFrame`, `ParseVarint`, `ParseTag`, `WalkFields`) for later traffic-based inference.
- Adds `google.golang.org/grpc`, `google.golang.org/protobuf`, and `github.com/jhump/protoreflect`.

Out of scope here (separate sub-issues): `.proto` generator (LAB-2781), CLI wiring (LAB-2782), test target (LAB-2780), docs (LAB-2783).

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/grpcwire/... ./pkg/classify/... ./pkg/probe/...` (unit tests for parser, classifier, and probe incl. reflection-unavailable path)
- [x] `go vet ./...` clean
- [x] `go mod tidy` no drift
- [x] Verified end-to-end against [gRPCGoat](https://github.com/rootxjs/grpc-goat): reflection-enabled lab enumerated correctly; reflection-disabled labs returned the structured `(gRPC Unimplemented)` finding; TLS-only labs over cleartext fell back cleanly to a bare error.

Refs LAB-2779.

PR #159, opened 2026-06-16
feat(grpc): gRPC protocol support — foundation, .proto generator, test target, CLI wiring (LAB-2778)

saullocarvalho:
## Summary

**Primary tracking ticket: [LAB-2778](https://linear.app/praetorianlabs/issue/LAB-2778) — Add gRPC protocol support (epic).**

gRPC becomes Vespasian's fourth target protocol alongside REST, GraphQL, and SOAP. This branch consolidates the full epic: foundation + test target + `.proto` generator + CLI wiring + docs. The items below map to LAB-2778's sub-issues.

> **Note:** Supersedes #127, which was closed when its head branch (`saullobranco/lab-2779-grpc-foundation`) was renamed to `saullobranco/lab-2778-grpc-support` to reflect the broader scope. The original review thread lives on #127.

### What's included (by sub-issue)

- **[LAB-2779](https://linear.app/praetorianlabs/issue/LAB-2779) — Foundation:**
  - Classifier (`pkg/classify/grpc.go`): detects `application/grpc`/`application/grpc-web*` content-types, `grpc-status`/`grpc-message` trailer headers, and the `/<pkg.Service>/<Method>` POST path shape; content-type + trailer get an HTTP/2-style confidence boost.
  - Reflection probe (`pkg/probe/grpc.go`): enumerates services/methods/message types via Server Reflection (`grpcreflect.NewClientAuto`, v1 → v1alpha fallback), walking the transitive `FileDescriptorProto` closure. SSRF preflight + dial-time TOCTOU re-check; structured `reflection-unavailable` finding (`Unimplemented`/`Unauthenticated`/`PermissionDenied`) instead of nil.
  - Wire-format parser (`internal/grpcwire/`): length-prefix framing + protobuf primitives (`ParseFrame`, `ParseVarint`, `ParseTag`, `WalkFields`) with varint-overflow and field-number bounds hardening.
- **[LAB-2780](https://linear.app/praetorianlabs/issue/LAB-2780) — Test target** (`test/grpc-server/`): Go gRPC server with reflection enabled, sample services, live-test harness integration.
- **[LAB-2781](https://linear.app/praetorianlabs/issue/LAB-2781) — `.proto` generator** (`pkg/generate/grpc/`): emits `.proto` from reflection results.
- **[LAB-2782](https://linear.app/praetorianlabs/issue/LAB-2782) — CLI wiring** (`cmd/vespasian/main.go`, `pkg/generate/registry.go`).
- **[LAB-2783](https://linear.app/praetorianlabs/issue/LAB-2783) — Documentation** (`README.md`, `CLAUDE.md`, `docs/grpc.md`): gRPC added to the protocol/feature tables, "API Type Support" classification + reflection sections, and CLI reference. Adds a standalone [`docs/grpc.md`](docs/grpc.md) protocol guide (classification signals, reflection probe behavior, `.proto` generation, TLS, end-to-end example) linked from the README, plus the [gRPC API Discovery](https://github.com/praetorian-inc/vespasian/wiki/gRPC-API-Discovery) wiki tutorial.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...` clean
- [x] `go test ./...` — all 15 packages pass (incl. `pkg/generate/grpc`)
- [x] `go mod tidy` no drift

Closes LAB-2783.

Refs LAB-2778 (LAB-2779, LAB-2780, LAB-2781, LAB-2782).


santiago-praetorian (changes requested):
## Praetorian Capability PR Review — REQUEST CHANGES

Automated multi-agent review of this PR against the LAB-2778 gRPC epic (LAB-2779/2780/2781/2782/2783). Head SHA `790dbb9`.

**Recommendation: REQUEST CHANGES** — 6 blocking, 7 non-blocking.

> **Two CLI-surface items were reconciled as Linear ticket amendments, not code changes** (see comments on LAB-2778/2781/2782): (1) the standalone `vespasian probe grpc` verb was correctly dropped — no protocol exposes a standalone `probe` verb; probing runs inside `scan`/`generate`. (2) the `generate grpc` verb name is correct — the `generate` verb is keyed by api-type (`rest`/`graphql`/`wsdl`/`grpc`), not output format. Neither requires a code change.

### 🔴 Blocking (6)

- **QUAL-003** (medium) `pkg/probe/grpc.go:182` — probeTarget function exceeds 60-line rubric threshold
  - Extract the TLS credential setup + connection phase and the descriptor-walk phase into unexported helpers. probeTarget becomes an orchestrating function of ~30 lines. Add a brief inline comment at each call site explaining why the sub-function can return early.
- **REQ-003** (medium) `cmd/vespasian/main.go:908` — gRPC content-type auto-detection not wired into scan
  - Either add the GRPCClassifier to detectAPIType so application/grpc[-web] captures auto-route to the gRPC pipeline, or amend LAB-2782 to record that gRPC is intentionally opt-in (binary HTTP/2 framing is rarely present in browser/proxy captures). The PR documents the opt-in decision but it contradicts the written criterion.
- **REQ-004** (medium) `pkg/generate/grpc/generator.go:46` — Traffic-derived .proto fallback (scalars/enums) not implemented
  - Either implement the traffic-inference fallback (consuming internal/grpcwire to reconstruct message fields from observed frames, mapping common scalar/enum wire types) with fixture-traffic unit tests, or amend LAB-2781 to defer traffic-only inference to a follow-up and mark the reflection path as the sole supported route.
- **TEST-001** (medium) `pkg/generate/grpc/generator.go:60` — Add test for cross-endpoint FileDescriptor aggregation
  - Add a test that builds two endpoints with disjoint FileDescriptors (e.g. users.proto on endpoint A, orders.proto on endpoint B), calls Generate once, and asserts both 'service UserService' and the second service render in the output. This makes the aggregation behavior a tested contract rather than implementation detail.
- **TEST-002** (medium) `pkg/generate/grpc/generator.go:65` — Add test for conflicting-descriptor error branch in Generate
  - Add a test that supplies two endpoints with the same filename key but differing bytes, assert Generate returns an error containing 'conflicting file descriptors'; and a companion test where two endpoints share an identical filename+bytes pair and assert no error (dedup path).
- **TEST-003** (medium) `pkg/probe/grpc.go:71` — Add test for host:port+TLS dedup keying in GRPCProbe.Probe
  - Add a Probe test with multiple endpoints sharing one host:port (different paths) against one in-process server, asserting all results receive the same GRPCSchema and reflection ran once; optionally assert two endpoints differing only by scheme (http vs https) are treated as distinct targets.

### 🟡 Non-blocking (7)

- **QUAL-001** (nit) `pkg/generate/grpc/generator.go:144` — else-after-early-return in buildDescriptorGraph is non-idiomatic Go
- **QUAL-002** (low) `pkg/probe/grpc.go:136` — Default port 80 for grpc:// scheme is non-standard and silently misleading
- **QUAL-004** (low) `pkg/probe/grpc.go:198` — InsecureSkipVerify hardcoded with no Config override path
- **REQ-001** (low) (not implemented) — Add ./docs/grpc.md protocol documentation
- **REQ-002** (low) (not implemented) — Publish and link a gRPC wiki tutorial from README Tutorials section
- **REQ-005** (low) `test/grpc-server/README.md:11` — Stale `vespasian probe grpc reflection` references in test docs
- **SEC-BE-001** (low) `pkg/probe/grpc.go:198` — gRPC reflection probe disables TLS verification (InsecureSkipVerify)

---
_Generated by `capability-pr-review`. Full roll-up with evidence and diffs available on request._

saullocarvalho:
## Review remediation — pushed in `a4103d5`

Thanks for the review. Dispositions below (verified at the new head; `gofmt`/`go build`/`go vet`/`golangci-lint`/`go test -race ./...` all green; coverage probe 87.8%, generate-grpc 91.8%).

### Blocking
| ID | Disposition |
|----|-------------|
| **QUAL-003** | ✅ Fixed — `probeTarget` split into `dialGRPC` + `runReflection`; now a ~23-line orchestrator (was 86). Return semantics preserved. |
| **REQ-003** (auto-detect) | 📝 **Ticket amendment** — gRPC stays opt-in (binary HTTP/2 framing rarely survives browser/proxy captures). Recorded on LAB-2782; no code change, per the review's own preface. |
| **REQ-004** (traffic inference) | 📝 **Deferred** to a follow-up on LAB-2781; reflection is the sole supported route (wire format strips field names). No code change. |
| **TEST-001** | ✅ Added `TestGenerator_Generate_AggregatesAcrossEndpoints` (disjoint descriptors on two endpoints → both render). |
| **TEST-002** | ✅ Added conflicting-descriptor error test + identical-bytes dedup test. |
| **TEST-003** | ✅ Added `TestGRPCProbe_Probe_DedupsByTarget` (one server/many paths → single reflection result fanned out; pointer identity asserted). |

### Non-blocking
| ID | Disposition |
|----|-------------|
| **QUAL-001** | ✅ Removed else-after-return in `buildDescriptorGraph`. |
| **QUAL-002** | ✅ `grpcTarget` now logs a debug line when assuming a default port. |
| **QUAL-004 / SEC-BE-001 / CodeQL #81** | ✅ TLS verification is now **on by default**; skip is opt-in via `Config.GRPCInsecureSkipVerify` + new `--grpc-insecure-skip-verify` flag (scan/generate). README updated. |
| **REQ-005** | ✅ The removed `probe grpc reflection` verb is gone from the live-test harness (`test_grpc_server` now builds a capture and runs `generate grpc`), `test/grpc-server/README.md`, and `doc.go`. |
| **REQ-001 / REQ-002** (docs/grpc.md, wiki) | 📝 Deferred (LAB-2783) — gRPC is documented in README + `doc.go`, consistent with how REST/GraphQL/WSDL are documented (the `docs/*.md` siblings never existed). |

### CodeRabbit inline
- ✅ `test/grpc-server/main.go` package doc moved to `doc.go`.
- ⏩ **Already resolved** on the current head (the inline review was against an earlier commit): port help already says `50051`; `test/README.md` already says `23`; the live-test python validator already scopes RPC matching per-service via `service_body()` and checks `streaming_methods`; `Generate()` already aggregates descriptors across endpoints with conflict detection.
- 📝 **lab.proto Buf `PACKAGE_DIRECTORY_MATCH`** — deferred: there is no buf tooling in the repo (advisory only), and `lab.v1` is baked into the generated `*.pb.go`, the runtime service paths, and `expected-paths.json`, so renaming the dir/package would break the live-test contract for a test fixture. Happy to do it if you'd prefer.

Note: the one remaining `golangci-lint` finding (`pkg/crawl/http_crawler.go:238`, gosec G704) is pre-existing and byte-identical to `main` (a newer gosec rule than CI's pinned version), not introduced here.

santiago-praetorian (changes requested):

**PR:** https://github.com/praetorian-inc/vespasian/pull/159
**Linear:** https://linear.app/praetorianlabs/issue/LAB-2778
**Head SHA:** 40cf8cd76a0935892f9f7eaba15dd61e182187fd
**Reviewed at:** 2026-06-30T19:05:00Z

> Round 2. All 9 round-1 "fixed" items were re-verified at the current head and confirmed resolved; the 5 intentional deferrals (LAB-2781/2782/2783) are not re-raised as blockers. 2 new blocking findings remain, both minor and quickly closed (one is a one-line test, one a one-line client option).

|---|---|---|---|---|---|
| Count | 0 | 0 | 1 | 6 | 1 |

---

<details>

<details>
<summary><code>SEC-BE-001</code> · medium · SEC-BE · <code>pkg/probe/grpc.go:231</code></summary>

**Issue:** `dialGRPC` builds the `grpc.ClientConn` without `grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(...))`. `walkFileDescriptors` caps the file count at `maxGRPCFileDescriptors=1000`, but each reflection response message is bounded only by gRPC's 4 MB default — so a malicious/compromised target the operator points the scanner at can return ~1000 maximally-sized `FileDescriptorProto`s (~4 GB aggregate), all retained in `result.FileDescriptors` and re-unmarshalled by the generator. This is a memory-amplification DoS that can OOM-kill the scanning host. The 10s reflection timeout bounds time, not peak memory. CVSS `AV:N/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:H` = 6.5.

**Recommendation:** Set an explicit per-message recv cap on the reflection client (`grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(maxReflectionMsgBytes))` in `dialGRPC`) and/or add an aggregate byte budget alongside the existing count cap in `walkFileDescriptors`, returning early once total marshalled bytes exceed the budget.

```go
return grpc.NewClient(t.hostPort,
	grpc.WithTransportCredentials(creds),
	grpc.WithContextDialer(dialer),
)
```

</details>

<details>
<summary><code>TEST-001</code> · low · TEST · <code>pkg/probe/grpc.go:306</code></summary>

**Issue:** `extractService` populates the new exported `GRPCMethod.ServerStreaming`/`ClientStreaming` fields from `m.IsServerStreaming()`/`m.IsClientStreaming()`. The probe unit tests only exercise the unary health-check RPC, so only the `false` branch is asserted (`grpc_test.go:190-191`). No `go test` asserts the `true` branch — the streaming path is covered solely by the live-test harness (`lab.proto` `ListUsers` stream), which does not run under `go test`. A regression inverting or hardcoding the streaming markers would pass the entire unit suite. AC-79-3 calls for unit tests for the probe.

**Recommendation:** Add an in-process probe test that registers a service with a server-streaming (ideally also client-streaming) RPC and asserts the discovered `GRPCMethod` has `ServerStreaming==true`, pinning the true branch alongside the existing unary coverage. Alternatively, file a follow-up ticket to track it — that would reclassify this finding as non-blocking.

```go
assert.False(t, checkMethod.ClientStreaming)
assert.False(t, checkMethod.ServerStreaming)
```

</details>

</details>

<details>

<details>
<summary>Ticket alignment (4)</summary>

<details>
<summary><code>REQ-001</code> · low · <code>README.md</code></summary>

**Acceptance criterion:** AC-83-4 (LAB-2783) — published wiki tutorial linked from the README Tutorials section.
**Status:** DEFERRED → LAB-2783 (intentional; round-1 REQ-002).
**Evidence:** README contains gRPC reference docs but no gRPC tutorial-wiki link.

**Recommendation:** Keep deferred under LAB-2783; publish and link the gRPC wiki tutorial when that ticket is worked.

</details>

<details>
<summary><code>REQ-002</code> · low · <code>cmd/vespasian/main.go:914</code></summary>

**Acceptance criterion:** AC-82-4 (LAB-2782) — Content-Type `application/grpc[-web]` treated as gRPC by default.
**Status:** DEFERRED → LAB-2782 (intentional; round-1 REQ-003). `detectAPIType` runs only REST/WSDL/GraphQL; gRPC requires explicit `--api-type grpc`.
**Evidence:** Binary HTTP/2 framing rarely survives browser/proxy captures, so gRPC stays opt-in. Amended on LAB-2782; README/CLAUDE.md document the opt-in behavior.

**Recommendation:** Keep deferred under LAB-2782. No code change required this round.

</details>

<details>
<summary><code>REQ-003</code> · low · <code>docs/grpc.md</code></summary>

**Acceptance criterion:** AC-83-1 (LAB-2783) — `./docs/grpc.md` exists.
**Status:** DEFERRED → LAB-2783 (intentional; round-1 REQ-001). gRPC is documented in README.md + package `doc.go` files.
**Evidence / correction:** The round-1 deferral note claimed "docs/*.md siblings never existed", but `docs/crawler.md` **does** exist — so that rationale is inaccurate even though the deferral itself stands.

**Recommendation:** Keep deferred under LAB-2783, but correct the note: `docs/crawler.md` already exists, so a `docs/grpc.md` sibling is consistent with repo convention and should be authored when LAB-2783 is worked.

</details>

<details>
<summary><code>REQ-004</code> · low · <code>pkg/generate/grpc/generator.go:75</code></summary>

**Acceptance criterion:** AC-81-3 (LAB-2781) — traffic-derived `.proto` fallback handling scalars+enums when reflection is unavailable.
**Status:** DEFERRED → LAB-2781 (intentional; round-1 REQ-004). Reflection path is type-correct; traffic-only fallback is not implemented (`Generate` returns an error when no `FileDescriptors` are present). `internal/grpcwire` exists as foundation but is not wired in.
**Evidence:** The wire format strips field names, so traffic-only inference is low-fidelity.

**Recommendation:** Keep deferred under LAB-2781. The reflection-derived half of AC-81-3 is met; only the traffic fallback remains.

</details>

</details>

<details>

<details>
<summary><code>QUAL-001</code> · nit · <code>pkg/classify/grpc.go:85</code></summary>

**Issue:** `url.Parse(req.URL)` is called inside an `err == nil` guard, so a parse failure silently suppresses the gRPC path-shape signal with no comment. Fail-open is correct (a malformed URL just skips the heuristic), but the QUAL rubric requires a comment when an error is intentionally ignored.

**Recommendation:** Add a one-line comment above the `if` noting the url.Parse failure intentionally skips the path-shape signal (no confidence penalty).

```go
if parsed, err := url.Parse(req.URL); err == nil && grpcPathRE.MatchString(parsed.Path) {
```

</details>

<details>
<summary><code>QUAL-002</code> · low · <code>pkg/probe/grpc.go:295</code></summary>

**Issue:** `proto.Marshal(fd.AsFileDescriptorProto())`'s error is discarded in the implicit false branch. The file is already marked `fetched` (line 293) before the marshal, so a serialization failure permanently omits the descriptor from the output map with no log or comment. Downstream `buildDescriptorGraph` then emits a `// WARNING` header or fails to resolve imports, but the root cause is invisible.

**Recommendation:** Log the error (e.g. `slog.Debug` with file + error) and add a comment stating the omission is intentional and non-fatal.

```go
if raw, err := proto.Marshal(fd.AsFileDescriptorProto()); err == nil {
	out[name] = raw
}
```

</details>

</details>

</details>

_20 findings dismissed as already addressed (9 round-1 fixes re-verified at head, 5 intentional deferrals tracked under LAB-2781/2782/2783, 6 CodeRabbit inline items) — see `.review/addressed.json`._

---

saullocarvalho:
## Round-2 review addressed — pushed in `20bba74`

Both blockers fixed, both quality items done; verified `gofmt`/`go build`/`go vet`/`golangci-lint`/`gosec` (0 HIGH)/`go test -race ./...` green, coverage probe 87.6% / generate-grpc 91.8%.

| ID | Disposition |
|----|-------------|
| **SEC-BE-001** (blocking) | ✅ Fixed — `dialGRPC` now sets `grpc.MaxCallRecvMsgSize` (explicit 4 MiB per-message cap), and `walkFileDescriptors` enforces a **64 MiB aggregate** serialized-descriptor budget alongside the existing 1000-file count cap. Bounds peak memory to ~64 MiB, closing the reflection memory-amplification OOM vector. |
| **TEST-001** (blocking) | ✅ Fixed — added `TestGRPCProbe_Probe_DiscoversServerStreaming`: registers `labpb` `UserService` (server-streaming `ListUsers`) on an in-process reflection server and asserts the probe reports `ServerStreaming==true` (and `ClientStreaming==false`), pinning the true branch under `go test`. |
| **QUAL-001** (nit) | ✅ Commented the intentional fail-open `url.Parse` path-shape skip in the classifier. |
| **QUAL-002** (low) | ✅ `walkFileDescriptors` now `slog.Debug`-logs a descriptor marshal failure instead of silently omitting it. |
| **REQ-001 / REQ-002 / REQ-004** | 📝 Remain deferred under LAB-2783 / LAB-2782 / LAB-2781 as agreed. |
| **REQ-003** | 📝 Deferred under LAB-2783 — and thanks for the catch: you're right that `docs/crawler.md` **does** exist, so my round-1 "siblings never existed" note was inaccurate. A `docs/grpc.md` sibling is consistent with convention and will be authored when LAB-2783 is worked. |

Thanks again for the thorough pass.

santiago-praetorian (changes requested):

**PR:** https://github.com/praetorian-inc/vespasian/pull/159
**Linear:** https://linear.app/praetorianlabs/issue/LAB-2779
**Head SHA:** 95068d5dc132fcb870ce3b10fb1338014d07040c
**Reviewed at:** 2026-07-07T22:22:21Z

|---|---|---|---|---|---|
| Count | 0 | 0 | 3 | 8 | 1 |
| Blockers | 0 | 0 | 3 | 4 | 0 |

---

<details>
<summary><strong>Blockers (7)</strong></summary>

<details>
<summary><code>TEST-001</code> · low · TEST · <code>cmd/vespasian/main.go:746</code></summary>

**Issue:** This PR adds `case pipeline.APITypeGRPC: return "gRPC"` to apiTypeDisplayName (cmd/vespasian/main.go:754). TestAPITypeDisplayName (cmd/vespasian/main_test.go:2294) tables APITypeREST/APITypeWSDL/APITypeGraphQL/unknown but omits the grpc case, so the new arm's display string is never verified. Low severity (verbose-output cosmetic), but it is trivially closable new public behavior.

**Recommendation:** Add `{pipeline.APITypeGRPC, "gRPC"}` to the TestAPITypeDisplayName table.

```
	case pipeline.APITypeGRPC:
```

</details>

<details>
<summary><code>TEST-002</code> · low · TEST · <code>internal/grpcwire/frame.go:155</code></summary>

**Issue:** WalkFields parses hostile wire input and returns errors on malformed data. Its truncation branches for length-delimited, 64-bit, and 32-bit fields are tested (TestWalkFields_TruncatedLengthDelim/Fixed64/Fixed32) and the unsupported-wire-type branch is tested, but two error-return branches have no coverage: (1) a malformed field tag — ParseTag error at line 159 returns immediately; and (2) a truncated varint VALUE inside the WireVarint case — ParseVarint error at line 166. Neither is reachable via existing tests (TruncatedLengthDelim exercises the payload length-check, not a varint parse error). For a security-facing wire parser, error propagation on malformed input is behavior worth pinning.

**Recommendation:** Add two small cases: WalkFields on a truncated tag varint (e.g. []byte{0x80}) and on a varint field whose value is truncated (tag(1, WireVarint) followed by a lone 0x80 continuation byte), asserting each returns an error.

```
		tag, n, err := ParseTag(b)
		if err != nil {
			return err
		case WireVarint:
			_, m, err := ParseVarint(b)
			if err != nil {
				return err
```

</details>

<details>
<summary><code>TEST-003</code> · low · TEST · <code>internal/pipeline/apitype.go:42</code></summary>

**Issue:** A documented, security-relevant design decision of this PR is that gRPC is intentionally opt-in: DetectAPIType (auto mode) instantiates only the WSDL/REST/GraphQL classifiers and must never return APITypeGRPC (CLAUDE.md: 'detectAPIType (auto mode) runs only the REST/WSDL/GraphQL classifiers — gRPC is never auto-selected'; ScanCmd help: 'grpc requires explicit selection'). TestDetectAPIType_* cases (apitype_test.go:107,126,183) cover GraphQL/REST/WSDL preference but none feed gRPC-shaped traffic (application/grpc content-type + /pkg.Service/Method path) and assert the result is NOT grpc. The invariant is currently structural (the classifier simply isn't in the list), so a future change adding GRPCClassifier into DetectAPIType would silently flip gRPC traffic to auto-selected with no failing test.

**Recommendation:** Add a TestDetectAPIType case supplying gRPC-shaped requests (application/grpc content-type, /pkg.Service/Method path, POST) and assert DetectAPIType returns APITypeREST (the documented default), never APITypeGRPC — pinning the opt-in invariant.

```
func DetectAPIType(requests []crawl.ObservedRequest, threshold float64) string {
	wsdlClassifier := &classify.WSDLClassifier{}
	restClassifier := &classify.RESTClassifier{}
	graphqlClassifier := &classify.GraphQLClassifier{}
```

</details>

<details>
<summary><code>TEST-004</code> · medium · TEST · <code>pkg/generate/registry.go:51</code></summary>

**Issue:** This PR adds the `case "grpc"` arm to generate.Get, the public dispatch mapping an API type to its SpecGenerator (returns &grpcgen.Generator{}). registry_test.go's TestGet table (registry_test.go:44-46) covers only rest/graphql/empty; the new grpc arm — the entry point a caller hits with `generate grpc` — is never exercised, so a refactor that drops or mis-wires it (wrong generator, or accidentally routing grpc to the unsupported-type error) would not be caught. (wsdl is also untested in this table, but that is pre-existing, not new in this PR.)

**Recommendation:** Add a case to TestGet: `{"grpc", "grpc", ...}` asserting Get("grpc") returns a non-nil generator with no error (and, if the table asserts type, that it is *grpcgen.Generator).

```
	case "grpc":
		return &grpcgen.Generator{}, nil
```

</details>

<details>
<summary><code>TEST-005</code> · medium · TEST · <code>pkg/probe/grpc.go:100</code></summary>

**Issue:** GRPCProbe.Probe bounds reflection fan-out with `if len(seen) >= p.config.MaxEndpoints { break }`, keyed on distinct grpcTargetInfo. This is an established package convention: TestWSDLProbe_MaxEndpointsRespected (wsdl_test.go:170), TestOptionsProbe_MaxEndpoints (options_test.go:250), and TestGraphQLProbe_MaxEndpointsRespected (graphql_test.go:248) each assert their probe honors a MaxEndpoints=2 limit. grpc_test.go has no equivalent — grep for MaxEndpoints in pkg/probe/grpc_test.go returns nothing. The cap branch (break) is never exercised, so a regression that dropped the limit (unbounded reflection dialing of every distinct host:port) would pass CI.

**Recommendation:** Add TestGRPCProbe_Probe_MaxEndpointsRespected mirroring the WSDL/Options/GraphQL tests: supply more distinct-target gRPC endpoints than Config.MaxEndpoints (e.g. 3 targets, MaxEndpoints=2), count dials via a wrapping Dialer, and assert at most MaxEndpoints reflection dials occur and endpoints beyond the cap receive no GRPCSchema.

```
		if len(seen) >= p.config.MaxEndpoints {
			break
		}
```

</details>

<details>
<summary><code>TEST-006</code> · low · TEST · <code>pkg/probe/grpc.go:214</code></summary>

**Issue:** The GRPCProbe doc comment explicitly promises 'Uses grpcreflect.NewClientAuto, which tries v1 first then falls back to v1alpha on Unimplemented', and the rubric calls out v1->v1alpha fallback as a probe edge case. Every test server (pkg/probe/grpc_test.go and internal/pipeline/pipeline_test.go) registers reflection via reflection.Register(s), which serves the v1 reflection API, so all tests take the v1 path; grep shows v1alpha appears in grpc_test.go only inside a negative service-name filter assertion, never as a server that speaks solely v1alpha. The fallback branch is therefore uncovered. The fallback logic itself lives in the third-party grpcreflect library (not vespasian code), which lowers severity, but the doc makes an explicit behavioral claim about a legacy-server compatibility path that no test protects.

**Recommendation:** Add a test that stands up a reflection server exposing only grpc.reflection.v1alpha.ServerReflection and assert the probe still returns ReflectionEnabled=true with discovered services, proving the fallback. If v1alpha-only cannot be registered with the current grpc-go version, soften the doc comment so it does not imply tested behavior.

```
// Uses grpcreflect.NewClientAuto, which tries v1 first then falls back to
// v1alpha on Unimplemented.
	client := grpcreflect.NewClientAuto(reqCtx, conn)
```

</details>

<details>
<summary><code>TEST-007</code> · medium · TEST · <code>pkg/probe/grpc.go:298</code></summary>

**Issue:** walkFileDescriptors enforces the memory-amplification DoS mitigation for the hostile-server case (AC-79-4 area) via two bounds — maxGRPCFileDescriptors (1000-file count cap) and maxGRPCDescriptorBytes (64 MiB aggregate-byte cap) — plus a per-name dedup guard (fetched map) and transitive-dependency recursion. addressed.json records these caps were added in commit 20bba74 as the fix for the 'reflection memory-amplification DoS' finding, but grep of pkg/probe/*_test.go shows NO test references walkFileDescriptors, maxGRPCFileDescriptors, or maxGRPCDescriptorBytes. The only coverage is incidental (the health-service probe walks a tiny descriptor graph). If a refactor removed or inverted either cap check, no test would fail — the security fix ships without a guarding regression test. The count-cap and byte-cap early-return branches are never exercised, and the fetched dedup branch (return on already-seen name) is only hit incidentally.

**Recommendation:** Add a unit test driving walkFileDescriptors with a synthesized *desc.FileDescriptor graph that exceeds each bound: one case where file count would exceed maxGRPCFileDescriptors and one where aggregate bytes would exceed maxGRPCDescriptorBytes, asserting len(out)/totalBytes stop growing at the cap; plus a shared/cyclic dependency case proving the fetched map prevents re-serialization. Assert on the resulting map size and byte total, not merely absence of panic.

```
func walkFileDescriptors(fd *desc.FileDescriptor, fetched map[string]bool, out map[string][]byte, totalBytes *int) {
	if fd == nil || len(fetched) >= maxGRPCFileDescriptors || *totalBytes >= maxGRPCDescriptorBytes {
		return
	}
```

</details>

</details>

<details>

<details>
<summary>Ticket alignment (2)</summary>

<details>
<summary><code>REQ-001</code> · low · <code>docs/grpc.md</code></summary>

**Status:** DEFERRED → LAB-2783

**Issue:** AC-79-DOC-DOCS asks for a protocol-specific guide (docs/grpc.md) when applicable. docs/ contains only crawler.md (plus benchmarks/ and images/); no grpc.md was added. gRPC is documented in README.md and package doc.go files. Authoring docs/grpc.md is explicitly deferred to follow-up ticket LAB-2783 per addressed.json, so this is not a blocker.

**Recommendation:** Author docs/grpc.md under LAB-2783 mirroring the existing docs/crawler.md convention (classification heuristics, reflection probe behavior, .proto generation, TLS / --grpc-insecure-skip-verify).

</details>

<details>
<summary><code>REQ-002</code> · low · <code>wiki</code></summary>

**Status:** DEFERRED → LAB-2783

**Issue:** AC-79-DOC-WIKI asks for a wiki follow-up edit if a tutorial page references affected behavior. No gRPC wiki tutorial was published. This is explicitly deferred to follow-up ticket LAB-2783 per addressed.json, so this is not a blocker.

**Recommendation:** Publish/update the gRPC wiki tutorial under LAB-2783.

</details>

</details>

<details>

<details>
<summary><code>QUAL-001</code> · nit · <code>pkg/classify/grpc.go:97</code></summary>

**Issue:** The bare `switch { ... }` that assigns `confidence` covers exactly four cases (hasContentType&&hasTrailer, hasContentType, hasTrailer, hasPath) and has no `default`. This is currently exhaustive only because of the earlier `if len(signals) == 0 { return false, 0, "" }` guard, which guarantees at least one of hasContentType/hasTrailer/hasPath is true whenever this switch runs. That invariant is not visible at the switch site itself -- a future contributor adding a fifth signal (e.g. a new header check) that sets a new bool but doesn't add a matching case would silently fall through with `confidence` left at its zero value, so `ClassifyDetail` would return `(true, 0.0, reason)`: classified as gRPC but with zero confidence, filtered out by any positive threshold with no error or log to explain why.

**Recommendation:** Add a `default:` branch (e.g. `default: confidence = GRPCPathConfidence` or a panic/log for an unreachable state) so the exhaustiveness is enforced at the switch itself rather than relying on the caller's earlier guard, or compute confidence as `max()` over all activated signal confidences so a missed case can't silently zero it out.

```
	var confidence float64
	switch {
	case hasContentType && hasTrailer:
		confidence = GRPCContentTypeTrailerConfidence
	case hasContentType:
		confidence = GRPCContentTypeConfidence
	case hasTrailer:
		confidence = GRPCTrailerConfidence
	case hasPath:
		confidence = GRPCPathConfidence
	}
```

</details>

<details>
<summary><code>QUAL-002</code> · low · <code>pkg/generate/grpc/generator.go:36</code></summary>

**Issue:** maxGRPCFileDescriptors (1000) and maxGRPCDescriptorBytes (64<<20) are defined independently in both pkg/probe/grpc.go (lines 40 and 50) and pkg/generate/grpc/generator.go (lines 36-37). The generator.go comment explicitly acknowledges the mirroring: 'These mirror the probe-path caps (pkg/probe/grpc.go) so the offline generate entry point shares the same descriptor-count and aggregate-byte bounds instead of trusting capture-file provenance (SEC-BE-001).' Both packages already import pkg/classify, so there is no import-cycle obstacle to defining the bounds once. As two independent literals, nothing enforces they stay equal if one is retuned later without updating the other, silently reintroducing the provenance-trust gap the duplication was added to close.

**Recommendation:** Hoist maxGRPCFileDescriptors and maxGRPCDescriptorBytes into a single shared location both packages already depend on (e.g. pkg/classify, alongside GRPCReflectionResult), and reference that constant from both pkg/probe/grpc.go and pkg/generate/grpc/generator.go, or add a test asserting the two pairs of constants stay equal.

```
	maxGRPCFileDescriptors = 1000
	maxGRPCDescriptorBytes = 64 << 20 // 64 MiB
```

</details>

<details>
<summary><code>QUAL-003</code> · low · <code>pkg/generate/grpc/generator.go:68</code></summary>

**Issue:** The FileDescriptors aggregation loop in Generator.Generate nests outer-for (endpoints) -> inner-for (FileDescriptors) -> if existing,ok (conflict check) -> if !bytes.Equal (mismatch check) -- 4 levels of nesting, exceeding the rubric's depth-3 guideline. The logic is simple (dedupe-by-filename with a conflict check) but the nesting buries the common-case assignment (merged[name] = raw) at the bottom of two nested ifs.

**Recommendation:** Flatten with a guard clause: `existing, ok := merged[name]; if !ok { merged[name] = raw; continue }; if !bytes.Equal(existing, raw) { return nil, fmt.Errorf(...) }` -- this drops one nesting level and puts the common case first. Alternatively extract the inner loop body into a small helper such as `mergeDescriptor(merged map[string][]byte, name string, raw []byte) error`.

```
	for _, ep := range endpoints {
		if ep.GRPCSchema == nil || !ep.GRPCSchema.ReflectionEnabled {
			continue
		}
		for name, raw := range ep.GRPCSchema.FileDescriptors {
			if existing, ok := merged[name]; ok {
				if !bytes.Equal(existing, raw) {
					return nil, fmt.Errorf("conflicting file descriptors for %q across gRPC endpoints", name)
				}
				continue
			}
			merged[name] = raw
		}
	}
```

</details>

</details>

</details>

_19 findings dismissed as already addressed (resolved in prior rounds or deferred to LAB-2781/2782/2783) — see `.review/addressed.json`._

---

santiago-praetorian:

> **Superseded — see the [APPROVED review](https://github.com/praetorian-inc/vespasian/pull/159#pullrequestreview) below.** This comment is the detailed evidence log for that review (head `2346252`). Verdict: **APPROVE**; the two items originally flagged as formula-blockers are now tracked as non-blocking follow-ups (**LAB-4693**, **LAB-4694**).

Full static/dynamic/live verification passed: build/vet/fmt/lint clean · `go test -race ./...` 18/18 · live harness 26/26 (incl. `grpc-server`) · `generate grpc` → proto3 → `protoc` compile ✅.

<details><summary>Findings detail (all non-blocking; evidence retained for reference)</summary>

### Recommended follow-ups (tracked, non-blocking)

**SEC-BE-001 · medium · `pkg/probe/grpc.go:87` → LAB-4693**
*CVSS:* `CVSS:3.1/AV:N/AC:H/PR:N/UI:R/S:U/C:N/I:N/A:H`

The per-target reflection memory-amplification DoS is fixed (MaxCallRecvMsgSize 4 MiB + 64 MiB aggregate descriptor budget + 1000-file cap per target, verified present in current code). However, GRPCProbe.Probe retains one *GRPCReflectionResult per distinct grpcTargetInfo in schemasByTarget, and the fan-out is bounded only by Config.MaxEndpoints (DefaultMaxEndpoints = 500 in pkg/probe/types.go:74). Each target's retained descriptors can reach the 64 MiB aggregate cap (classify.MaxGRPCDescriptorBytes), so a capture referencing many distinct, reachable, hostile gRPC host:port targets can drive aggregate retained memory to ~MaxEndpoints * 64 MiB (~32 GiB), OOM-ing the scanning host. This is a distinct dimension from the addressed per-target fix: there is no global/aggregate cap across targets, and the addressed note's claim that memory is 'bounded to ~64 MiB' is only per-target-accurate. Attack requires an unusual capture (imported Burp/HAR with hundreds of hostile gRPC targets, or attacker-controlled DNS for many names), so likelihood is low and the impact is confined to the operator's own scanning process.

→ Add a global budget across targets in GRPCProbe.Probe: track cumulative retained descriptor bytes across all entries in schemasByTarget and stop probing further targets once a total cap (e.g. a small multiple of MaxGRPCDescriptorBytes, exposed as a Config field) is reached, mirroring the existing per-target loop-break guard in runReflection.

**TEST-001 · low · `cmd/vespasian/main.go:539` → LAB-4694**

The new user-facing --grpc-insecure-skip-verify flag is declared on both GenerateCmd (main.go:471) and ScanCmd (main.go:562) and copied into pipeline.Options at main.go:539 (GenerateCmd.Run) and main.go:666 (ScanCmd.Run). No test in cmd/vespasian/main_test.go references GRPCInsecureSkipVerify at all (grep returns zero hits), so the CLI-struct -> pipeline.Options wiring for this flag is uncovered. The underlying semantic behavior IS well tested one layer down (internal/pipeline/pipeline_test.go TestClassifyProbeGenerate_GRPCInsecureSkipVerify proves the Options field toggles TLS trust-chain verification through the probe, with both true and verify-by-default halves), and TestSlugOptions_Embedded / TestDangerousAllowPrivate_GenerateCmd establish the precedent that CLI-boundary flag wiring gets its own test. The residual risk is limited to a silent regression where the command struct field stops reaching Options (e.g., a dropped field during a refactor), which the pipeline-level test would not catch. Severity is low because the effective behavior is covered and only the one-line assignment is unverified.

→ Add a small test (mirroring TestSlugOptions_Embedded) asserting that GenerateCmd.GRPCInsecureSkipVerify and ScanCmd.GRPCInsecureSkipVerify are exposed via the kong struct, or extend an existing GenerateCmd.Run-driven test to assert the flag reaches the pipeline (e.g., a gRPC capture whose reflection only succeeds with skip-verify=true), so a dropped field assignment at main.go:539/666 is caught at the CLI boundary.

### Deferred / quality (non-blocking)

- **REQ-001** (low, Ticket alignment) `README.md:279` — AC-83-4: wiki tutorial not linked from a README 'Tutorials' section _(→ LAB-2783)_
  AC-83-4 requires a published wiki tutorial linked from README's Tutorials section. The wiki page (gRPC API Discovery) IS referenced from the README (line 279) and from docs/grpc.md (line 187), but README has no dedicated 'Tutorials' section (the link sits inside the 'gRPC Server Reflection' subsection). Wiki publication is external and cannot be verified in a static repo review. Deferred to LAB-2783 per prior review rounds. Non-blocking.

- **REQ-002** (low, Ticket alignment) `internal/pipeline/apitype.go:42` — AC-82-4: gRPC content-type auto-detection not wired into detectAPIType _(→ LAB-2782)_
  AC-82-4 (original) called for auto-detection of application/grpc[-web]. DetectAPIType (apitype.go:42-70) instantiates only the WSDL/REST/GraphQL classifiers and never the GRPCClassifier, so gRPC is never auto-selected and requires explicit --api-type grpc. This is intentional and amended on LAB-2782: gRPC binary HTTP/2 framing rarely survives browser/proxy captures, so it is opt-in by design (documented in README line 45, docs/grpc.md, CLAUDE.md, and pinned by TestDetectAPIType_NeverAutoSelectsGRPC). Not a genuine gap; recorded here only for AC traceability. Non-blocking (wontfix-by-design).

- **REQ-003** (low, Ticket alignment) `pkg/generate/grpc/generator.go:86` — AC-81-3: traffic-derived .proto fallback (scalars/enums) not implemented _(→ LAB-2781)_
  AC-81-3 requires reflection-derived output to be type-correct AND a traffic-derived fallback that handles common scalar types and enums when reflection is unavailable. The reflection route is fully implemented and type-correct (renderProto reconstructs the FileDescriptorProto graph via protoprint). The traffic-derived fallback is NOT implemented: Generate() returns 'gRPC spec generation requires server reflection; no FileDescriptors available' (generator.go:86-88) when no reflection descriptors are present. internal/grpcwire exists as foundation for this path but is explicitly not wired into the generator (see internal/grpcwire/doc.go and CLAUDE.md). This is an intentional deferral — reflection is the sole supported route because the wire format strips field names. Non-blocking under the follow-up ticket.

- **QUAL-001** (low, Quality) `README.md:288` — README CLI Reference for scan/generate omits --grpc-insecure-skip-verify flag
  The new `--grpc-insecure-skip-verify` flag (added to both `ScanCmd` and `GenerateCmd` in cmd/vespasian/main.go) is documented in CLAUDE.md's CLI Commands table and in docs/grpc.md, but the README's per-command '### `vespasian scan`' and '### `vespasian generate`' CLI Reference code blocks — which otherwise itemize every flag including --dangerous-allow-private, --confidence, --probe, --deduplicate — do not list it. A reader using README.md alone as the CLI reference for scan/generate would not learn this flag exists.

- **QUAL-002** (medium, Quality) `internal/grpcwire/frame.go:1` — internal/grpcwire package ships as dead code with zero production callers _(→ LAB-2781)_
  internal/grpcwire (frame.go, 207 lines, plus doc.go) adds a full length-prefix + protobuf wire-format parser (ParseFrame, ParseVarint, ParseTag, WalkFields), but nothing in the classifier, probe, or generator packages imports it — a repo-wide grep for `grpcwire` outside internal/grpcwire/ and its own test file returns no matches. The package's own doc.go states it is 'foundation for a future traffic-based RPC inference path and is not yet wired into the classifier, probe, or generator.' This is a substantial chunk of exported API (Frame, ParseFrame, ParseFrames, WireType, Tag, ParseVarint, ParseTag, WalkFields) merged with no current caller, which is dead code from the perspective of this PR and adds public surface area that must be maintained before it is exercised by any real code path.

</details>

---
_Prior rounds' 30 findings verified resolved/deferred (LAB-2781/2782/2783) and not re-raised. Generated by `capability-pr-review` + manual verification._

santiago-praetorian (approved):
## Praetorian Capability PR Review — APPROVE

Re-reviewed at head `2346252` against the LAB-2778 gRPC epic, with a full static/dynamic/live verification pass. **Approving** — the two items my prior comment flagged are now tracked as follow-ups, so there are no open blockers.

### Why APPROVE
The earlier `capability-pr-review` run surfaced two formula-blockers (SEC-BE-001, TEST-001). Both were marginal — not runtime defects — and are now captured as follow-up tickets, which clears them under the review's own blocker rule:

- **SEC-BE-001** (medium) → **[LAB-4693](https://linear.app/praetorianlabs/issue/LAB-4693)** — add a cross-target aggregate memory budget in `GRPCProbe.Probe` (per-target cap already in place; needs a global cap for the hundreds-of-hostile-targets case).
- **TEST-001** (low) → **[LAB-4694](https://linear.app/praetorianlabs/issue/LAB-4694)** — add a CLI-boundary test for the `--grpc-insecure-skip-verify` flag wiring (behavior already covered one layer down in `pipeline_test.go`).

Please address both as follow-ups; neither blocks merge.

### Verification (this review, at PR head)
- **Static:** `go build`, `go vet`, `gofmt -s -l`, `golangci-lint run` — all clean (0 lint issues).
- **Dynamic:** `go test -race ./...` — 18/18 packages pass, no races. gRPC coverage 88–94% (`cmd/vespasian` 67%, the CLI-wiring gap now tracked in LAB-4694).
- **Live:** full harness 26/26 pass incl. `grpc-server`; `generate grpc` ran live reflection → valid proto3 (3 services, 4 rpcs, `ListUsers` streaming) matching `expected-paths.json`, and the emitted `.proto` compiles with `protoc`.

### Remaining deferrals (unchanged, non-blocking)
- REQ-003 → LAB-2781 (traffic-derived `.proto` fallback), QUAL-002 → LAB-2781 (`internal/grpcwire` wired in there)
- REQ-002 → LAB-2782 (auto-detect intentionally opt-in, amended)
- REQ-001 → LAB-2783 (wiki tutorial), QUAL-001 (README CLI-reference flag listing — trivial, optional)

Nice work landing the full epic. 🚀

PR #160, opened 2026-06-18
feat(grpc): reflection-off enumeration — gRPC-Web bindings + grpc-gateway (LAB-3864 Part B)

saullocarvalho:
## Summary

Implements **LAB-3864 Part B** — gRPC discovery techniques for when **Server Reflection is disabled** (the common production case). Part A + Phase 1 (reflection probe, `.proto` generator, hardening) shipped in #159, which is now **merged to `main`**; this PR is based on `main` (no longer stacked) and also picks up #165 (`--proxy`).

The reflection-off techniques are delivered **only through the `scan` / `generate --api-type grpc` pipeline** — there is no standalone `probe` verb. The ticket originally named `probe grpc bindings`/`probe grpc gateway` subcommands; that was reversed in favor of pipeline integration (consistent with #159 dropping the `probe` verb), and LAB-3864's B1/B2 acceptance criteria have been amended to match.

### What's included

- **B1 — gRPC-Web JS bindings scrape** (`pkg/analyze/grpc_web_bindings.go`)
  Recovers service/method/type names + streaming flags from `*_connect.js` / `*_pb_service.js` / `*_grpc_web_pb.js` bundles in a capture, via jsluice (AST-based, not filename-based). Applied by `enrichGRPCFromBindings` in the pipeline. Because it reads JS bodies from the capture (no network), it runs even with `--probe=false`.
- **B2 — grpc-gateway / Envoy OpenAPI scrape** (`pkg/probe/grpcgateway.go`, a `ProbeStrategy`)
  Fetches a bounded set of well-known swagger paths (`/swagger.json`, `/openapiv2/*.swagger.json`, …), SSRF-gated, recognizes grpc-gateway docs by their `operationId`/`tags` shape (including the `fqn` naming strategy, e.g. `greet.v1.Greeter_SayHello`), and records service names. Descriptor synthesis is deferred to the generator.
- **Generator synthesis** (`pkg/generate/grpc/synthesize.go`)
  `FileDescriptorsFromServices` synthesizes `FileDescriptorProto` wire bytes from recovered names (empty message stubs, streaming preserved) so B1/B2 **reuse the existing protoprint render path** — no second emitter. Recovered identifiers are validated against the proto grammar before synthesis (malformed names dropped, not injected). Descriptor count/byte caps are enforced **before** any parse.
- **Orchestration** (`internal/pipeline/` — `apitype.go`, `pipeline.go`, `grpc_enrich.go`)
  `StrategiesForType("grpc")` chains reflection + gateway probes; `ClassifyProbeGenerate` seeds synthetic host endpoints, runs the probes, then applies bindings enrichment. Techniques compose with a **method-level union** under `reflection > gateway > bindings` precedence: a method exposed only in the JS bindings (e.g. a streaming RPC the gateway can't transcode) is added to a service the gateway partially covers, while a higher-precedence method definition always wins. Reflection's real descriptors are never overwritten with synthetic stubs. Synthetic filenames are namespaced to avoid generator merge conflicts.

### Security

- **SSRF**: the gateway probe validates both the base URL and every `base+path` via `URLValidator`, uses the DNS-rebinding-resistant dialer (re-resolve at connect), does not follow redirects, and bounds fetches (fixed path list, response-size cap, timeout). Private targets still require `--dangerous-allow-private`.
- **Untrusted-input handling**: recovered service/method/type names (from JS bundles and OpenAPI docs on the scanned target) are validated against the proto identifier grammar before descriptor synthesis; the gateway FQN check is anchored so injected control/newline content can't be treated as a service name.

### Out of scope (per the ticket)
- **B3** (`grpc-status: 12` Unimplemented oracle) — depends on LAB-2781's traffic-inference path, not yet implemented.
- **B4** (`import grpcdesc` FileDescriptorSet mining) — optional; can be its own sibling.
- **Standalone `probe grpc *` subcommands** — intentionally not built (pipeline-integrated by design; LAB-3864 B1/B2 criteria amended).

## Test plan
- [x] Unit + pipeline tests across `pkg/analyze`, `pkg/probe`, `pkg/generate/grpc`, and `internal/pipeline` — including the method-level union, gateway `fqn` operationId recovery, the `--probe=false` bindings path, and a mutation-verified regression pinning the gateway-partially-covered → bindings-method carry-through.
- [x] `go test -race ./...` — full suite green.
- [x] `go vet` / `gofmt` clean; `make lint` clean on changed packages (two pre-existing `gosec` G704 findings — `pkg/crawl/http_crawler.go`, `internal/pipeline/wsdl_probe.go` — are byte-identical to `main` and unrelated).
- [x] CI green, including the 80% coverage gate.

Fixes LAB-3864.


saullocarvalho:
⚠️ **Merge order: land #159 first.**

This PR is stacked on #159 (base branch = \`saullobranco/lab-2778-grpc-support\`), and its branch contains #159's commits. Please **merge #159 → \`main\` before this one.**

Once #159 merges, GitHub auto-retargets this PR's base to \`main\` and its diff collapses to just the LAB-3864 delta — no manual retarget needed. Merging this PR *first* would pull #159's foundation code into \`main\` too and leave #159 redundant, so don't merge out of order.

saullocarvalho:
Thanks @coderabbitai — all three items from the latest review are addressed in `260dfda`:

- **`apitype_test.go` (actionable — "APITypeGRPC only produces one probe"):** this was an incomplete merge commit (`42c8edf`) where `apitype.go` was not committed alongside the test. Fixed: `StrategiesForType("grpc")` now returns `GRPCProbe` + `GRPCGatewayProbe` in priority order (reflection → gateway), matching `wantLen: 2`.
- **`apitype_test.go` (nitpick):** added a `checkSecond` hook asserting `strategies[1]` is `*probe.GRPCGatewayProbe`.
- **`grpc_enrich_test.go` (nitpick):** corrected the `grpcServiceNames` doc comment — it iterates `schema.Services` in slice order (no sort), so "sorted" was removed.

Also addressed from the earlier (01:26) review: `enrichGRPCFromBindings` now runs outside the `--probe` gate (gRPC-Web binding recovery is static capture analysis, so it works with `--probe=false`; added a regression test), and the `GRPCGatewayProbe` doc comment in `pkg/probe/doc.go` no longer overstates descriptor synthesis.

Verified at `260dfda`: `gofmt`/`go build`/`go vet` clean, `go test -race ./...` all pass, `golangci-lint` clean (the two `gosec` G704 findings are pre-existing on `main`).

santiago-praetorian (changes requested):
# Praetorian Capability PR Review — LAB-3864 Part B

**Scope reviewed:** B1 (gRPC-Web JS bindings), B2 (grpc-gateway/OpenAPI scrape), and pipeline orchestration (`reflection → gateway → bindings`). B3/B4 and the dropped standalone `probe grpc *` subcommands treated as accepted out-of-scope per the ticket's 2026-07-09 amendment.

**Recommendation: REQUEST CHANGES** — 1 blocker, 0 critical/high. All findings from the prior review rounds (Codex/Claude/Gemini/CodeRabbit) were verified fixed in the current head, including the gateway-unreachable-in-scan-path bug (now fixed via `seedGRPCHostEndpoints`, `pipeline.go`) and the `synthetic.proto` descriptor-collision bug (method-level union + namespaced filenames).

|---|---|---|---|---|---|
| Count | 0 | 0 | 1 | 6 | 0 |

---

## 🔴 Requested change (blocker)

### TEST-001 — No end-to-end test of the seed → gateway-probe → generate chain
`internal/pipeline/pipeline.go` (grpc path in `ClassifyProbeGenerate`)

B2's headline acceptance is that a **pure grpc-gateway target** (HTTP/JSON REST traffic, reflection off) yields a non-empty `.proto` through the `scan`/`generate` pipeline. Every link is tested individually — `seedGRPCHostEndpoints` (isolation), `StrategiesForType` returning the gateway probe, `GRPCGatewayProbe` fetching `/swagger.json` (httptest), and `enrichGRPCFromBindings` + `Generate` (the `PipelineGlue` test) — but **no single test drives the fully-assembled wire**: pure-REST capture + `Probe=true` → seed → `RunStrategies` (live gateway swagger fetch) → `Generate` → non-empty `.proto`. The `PipelineGlue` test hand-builds the gateway's output rather than probing for it, so a regression in the *assembly* (e.g. seed not invoked for `Probe=true`, or probed services not surviving into `Generate`) would go uncaught — and that assembly is exactly the path a prior review found broken.

**Requested:** add one `ClassifyProbeGenerate`-level test — an `httptest` server serving a grpc-gateway `/swagger.json`, fed a pure-REST capture with `Probe=true`, asserting a non-empty `.proto` with the expected services. ~30 lines; locks in the B2 headline path end-to-end.

> Note: this is an integration wire-through gap, not absent coverage — the behavior is otherwise well tested. If you'd prefer to track it as a fast-follow rather than block, that's a reasonable call; flagging it as the blocker because it guards the exact regression class this PR exists to fix.

---

## 🟡 Non-blocking (low)

**Quality**
- **QUAL-001** `internal/pipeline/pipeline.go:115` — the `--dangerous-allow-private` path builds a bare `http.Transport` instead of cloning `DefaultConfig`'s transport; future transport defaults (proxy/TLS/CA) would be silently dropped. Consider `Clone()` + override `DialContext` only. (Also raised by Gemini.)
- **QUAL-002** `pkg/analyze/grpc_web_bindings.go` — `isJSContentTypeForGRPC` duplicates the jsstatic content-type predicate. Rule-of-Three not yet met and the duplication is documented; extract a shared helper if a third copy appears.
- **QUAL-003** `pkg/probe/grpcgateway.go:41` — `grpcGatewayOpenAPIPaths` fixed list doesn't cover per-service swagger files (`/openapiv2/<service>.swagger.json`). Bounding the set is correct SSRF hygiene; just a documented coverage limit.

**Security**
- **SEC-BE-001** `pkg/generate/grpc/generator.go:172` — name-only descriptor synthesis (`unionRecoveredServices` → `FileDescriptorsFromServices`) is not subject to the count/byte caps that guard the reflection path (`enforceDescriptorCaps`). Service names come from untrusted sources (target OpenAPI docs, JS bundles, and the persisted `grpc_schema.services` in an imported capture). Impact is bounded CPU/memory amplification on the operator's machine, not injection (identifiers are grammar-validated) — CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:L (**3.3, low**). Consider mirroring `maxGRPCFileDescriptors`/`maxGRPCDescriptorBytes` at the synthesis entry.

**Tests**
- **TEST-002** `pkg/probe/grpcgateway.go:273` — `serviceFromOperation`'s `info.title` FQN fallback branch (title matches an FQN, no FQN-shaped tag) isn't directly exercised.

**Documentation**
- **DOC-001** `README.md:44` — the gRPC discovery row still describes only reflection-based discovery; it omits the reflection-off techniques added here (JS-bindings + grpc-gateway scrape). `doc.go` files and `CLAUDE.md` were updated to match; README wasn't.

---

## On live-target tests (`test/run-live-tests.sh`)

**Not required for this PR.** `test/run-live-tests.sh` isn't in this PR's diff — it shipped with #159 (Part A/Phase 1), and its `test_grpc_server()` currently exercises only the **reflection** path against `test/grpc-server/`. The ticket explicitly defers live-harness wiring for the gRPC techniques to **LAB-2780** ("integration with setup-live-targets.sh / run-live-tests.sh is non-trivial … Tracked in LAB-2780"). A live grpc-gateway target (to exercise B2 against a real `/swagger.json` rather than httptest) would be valuable but belongs in LAB-2780, not here. The right fix for the gap above is the in-process Go integration test (TEST-001), not a live-harness test.

---

_Automated `capability-pr-review` pass, then human-reviewed. Rerun against the same head SHA + ticket state to reproduce this finding set._

saullocarvalho:
@santiago-praetorian — thanks for the thorough review. All findings addressed; PR is now at `12cf0a2`.

**Blocker**
- **TEST-001** ✅ `0bbfa79` — added `TestClassifyProbeGenerate_GRPCGatewayE2E_PureRESTCaptureYieldsProto`: an `httptest` grpc-gateway `/swagger.json` + a pure-REST capture with `Probe=true` driven through `ClassifyProbeGenerate`, asserting a non-empty `.proto`. It exercises the real seed → gateway-probe → generate assembly (swagger fetched, not hand-built) and is **mutation-verified** — no-oping `seedGRPCHostEndpoints` fails it (`generate failed: no endpoints provided`).

**Low**
- **QUAL-001** ✅ `0bbfa79` — `--dangerous-allow-private` now clones the new `probe.DefaultTransport()` and overrides only `DialContext`; other transport defaults (TLS/idle) preserved, default path still SSRF-safe.
- **SEC-BE-001** ✅ `0bbfa79` + `12cf0a2` — caps now apply to the name-only synthesis path; consolidated into a **single combined ceiling** on the merged reflection+synth set before render (bounded 1× / ≤1000 files / ≤64 MiB, fail-closed) with a regression test. (Note: this is a slight tightening from the 2× you deemed acceptable — flagging in case you prefer the looser bound.)
- **QUAL-003** ✅ `0bbfa79` — documented that per-service swagger paths arent enumerated (unknown service pre-scrape; SSRF-bounded list).
- **TEST-002** ✅ `0bbfa79` — added coverage for `serviceFromOperation`s `info.title` FQN-fallback branch.
- **DOC-001** ✅ `0bbfa79` — README gRPC Key-Features + API-type rows now mention the reflection-off fallbacks.
- **QUAL-002** — left un-extracted per Rule-of-Three (per your note: extract on a third copy). Documented in the commit.

All green: `gofmt`/`go build`/`go vet` clean, `go test -race ./...` pass, `golangci-lint` clean (the 2 `gosec` G704 findings are byte-identical to `main`). Re-requesting your review.

santiago-praetorian (changes requested):
## Praetorian Capability PR Review — follow-up (head `12cf0a2`)

Re-reviewed after the fixes in `0bbfa79` + `12cf0a2`. **All previously-raised findings are verified fixed** — the 7 from my prior human review (including the TEST-001 blocker, now mutation-verified closed by `grpc_gateway_e2e_test.go`, and the SEC-BE-001 descriptor cap) and the 11 from the earlier AI-review rounds.

Requesting changes to land the remaining polish before merge. **For transparency: none of the items below are critical/high/medium and none is a correctness or security blocker on its own** — severities are 4×low + 1×nit. The one I'd genuinely gate on is DOC-001; the rest are defer-OK or optional and are flagged as such so you can triage.

|---|---|---|---|---|---|
| Count | 0 | 0 | 0 | 4 | 1 |

---

### 1. `DOC-001` · low · `CLAUDE.md:76`
CLAUDE.md:76 (pkg/probe Key Packages entry) still reads '...and synthesizes descriptors via `generate/grpc.FileDescriptorsFromServices`.' The prior DOC fix corrected pkg/probe/doc.go to say the probe records service names only and defers synthesis to pkg/generate/grpc, but the identical wrong claim in CLAUDE.md was not updated. This is the exact confusion the earlier fix targeted, now surviving in the repo's primary agent-facing doc.

**Recommendation:** Update the CLAUDE.md pkg/probe sentence: replace '...and synthesizes descriptors via `generate/grpc.FileDescriptorsFromServices`' with '...and records the recovered service names on the endpoint GRPCSchema; descriptor synthesis is deferred to pkg/generate/grpc.'

**Triage:** **Address before merge** — one-line edit; finishes the fix already applied to `doc.go`.

```
CLAUDE.md:76 — '...and synthesizes descriptors via `generate/grpc.FileDescriptorsFromServices`'
pkg/probe/doc.go — 'It does not synthesize descriptors itself — descriptor generation from those names is deferred to pkg/generate/grpc.'
```

---

### 2. `QUAL-002` · low · `internal/pipeline/grpc_enrich.go:185`
filterUncoveredServices marks an FQN covered only when it appears in GRPCSchema.Services. A reflection endpoint stores results in FileDescriptors (raw bytes) and may have an empty Services slice, so a bindings-recovered FQN already defined by reflection is treated as 'uncovered' and carried through as a trailing synthetic endpoint. The generator's unionRecoveredServices backstop (parseDescriptorSet -> reflectedFQNs) drops it, so output is correct, but the redundant service is passed through unnecessarily. nameOnlyCoveredServices already uses the correct 'ReflectionEnabled || len(FileDescriptors)>0' heuristic; filterUncoveredServices is inconsistent with it.

**Recommendation:** Mirror nameOnlyCoveredServices's heuristic in filterUncoveredServices: treat an endpoint's FQNs as covered when Services lists them OR the endpoint carries FileDescriptors (reflection ran). Minimal, avoids an expensive extra parseDescriptorSet.

**Triage:** **Defer OK** (cheap fast-follow) — output is correct today via the `unionRecoveredServices` backstop.

```
grpc_enrich.go:185-201 — filterUncoveredServices only checks s.Services, not FileDescriptors
grpc_enrich.go:217-230 — nameOnlyCoveredServices uses 'reflected := s.ReflectionEnabled || len(s.FileDescriptors) > 0'
```

---

### 3. `SEC-BE-001` · low · `pkg/generate/grpc/generator.go:85`
SEC-BE-001's primary vector is resolved: the authoritative combined cap (generator.go ~99) fails closed on the merged reflection+synth set BEFORE the expensive renderProto -> desc.CreateFileDescriptorsFromSet -> protoprint step, and is mutation-verified by TestGenerator_Generate_SynthesizedDescriptorsExceedCountCapErrors. RESIDUAL: mergeRecoveredServices -> unionRecoveredServices -> FileDescriptorsFromServices -> proto.Marshal synthesizes/serializes the full recovered-service set BEFORE the cap check. This pre-cap work is a single linear pass whose size is input-bounded (OpenAPI docs capped 5 MB/fetch; JS bundles capped 5 MB each; a malicious imported capture's persisted grpc_schema.services is bounded by file size), so it is minor local CPU/alloc pressure, not exhaustion. Identifiers remain grammar-validated (no injection). LOW.

**Recommendation:** Optional hardening (non-blocking): add a service/method count check inside unionRecoveredServices or at the FileDescriptorsFromServices entry (truncate + debug-log overflow) so the intermediate marshal is bounded before proto.Marshal. Shippable as-is since the quadratic parse/render is already capped.

**Triage:** **Defer OK** (optional hardening) — primary amplification is already gated; residual is input-bounded.

```
if err := mergeRecoveredServices(merged, endpoints); err != nil {
if err := enforceDescriptorCaps(merged); err != nil {
synthServices := unionRecoveredServices(endpoints, reflectedFQNs)
raw, err := proto.Marshal(fdp)
```

---

### 4. `QUAL-003` · low · `pkg/generate/grpc/generator.go:75`
Generate calls enforceDescriptorCaps twice. The first (pre-parse) call runs on the reflection-only map before mergeRecoveredServices. On the common name-only path (gateway-only/bindings-only capture) merged is empty, so the first call is a no-op AND mergeRecoveredServices skips parseDescriptorSet entirely (len(merged)==0 branch). The comment claims the first call bounds that parse, but on the empty-merged path there is no parse to bound — the authoritative combined cap (line ~99) is the only effective bound. Not a correctness bug (combined cap is sound), but the comment implies dual protection that does not exist on the name-only path.

**Recommendation:** Clarify the comment to state the pre-parse guard only applies when reflection descriptors are present, and the combined cap is the sole bound on the name-only path; or wrap the first call in 'if len(merged) > 0'.

**Triage:** **Optional** — comment-accuracy only; the cap logic itself is correct.

```
generator.go:75-81 — pre-parse guard called unconditionally before the len(merged) check
generator.go — mergeRecoveredServices skips parseDescriptorSet when len(merged)==0
```

---

### 5. `QUAL-001` · nit · `internal/pipeline/grpc_enrich.go:76`
grpcHostKey (grpc_enrich.go:76-86) and openAPIBaseURL (grpcgateway.go:424-438) both parse a URL to scheme://host, differing only in scheme normalization (grpcHostKey preserves scheme / empty->https; openAPIBaseURL maps grpc/grpcs->http/https). Two copies of the same concern with subtly different rules; Rule-of-Three not met so no extraction required, but the divergence could mislead maintainers into treating them as equivalent. No incorrect behavior today.

**Recommendation:** Add a cross-referencing inline comment in each noting the intentional scheme-handling difference; extract a shared helper only if a third call site appears.

**Triage:** **Optional** (nit) — Rule-of-Three not met; a cross-ref comment suffices.

```
grpc_enrich.go:76-86 — grpcHostKey preserves original scheme
grpcgateway.go:424-438 — openAPIBaseURL maps grpc/grpcs to http/https
```

---

### Suggested path to merge
- **Now:** DOC-001 (one line in `CLAUDE.md`).
- **Follow-up ticket** (optional, recommended): QUAL-002 + SEC-BE-001 as "gRPC reflection-off hardening polish" — both are real but currently masked by a backstop / input bounds.
- **Optional / can dismiss:** QUAL-003, QUAL-001.

_Automated `capability-pr-review` pass (re-run @ `12cf0a2`), human-reviewed._

saullocarvalho:
@santiago-praetorian thanks for the follow-up pass — all five points are addressed in `bbb5604` (pushed on top of `12cf0a2`). Summary of each:

**DOC-001** — `CLAUDE.md`'s `pkg/probe` entry now matches the `doc.go` correction: `GRPCGatewayProbe` records the recovered service names on the endpoint's `GRPCSchema`; descriptor synthesis is deferred to `pkg/generate/grpc` (`FileDescriptorsFromServices`).

**SEC-BE-001** — bounded the pre-cap synthesis marshal. New `capSynthesizedServices` at the `FileDescriptorsFromServices` entry truncates the synthesized set to `maxSynthesizedPackages` (= `maxGRPCFileDescriptors + 1` = 1001) distinct proto packages *before* `proto.Marshal`, `slog.Debug`-logging any overflow. The deliberate `+1` headroom keeps an over-limit set tripping the authoritative combined `enforceDescriptorCaps` (fail-closed) rather than slipping under it — truncating at exactly the cap would have let a would-be-rejected set render silently. Pinned by three tests: `TestCapSynthesizedServices_TruncatesBeyondCap` (mutation-verified — identity-replacing the helper turns it red), `TestCapSynthesizedServices_NoTruncationUnderCap`, and `TestGenerator_Generate_SynthesizedServicesOverSynthesisCapStillErrors` (post-truncation still fails closed).

**QUAL-002** — on `filterUncoveredServices`: I went with documentation rather than a behavior change, and want to flag the reasoning for your call. Precise per-FQN dedup of the rare *Services-empty-but-`FileDescriptors`-present* reflection endpoint requires a `parseDescriptorSet` (the expensive parse your recommendation explicitly wanted to avoid); keying coverage on `len(FileDescriptors) > 0` alone can't yield per-FQN coverage without over-dropping legitimate bindings-only services. In practice the live probe populates `.Services` alongside `FileDescriptors` (via `extractService` in `runReflection`), and imported captures persist `.Services` but not `FileDescriptors` (`json:"-"`) — so `.Services` is the durable dedup signal. The rare gap is a harmless redundant pass-through that the generator's `unionRecoveredServices → reflectedFQNs` backstop drops, so the emitted `.proto` is already correct. I added a comment documenting this and cross-referencing `nameOnlyCoveredServices`. Happy to spin the parse-based version into a follow-up ticket if you'd prefer the code path over the doc.

**QUAL-003** — the pre-parse `enforceDescriptorCaps(merged)` is now wrapped in `if len(merged) > 0` and the comment corrected: on the name-only path `merged` is empty (no reflection parse to guard), so the combined cap is the sole effective bound there.

**QUAL-001 (nit)** — added reciprocal cross-reference comments to `grpcHostKey` and `openAPIBaseURL` noting the intentional scheme-handling divergence (preserve original vs `grpc`/`grpcs`→`http`/`https`) and that unifying them isn't warranted (Rule of Three not met).

Verification: `gofmt`/`go vet` clean, `go test -race ./...` green (incl. the new synthesis-cap tests), `make lint` shows only the two pre-existing G704 findings in `wsdl_probe.go` / `http_crawler.go` (unchanged files, byte-identical on the stashed baseline — zero new lint issues). Re-requesting your review.

santiago-praetorian:

**PR:** https://github.com/praetorian-inc/vespasian/pull/160
**Linear:** https://linear.app/praetorianlabs/issue/LAB-3864
**Head SHA:** bbb5604ca8be098d6d6552687a5f17835d6602ef
**Reviewed at:** 2026-07-10

**Recommendation:** APPROVE — the findings below are all non-blocking (optional hardening / fast-follow / nit); none gate merge.

|---|---|---|---|---|---|
| Count | 0 | 0 | 0 | 4 | 1 |

---

_Scope reviewed: Part B — B1 (gRPC-Web JS bindings), B2 (grpc-gateway/OpenAPI scrape), and the `reflection → gateway → bindings` pipeline orchestration. Standalone `probe grpc *` subcommands (dropped per the ticket's 2026-07-09 amendment) and B3/B4 (deferred to siblings) treated as accepted out-of-scope. All three in-scope acceptance criteria (B1, B2, pipeline orchestration + per-technique failure isolation) verified **met** with backing tests. All 18 findings from the prior review rounds (11 AI + 7 human round-1 + 5 human round-2) were re-verified fixed at this head and are not re-reported. The five items below are fresh, all non-blocking polish/fast-follow._

<details>

<details>

<details>
<summary><code>SEC-BE-001</code> · low · <code>pkg/generate/grpc/synthesize.go:327</code></summary>

**Issue:** `capSynthesizedServices` caps the number of distinct proto *packages* (`maxSynthesizedPackages` = 1001) but not the total services/methods within a single package. A recovered set where every service shares one package is not truncated, so `proto.Marshal` runs on the full set before the authoritative combined byte cap `enforceDescriptorCaps` (≤1000 files / ≤64 MiB, fail-closed) runs. This refines the previously-accepted SEC-BE-001 residual. Recovered names come only from size-bounded inputs (5 MB OpenAPI docs / 5 MB JS bundles), so the pre-cap marshal is an input-bounded linear pass and impact is minor, bounded, local CPU/alloc pressure on the operator's own machine. The backend-security reviewer independently traced this path and found no exploitable exhaustion; identifiers are grammar-validated (no injection). CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:L (3.3, low).

**Recommendation:** Optional hardening (non-blocking): bound services/methods per package inside `capSynthesizedServices` (truncate + `slog.Debug` overflow) so the intermediate marshal is bounded even in the single-package case, and add a same-package fixture test; or document that the combined byte cap is the sole bound for same-package sets. Reasonable to track as the "gRPC reflection-off hardening polish" fast-follow alongside the residual already noted in round 2.

```go
func capSynthesizedServices(services []classify.GRPCService) []classify.GRPCService {
	if len(pkgs) >= maxSynthesizedPackages {
```

</details>

</details>

<details>

<details>
<summary><code>QUAL-001</code> · low · <code>pkg/analyze/grpc_web_bindings.go:75</code></summary>

**Issue:** `mergeServices` (`pkg/analyze/grpc_web_bindings.go:75-101`) and `buildGatewayServices` (`pkg/probe/grpcgateway.go:295-325`) repeat the same map→sorted-slice flatten skeleton verbatim across two new files. Rule-of-Three is not yet met (two copies), so extraction is not mandatory — but worth a note.

**Recommendation:** Leave as-is for now, or add a cross-reference comment (consistent with how the analogous `isJSContentTypeForGRPC` duplication was handled in a prior round); extract a shared `sortedServicesFromMap` helper only if a third call site appears.

```go
	if len(byService) == 0 {
		return nil
	}
	names := make([]string, 0, len(byService))
	for name := range byService {
		names = append(names, name)
	}
	sort.Strings(names)
```

</details>

<details>
<summary><code>QUAL-002</code> · nit · <code>pkg/probe/doc.go:39</code></summary>

**Issue:** The `GRPCGatewayProbe` doc block breaks a sentence across lines so an em-dash is left dangling at the end of a line mid-sentence. Cosmetic godoc formatting only.

**Recommendation:** Reflow the two comment lines so the em-dash clause stays on one logical sentence.

```go
//     descriptors itself —
//     descriptor generation from those names is deferred to pkg/generate/grpc.
```

</details>

</details>

<details>
<summary>Tests (2)</summary>

<details>
<summary><code>TEST-001</code> · low · <code>pkg/probe/grpcgateway.go:105</code></summary>

**Issue:** The `MaxEndpoints` host-cap guard (`len(seen) >= p.config.MaxEndpoints`) inside `GRPCGatewayProbe.Probe` has no test hitting the boundary; existing tests use ≤2 hosts with `MaxEndpoints:100`. This is an uncovered edge branch (coverage-of-edge-cases) within otherwise well-tested public behavior — not absent coverage of new behavior — hence low and non-blocking.

**Recommendation:** Add a table case with more candidate hosts than `MaxEndpoints` and assert the probe stops at the cap.

```go
		if len(seen) >= p.config.MaxEndpoints {
```

</details>

<details>
<summary><code>TEST-002</code> · low · <code>pkg/probe/grpcgateway.go:435</code></summary>

**Issue:** `seedGRPCHostEndpoints` emits `grpc://host` synthetic endpoints, but every gateway test uses `http://` httptest URLs, so `openAPIBaseURL`'s `grpc→http` / `grpcs→https` mapping branch (which the pipeline's seeding relies on) is never exercised (helper ~75% covered). Uncovered edge branch of tested behavior — low, non-blocking.

**Recommendation:** Add a unit case feeding a `grpc://` (and `grpcs://`) endpoint through the `openAPIBaseURL` path and assert the `http`/`https` rewrite.

```go
	case "grpc", "http":
```

</details>

</details>

</details>

---

santiago-praetorian (approved):
Approving — re-reviewed at head `bbb5604`. All 23 findings from the prior review rounds are verified fixed; the 5 fresh items posted in the review comment above are all non-blocking (1 low SEC already backstopped, 1 DRY note where Rule-of-Three isn't met, 1 doc nit, 2 edge-branch coverage gaps). B1/B2/pipeline acceptance criteria met with backing tests. Recommend folding SEC-BE-001 + the two TEST gaps into the existing gRPC reflection-off hardening/coverage follow-up, and a one-line fix for the doc nit if convenient. None block merge.

PR #164, opened 2026-07-06
ci: add live test suite to GitHub Actions (LAB-4012)

eli-wald:
## Summary

- Adds `live-tests.yml` workflow with two parallel jobs: **offline-tests** (imports, deterministic generates, classifier/spec edge cases — no services needed) and **live-tests** (boots 4 target services, installs Chrome, runs crawl-based tests)
- Label-gated on PRs (`live-tests` label required), always runs on push to `main` and `workflow_dispatch`
- Teardown and artifact upload run unconditionally (`if: always()`)

Closes [LAB-4012](https://linear.app/praetorianlabs/issue/LAB-4012/vespasian-run-the-live-test-suite-in-github-actions-ci)

## Test plan

- [x] Offline tests validated locally via `act` — all 19 tests pass
- [x] Live tests validated locally via `act` — workflow structure correct, Chrome step fails due to `act` slim image limitation (not a real GHA issue)
- [x] Verify both jobs pass on real GHA runner (this PR push triggers the workflow)
- [x] Verify PR label gating works (remove `live-tests` label → jobs should skip)
- [x] Verify deliberate test break fails the job

### DIFF
--- a/.github/workflows/live-tests.yml
+++ b/.github/workflows/live-tests.yml
@@ -5,12 +5,12 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
-    types: [opened, synchronize, reopened, labeled]
+    types: [opened, synchronize, reopened, labeled, unlabeled]
   workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 permissions:
   contents: read
@@ -20,6 +20,7 @@ jobs:
   # Push-to-main and workflow_dispatch always run.
   check-label:
     runs-on: ubuntu-latest
+    timeout-minutes: 5
     outputs:
       should-run: ${{ steps.gate.outputs.run }}
     steps:
@@ -43,6 +44,8 @@ jobs:
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
+        with:
+          persist-credentials: false
 
       - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5  # v5.5.0
         with:
@@ -81,8 +84,12 @@ jobs:
     if: needs.check-label.outputs.should-run == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    env:
+      VESPASIAN_NO_SANDBOX: "true"
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
+        with:
+          persist-credentials: false
 
       - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5  # v5.5.0
         with:
@@ -103,10 +110,6 @@ jobs:
       - name: Run live tests
         run: ./test/run-live-tests.sh --no-build --targets rest-api,soap-service,graphql-server,concat-spa,edge-cases,crawl-depth
 
-      - name: Teardown
-        if: always()
-        run: ./test/setup-live-targets.sh --teardown
-
       - name: Upload results
         if: always()
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02  # v4.6.2
@@ -114,3 +117,7 @@ jobs:
           name: live-test-results
           path: test/.results/
           retention-days: 7
+
+      - name: Teardown
+        if: always()
+        run: ./test/setup-live-targets.sh --teardown
--- a/.github/workflows/live-tests.yml
+++ b/.github/workflows/live-tests.yml
@@ -1,0 +1,116 @@
+name: Live Tests
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize, reopened, labeled]
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  # Gate: skip PR runs unless the 'live-tests' label is present.
+  # Push-to-main and workflow_dispatch always run.
+  check-label:
+    runs-on: ubuntu-latest
+    outputs:
+      should-run: ${{ steps.gate.outputs.run }}
+    steps:
+      - id: gate
+        run: |
+          if [[ "${{ github.event_name }}" != "pull_request" ]]; then
+            echo "run=true" >> "$GITHUB_OUTPUT"
+          elif [[ "${{ contains(github.event.pull_request.labels.*.name, 'live-tests') }}" == "true" ]]; then
+            echo "run=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "run=false" >> "$GITHUB_OUTPUT"
+            echo "::notice::Skipping live tests — add the 'live-tests' label to run them on this PR."
+          fi
+
+  # Offline tests: imports, deterministic generates, classifier/spec edge cases.
+  # No services or Chrome needed — runs fast (~1-2 min).
+  offline-tests:
+    needs: check-label
+    if: needs.check-label.outputs.should-run == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
+
+      - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5  # v5.5.0
+        with:
+          go-version-file: go.mod
+
+      - name: Build vespasian
+        run: go build -o bin/vespasian ./cmd/vespasian
+
+      - name: Write stub config
+        run: |
+          printf '%s\n' \
+            "REST_API_PORT=8990" \
+            "SOAP_SERVICE_PORT=8991" \
+            "GRAPHQL_SERVER_PORT=8992" \
+            "CONCAT_SPA_PORT=8993" \
+            "TARGETS_SETUP=" \
+            > test/.live-test-config
+
+      - name: Run offline tests
+        run: |
+          ./test/run-live-tests.sh --no-start --no-build \
+            --targets import-burp,import-har,import-base64,import-mitmproxy,import-mitmproxy-native,import-unicode,import-duplicates,import-malformed,import-empty,generate-rest,generate-wsdl,generate-wsdl-matrix,generate-graphql,generate-graphql-imports,generate-js-static,generate-merge-slugs,crawl-unreachable,classifier-edge,spec-edge
+
+      - name: Upload results
+        if: always()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02  # v4.6.2
+        with:
+          name: offline-test-results
+          path: test/.results/
+          retention-days: 7
+
+  # Live tests: boots target services, crawls them, generates specs, validates.
+  # Requires Go (targets), Node (GraphQL/Apollo), and Chrome (rod backend).
+  live-tests:
+    needs: check-label
+    if: needs.check-label.outputs.should-run == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
+
+      - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5  # v5.5.0
+        with:
+          go-version-file: go.mod
+
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020  # v4.4.0
+        with:
+          node-version: 20
+
+      - name: Install Chrome
+        uses: browser-actions/setup-chrome@2e1d749697dd1612b833dba4a722266286fbefcd  # v2.1.2
+        with:
+          chrome-version: stable
+
+      - name: Setup live targets
+        run: ./test/setup-live-targets.sh
+
+      - name: Run live tests
+        run: ./test/run-live-tests.sh --no-build --targets rest-api,soap-service,graphql-server,concat-spa,edge-cases,crawl-depth
+
+      - name: Teardown
+        if: always()
+        run: ./test/setup-live-targets.sh --teardown
+
+      - name: Upload results
+        if: always()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02  # v4.6.2
+        with:
+          name: live-test-results
+          path: test/.results/
+          retention-days: 7
--- a/.github/workflows/live-tests.yml
+++ b/.github/workflows/live-tests.yml
@@ -19,7 +19,7 @@ jobs:
   # Gate: skip PR runs unless the 'live-tests' label is present.
   # Push-to-main and workflow_dispatch always run.
   check-label:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     timeout-minutes: 5
     outputs:
       should-run: ${{ steps.gate.outputs.run }}
@@ -40,7 +40,7 @@ jobs:
   offline-tests:
     needs: check-label
     if: needs.check-label.outputs.should-run == 'true'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
@@ -82,7 +82,7 @@ jobs:
   live-tests:
     needs: check-label
     if: needs.check-label.outputs.should-run == 'true'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     timeout-minutes: 20
     env:
       VESPASIAN_NO_SANDBOX: "true"
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -129,4 +129,7 @@ See `test/README.md` for how to run the suite, including the `TEST_HOST` overrid
 
 ## CI
 
-GitHub Actions runs on push to main and PRs: build, test (with 80% coverage threshold), lint (golangci-lint v2), and format check.
+GitHub Actions runs on push to main and PRs:
+
+- **ci.yml**: Build, test (`go test -race`, 80% coverage threshold), lint (golangci-lint v2), and format check. Runs on all pushes and PRs.
+- **live-tests.yml**: Two parallel jobs — **offline-tests** (imports, deterministic generates, classifier/spec edge cases; no services or Chrome needed) and **live-tests** (boots REST, SOAP, GraphQL, and concat-SPA target services, installs Chrome, runs crawl-based end-to-end tests). Label-gated on PRs: add the `live-tests` label to trigger. Always runs on push to `main` and `workflow_dispatch`.
--- a/pkg/crawl/browser.go
+++ b/pkg/crawl/browser.go
@@ -56,16 +56,13 @@ type BrowserManager struct {
 	cleanupOnce sync.Once
 }
 
-// NewBrowserManager launches a Chrome instance with the given options and
-// returns a manager that owns its lifecycle.
-func NewBrowserManager(opts BrowserOptions) (*BrowserManager, error) {
+// configureLauncher applies BrowserOptions to a new launcher without
+// launching Chrome. Extracted for testability of the CI auto-detection logic.
+func configureLauncher(opts BrowserOptions) (*launcher.Launcher, error) {
 	l := launcher.New().
 		Headless(opts.Headless)
 
-	if opts.NoSandbox {
-		l = l.NoSandbox(true)
-	}
-	if !opts.NoSandbox && os.Getenv("CI") != "" {
+	if opts.NoSandbox || os.Getenv("CI") != "" {
 		l = l.NoSandbox(true)
 	}
 	if opts.ChromePath != "" {
@@ -76,6 +73,17 @@ func NewBrowserManager(opts BrowserOptions) (*BrowserManager, error) {
 			return nil, err
 		}
 		l = l.Set("proxy-server", opts.Proxy)
+	}
+
+	return l, nil
+}
+
+// NewBrowserManager launches a Chrome instance with the given options and
+// returns a manager that owns its lifecycle.
+func NewBrowserManager(opts BrowserOptions) (*BrowserManager, error) {
+	l, err := configureLauncher(opts)
+	if err != nil {
+		return nil, err
 	}
 
 	wsURL, err := l.Launch()
--- a/pkg/crawl/browser.go
+++ b/pkg/crawl/browser.go
@@ -57,12 +57,13 @@ type BrowserManager struct {
 }
 
 // configureLauncher applies BrowserOptions to a new launcher without
-// launching Chrome. Extracted for testability of the CI auto-detection logic.
+// launching Chrome. Disables the sandbox when opts.NoSandbox is set or
+// when the VESPASIAN_NO_SANDBOX env var is "true" (set by CI workflows).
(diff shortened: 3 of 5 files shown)


santiago-praetorian:

**PR:** https://github.com/praetorian-inc/vespasian/pull/164
**Linear:** https://linear.app/praetorianlabs/issue/LAB-4012/vespasian-run-the-live-test-suite-in-github-actions-ci
**Head SHA:** 0fa4de0301a83aeb0c18ae3822ecc03fa7110cd2
**Reviewed at:** 2026-07-09T14:13:44Z

|---|---|---|---|---|---|
| Count | 1 | 2 | 5 | 5 | 1 |
| Blockers | 1 | 2 | 5 | 0 | 0 |

---
> **Note on the automated bot reviews:** the issues raised on this PR by **CodeRabbit** and the **GitHub Actions** reviewers (Codex, Gemini, Claude) also need to be addressed. Several overlap with the findings below — the teardown-before-upload artifact loss, the `CI`-env sandbox behavior, and `persist-credentials: false` on the checkout steps — and the remainder that this review does not separately reproduce (e.g. re-running the label gate on `unlabeled` so it flips back to skip, and the `CI=false`/`CI=0` truthiness edge) should be triaged and either fixed or explicitly dismissed with a reason.

<details>
<summary><strong>Blockers (8)</strong></summary>

<details>
<summary><code>QUAL-001</code> · medium · QUAL · <code>.github/workflows/live-tests.yml:11</code></summary>

**Issue:** The concurrency group is keyed on `${{ github.workflow }}-${{ github.ref }}` with cancel-in-progress: true applied unconditionally. On push to main this ref is the same for every push (refs/heads/main), so if two commits land on main in quick succession (e.g. two PRs merged back-to-back), the in-progress live-tests run for the first commit is cancelled before completion. Given AC2 ('The job fails the build when a live test regresses'), a cancelled run reports neutral/cancelled status rather than failure, so a regression introduced by the first commit could go unnoticed if the second commit's run happens to pass.

**Recommendation:** Scope cancel-in-progress to PR runs only, e.g. `cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}`, so main-branch runs always complete and report their own pass/fail status.

```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

</details>
<details>
<summary><code>QUAL-005</code> · high · QUAL · <code>.github/workflows/live-tests.yml:106</code></summary>

**Issue:** In the live-tests job, the 'Teardown' step (line 106-108, if: always()) runs ./test/setup-live-targets.sh --teardown before the 'Upload results' step (line 110-116, if: always()). test/setup-live-targets.sh's do_teardown() unconditionally executes `rm -rf "${SCRIPT_DIR}/.results"` (test/.results), which is exactly the directory the subsequent Upload results step uploads (path: test/.results/). Because both steps run unconditionally on failure (if: always()), a failing live-test run — the case where diagnostics matter most — will have its results wiped by Teardown before Upload results executes, so the uploaded artifact will be empty/missing. This defeats the purpose of artifact upload for debugging live-test failures, which is a stated design goal of this workflow.

**Recommendation:** Reorder the steps so 'Upload results' runs before 'Teardown', or have the teardown script preserve/not touch test/.results (e.g. drop the `rm -rf "${SCRIPT_DIR}/.results"` line from --teardown, or add a --keep-results flag used by CI).

```
      - name: Teardown
        if: always()
        run: ./test/setup-live-targets.sh --teardown

      - name: Upload results
        if: always()
```

</details>
<details>
<summary><code>QUAL-006</code> · high · QUAL · <code>pkg/crawl/browser.go:65</code></summary>

**Issue:** configureLauncher now disables the Chrome sandbox whenever os.Getenv("CI") != "", regardless of the caller-supplied BrowserOptions.NoSandbox value. The BrowserOptions.NoSandbox field doc comment (lines 32-35) states it 'removes a primary exploit mitigation barrier and should only be set in containerized or CI environments where the sandbox cannot be enabled' — i.e. it documents an explicit, caller-controlled opt-in. This change breaks that contract for every existing call site: pkg/crawl/rod_crawler.go:45 and cmd/vespasian/main.go:323 both call NewBrowserManager(BrowserOptions{Headless: true, ...}) without setting NoSandbox, and pkg/crawl is consumed by pkg/sdk (per CLAUDE.md: 'exposing the vespasian pipeline to chariot/Guard hosts'), which commonly runs under CI=true in production pipelines. Those consumers would silently crawl untrusted target pages with the OS sandbox disabled, with no way to opt back in since the check is unconditional (opts.NoSandbox || os.Getenv("CI") != ""). Since vespasian's core function is crawling potentially adversarial/untrusted URLs, the sandbox is a meaningful isolation boundary against renderer exploits, and disabling it ambiently is a real defense-in-depth regression, not just a CI convenience.

**Recommendation:** Do not gate sandbox behavior on the ambient CI variable. Either require an explicit, project-owned opt-in (e.g. a dedicated env var like VESPASIAN_CI_NO_SANDBOX, or have the live-tests.yml workflow set opts.NoSandbox=true directly when constructing BrowserOptions), or fall back to NoSandbox only after a sandboxed launch attempt fails. Keep opts.NoSandbox as the sole caller-facing control for library consumers.

```
if opts.NoSandbox || os.Getenv("CI") != "" {
l = l.NoSandbox(true)
```

</details>
<details>
<summary><code>QUAL-007</code> · medium · QUAL · <code>pkg/crawl/browser.go:65</code></summary>

**Issue:** os.Getenv("CI") != "" only checks that the variable is set to a non-empty string, not that it is truthy. A developer who explicitly sets CI=false or CI=0 to disable CI-specific behavior while testing locally will still have the sandbox disabled, because "false" and "0" are both non-empty strings. This is a separate correctness bug from the caller-intent issue above: even accepting the ambient-env-var design, the comparison itself doesn't match common CI-detection conventions used elsewhere (many tools parse CI as a boolean, not presence-of-any-value).

**Recommendation:** If an ambient CI signal is kept at all, parse it as a boolean (e.g. strconv.ParseBool, treating parse errors/empty as false) rather than checking for non-empty string.

```
if opts.NoSandbox || os.Getenv("CI") != "" {
```

</details>
<details>
<summary><code>SEC-BE-002</code> · critical · SEC-BE · <code>pkg/crawl/browser.go:65</code></summary>

**Issue:** configureLauncher enables Chrome's --no-sandbox whenever os.Getenv("CI") is non-empty, regardless of whether the caller opted in via BrowserOptions.NoSandbox. pkg/crawl is a shared library consumed by pkg/sdk, which exposes the vespasian pipeline to chariot/Guard hosts (production). The production path proves the exposure: pkg/sdk/capability.go:226 defaultCrawl builds crawl.CrawlerOptions with Headless:true and no NoSandbox, and rod_crawler.go:45 then calls NewBrowserManager(BrowserOptions{Headless:true, Proxy:...}) — again without NoSandbox. When any of those hosts run under CI=true (near-universal in CI/CD and many container pipeline runners), the OS sandbox is dropped without the consumer ever requesting it. The crawler drives a real headless Chrome against untrusted remote target content, so the renderer sandbox is the primary exploit-mitigation barrier against a malicious page escaping the renderer to achieve code execution on the host. Auto-disabling it based on a ubiquitous ambient variable is a silent defense-in-depth regression for downstream production consumers. This is an env-driven security-posture change, not a caller-visible option; it contradicts the struct's own doc comment on NoSandbox which states the sandbox 'should only be set in containerized or CI environments where the sandbox cannot be enabled'.

**Recommendation:** Do not couple sandbox-disabling to the generic CI variable. Gate it on an explicit, project-owned opt-in that the CI workflow sets (e.g., a VESPASIAN_NO_SANDBOX env var, or have the CI harness pass BrowserOptions.NoSandbox=true). Prefer a launch-then-fallback approach: attempt a sandboxed launch first and only retry with NoSandbox after a sandbox-specific launch failure, rather than pre-emptively disabling based on environment. Ensure pkg/sdk consumers keep the sandbox by default in production.

```
	if opts.NoSandbox || os.Getenv("CI") != "" {
		l = l.NoSandbox(true)
	}
```

</details>
<details>
<summary><code>SEC-BE-003</code> · medium · SEC-BE · <code>pkg/crawl/browser.go:65</code></summary>

**Issue:** The condition os.Getenv("CI") != "" treats ANY non-empty value as an instruction to disable the sandbox, so CI=false and CI=0 both evaluate truthy and still remove the sandbox. An operator who deliberately sets CI=false to opt out of CI-specific behavior, or who has correctly provisioned container capabilities (e.g. a seccomp profile) and explicitly wants sandbox isolation, has no mechanism to keep it: once CI is set to anything non-empty, the OR short-circuit forces NoSandbox(true) with no BrowserOptions field or env override that can re-enable the sandbox. This is a logic defect that makes the primary defense-in-depth regression un-opt-outable and can surprise developers running local scripts under a set CI variable.

**Recommendation:** If an env-based toggle is retained, parse it as a boolean (e.g. strconv.ParseBool) so false/0 disable the behavior, and add an explicit caller-controlled way to force the sandbox ON that takes precedence over the env check (e.g. a BrowserOptions.ForceSandbox / tri-state field). Better: remove the env coupling entirely and rely solely on the explicit BrowserOptions.NoSandbox opt-in.

```
	if opts.NoSandbox || os.Getenv("CI") != "" {
```

</details>
<details>
<summary><code>TEST-001</code> · medium · TEST · <code>pkg/crawl/browser.go:68</code></summary>

**Issue:** configureLauncher() was extracted 'for testability', but the only new test (TestConfigureLauncher_CIAutoDetectsNoSandbox) always passes BrowserOptions{NoSandbox: ...} with no Proxy and no ChromePath, so err is always nil. Two of the function's three configuration branches have zero direct coverage: (a) the ChromePath branch (l.Bin(opts.ChromePath)), and (b) the Proxy branch, including the new err != nil early return that NewBrowserManager now relies on to abort before launching Chrome. validateProxyAddr itself is covered by TestValidateProxyAddr in crawler_test.go, so the residual gap is the wiring inside configureLauncher: that an invalid Proxy causes configureLauncher to return an error (rather than the launcher swallowing it) and that a valid Proxy actually sets the proxy-server flag. These are trivially testable without launching Chrome (same style as the existing table test using l.Has(...)).

**Recommendation:** Add table cases to TestConfigureLauncher (or a sibling test) that call configureLauncher(BrowserOptions{Proxy: "http://bad addr"}) and assert err != nil, configureLauncher(BrowserOptions{Proxy: "http://127.0.0.1:8080"}) and assert err == nil && l.Has("proxy-server"), and configureLauncher(BrowserOptions{ChromePath: "/tmp/chrome"}) asserting the binary override is applied. This exercises the error-return path NewBrowserManager depends on.

```
	if opts.ChromePath != "" {
		l = l.Bin(opts.ChromePath)
	}
	if opts.Proxy != "" {
		if err := validateProxyAddr(opts.Proxy); err != nil {
			return nil, err
		}
```

</details>
<details>
<summary><code>REQ-001</code> · medium · REQ · <code>test/live-test-gaps.md</code></summary>

**Issue:** AC5 requires test/live-test-gaps.md to be reconciled with what CI now covers. That file does not exist in the PR worktree and does not exist on origin/main, and the PR diff (4 files: live-tests.yml, CLAUDE.md, pkg/crawl/browser.go, pkg/crawl/browser_test.go) neither creates nor references any gaps document (0 'gap' hits in the diff; nothing found via find -iname '*gap*'). The acceptance criterion is therefore completely unaddressed by this PR; there is no artifact documenting which live-test scenarios CI now runs versus which remain uncovered.

**Recommendation:** Add test/live-test-gaps.md (or update the existing gaps doc if one is expected) enumerating the live-test coverage the new live-tests.yml provides vs. remaining gaps: e.g. note that CI runs rest-api, soap-service, graphql-server, concat-spa, edge-cases, and crawl-depth targets, that offline-tests cover the 19 import/generate/classifier/spec cases, and record known uncovered areas. Reconcile the doc's 'covered' section with the exact --targets lists in the workflow so the doc stays accurate.

</details>

</details>

<details>

<details>

<details>
<summary><code>SEC-BE-001</code> · low · <code>.github/workflows/live-tests.yml:51</code></summary>

**Issue:** Both actions/checkout steps (offline-tests and live-tests jobs) rely on default credential persistence, leaving the GITHUB_TOKEN in the on-runner .git/config for the duration of the job. This is a defense-in-depth gap: a compromised or misbehaving later step (the workflow runs target services and an untrusted-content crawl in the live-tests job) could read the persisted token, and artifact uploads could inadvertently include it. Concrete impact here is limited because the top-level permissions block is contents:read (minimal token scope) and the upload-artifact paths are scoped to test/.results/ (not the repo checkout / .git), so no direct exfiltration path is present in this diff. Reported as hardening consistent with least-credential-persistence.

**Recommendation:** Set persist-credentials: false under with: on each actions/checkout step, since neither job pushes back to the repo. Keep the contents:read least-privilege permissions block.

```
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
permissions:
  contents: read
```

</details>

</details>
<details>

<details>
<summary><code>QUAL-002</code> · low · <code>.github/workflows/live-tests.yml:45</code></summary>

**Issue:** The offline-tests job (lines 45-49) and live-tests job (lines 85-89) both contain the identical checkout@11bd71901bbe5b1630ceea73d27597364c9af683 + setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 (go-version-file: go.mod) step pair. This is 3+ contiguous non-comment lines copy-pasted between the two jobs in the same file.

**Recommendation:** Low priority given GitHub Actions' per-job step model makes some duplication idiomatic, but consider a reusable/composite action (e.g. .github/actions/setup-go-env) if a third job is added later to avoid a third copy.

```
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2

      - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5  # v5.5.0
        with:
          go-version-file: go.mod
```

</details>
<details>
<summary><code>QUAL-003</code> · low · <code>.github/workflows/live-tests.yml:45</code></summary>

**Issue:** Both checkout steps (line 45, line 85) rely on actions/checkout's default of persisting the GITHUB_TOKEN in the local git config for the remainder of the job. Since the token is read-only (permissions: contents: read at line 15-16) and no artifact upload path includes the .git directory, actual risk is low, but persisting an unused token in git config is unnecessary attack surface if a later step is added that reads/uploads repo internals.

**Recommendation:** Add `persist-credentials: false` to both actions/checkout steps since no step in this workflow needs to push or authenticate back to GitHub.

```
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
```

</details>
<details>
<summary><code>QUAL-004</code> · nit · <code>.github/workflows/live-tests.yml:100</code></summary>

**Issue:** The 'Setup live targets' step (line 100-101) calls ./test/setup-live-targets.sh, which unconditionally runs `go build` for vespasian internally. The subsequent 'Run live tests' step then passes --no-build to run-live-tests.sh, so the build from setup-live-targets.sh is the only one — functionally correct, but it means the live-tests job cannot share a pre-built binary from the offline-tests job (both jobs perform their own independent go build), and the 'Build vespasian' step naming in offline-tests obscures that live-tests also builds, just implicitly inside the setup script.

**Recommendation:** Optional: build once and pass the binary between jobs via actions/upload-artifact + download-artifact if build time becomes a bottleneck. Not required for correctness.

```
      - name: Setup live targets
        run: ./test/setup-live-targets.sh
```

</details>

</details>
<details>
<summary>Tests (1)</summary>

<details>
<summary><code>TEST-002</code> · low · <code>pkg/crawl/browser_test.go:62</code></summary>

**Issue:** The production check is `os.Getenv("CI") != ""` (browser.go:65), which auto-disables the Chrome sandbox for ANY non-empty CI value — including CI=false or CI=0. The table only exercises the empty string ("") and "true", so the security-relevant boundary between empty and non-empty is never pinned. The surprising behavior that CI=false still disables the sandbox (raised by the Gemini reviewer) ships with no test either confirming it is intended or guarding against it. Whatever the final prod decision, the test suite should pin behavior at this boundary.

**Recommendation:** Add a case such as {"CI set to false still auto-detects", false, "false", <expected>} so the boundary is explicit. If prod is fixed to treat falsy values as disabled, this case pins the corrected behavior; if left as-is, it documents the known behavior. Note t.Setenv("CI", "") sets an empty (present) value, which is fine because the check keys on emptiness.

```
		{"no flags no CI", false, "", false},
		{"CI auto-detects NoSandbox", false, "true", true},
```

</details>

</details>
<details>

<details>
<summary><code>DOC-001</code> · low · <code>pkg/crawl/browser.go:24</code></summary>

**Issue:** The PR changes NewBrowserManager's observable behavior: the Chrome OS sandbox is now silently disabled whenever os.Getenv("CI") != "", independent of BrowserOptions.NoSandbox. The BrowserOptions.NoSandbox godoc still reads as if the sandbox is only disabled when the caller sets NoSandbox=true, and neither NewBrowserManager nor configureLauncher documents the CI auto-detection. Callers (including pkg/sdk) reading the doc comment cannot discover this security-relevant behavior.

**Recommendation:** Document the CI auto-detection on BrowserOptions.NoSandbox and/or NewBrowserManager: state that a non-empty CI env var forces NoSandbox, and (once addressed) prefer an explicit project-owned opt-in.

```
	if opts.NoSandbox || os.Getenv("CI") != "" {
		l = l.NoSandbox(true)
```

</details>

</details>

</details>

---

santiago-praetorian (changes requested):
**REQUEST CHANGES** — Praetorian Capability PR Review (head 0fa4de0).

8 blockers: 1 critical, 2 high, 5 medium. Most important: the ambient `CI` env var silently disables Chrome's OS sandbox in `pkg/crawl` (CVSS 9.0), which `pkg/sdk` ships to chariot/Guard production hosts that crawl untrusted content; and the live-tests teardown deletes `test/.results` before the artifact upload, wiping diagnostics exactly on failing runs. AC5 (reconcile `test/live-test-gaps.md`) is unaddressed.

Full findings and the note on the CodeRabbit / GitHub Actions bot reviews are in the review comment above. Those bot findings also need to be addressed.

santiago-praetorian:

**PR:** https://github.com/praetorian-inc/vespasian/pull/164
**Linear:** https://linear.app/praetorianlabs/issue/LAB-4012/vespasian-run-the-live-test-suite-in-github-actions-ci
**Head SHA:** 7befe3e495395b34aab863b8f1002b142ed27fc5
**Reviewed at:** 2026-07-09

**Recommendation:** COMMENT

|---|---|---|---|---|---|
| Count | 0 | 0 | 0 | 5 | 0 |

> All 5 acceptance criteria for LAB-4012 are met, and every prior automated-review finding (Codex, Gemini, CodeRabbit) is either addressed at HEAD or was invalid. No blockers — the items below are optional low-severity polish.

<details>

<details>
<summary>Security (2)</summary>

<details>
<summary><code>SEC-BE-001</code> · low · <code>.github/workflows/live-tests.yml:22</code></summary>

**Issue:** All jobs use `runs-on: ubuntu-latest`. The label floats across GitHub runner image releases, so the OS/toolchain underneath the tests can change between runs, hurting reproducibility. (Actions themselves are correctly SHA-pinned — good.) CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:L/A:N (3.7, low).

**Recommendation:** Optionally pin to a specific image (e.g. `ubuntu-24.04`) for reproducible CI. Low priority for a test workflow.

```diff
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```

</details>

<details>
<summary><code>SEC-BE-002</code> · low · <code>.github/workflows/live-tests.yml:88</code></summary>

**Issue:** The `live-tests` job installs Chrome from a third-party action and runs a crawler with unrestricted network egress. Blast radius is limited (local-only test targets, SHA-pinned actions, `contents: read` token), so this is defense-in-depth only. CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:L/I:N/A:N (3.7, low).

**Recommendation:** Optional — add a `step-security/harden-runner` egress policy if the org standardizes on it. Non-blocking.

</details>

</details>

<details>

<details>
<summary><code>QUAL-001</code> · low · <code>.github/workflows/live-tests.yml:46</code></summary>

**Issue:** The `offline-tests` and `live-tests` jobs repeat identical checkout (`persist-credentials: false`), `setup-go` (same pinned SHA + `go-version-file`), and `upload-artifact` blocks. Divergence risk: a future security bump to one pinned SHA can silently miss the other copy.

**Recommendation:** Acceptable for a 2-job workflow. Consider a composite action / reusable workflow only if a third job is added. Non-blocking.

</details>

</details>

<details>
<summary>Tests (2)</summary>

<details>
<summary><code>TEST-001</code> · low · <code>pkg/crawl/browser.go:85</code></summary>

**Issue:** `NewBrowserManager` now delegates to `configureLauncher` and returns its error before launching Chrome (a new branch in this PR). `configureLauncher`'s invalid-proxy error is unit-tested directly, but no test drives that error through `NewBrowserManager`. Cheap to cover (no Chrome needed), but the underlying validation is already tested — low priority.

**Recommendation:** Optionally add a test calling `NewBrowserManager(BrowserOptions{Proxy: "not a valid proxy"})` asserting a non-nil error and nil manager.

```diff
 l, err := configureLauncher(opts)
 if err != nil {
     return nil, err
```

</details>

<details>
<summary><code>TEST-002</code> · low · <code>pkg/crawl/browser_test.go:91</code></summary>

**Issue:** The proxy and chrome-path subtests assert `l.Has("proxy-server")` / `l.Has("rod-bin")` rather than the value passed through; the invalid-proxy test only asserts `err != nil`, not the message. A regression that set the wrong flag value or reworded the validation error would pass unnoticed.

**Recommendation:** Assert the flag value (e.g. the configured proxy-server address) and use a substring check on the invalid-proxy error message.

```diff
-if !l.Has("proxy-server") {
+if got := l.Get("proxy-server"); got != "http://127.0.0.1:8080" {
```

</details>

</details>

</details>

---

_6 prior findings dismissed as already addressed or invalid — persist-credentials, upload-before-teardown, unlabeled re-run gate, cancel-in-progress on main, ambient-CI→VESPASIAN_NO_SANDBOX opt-in, and the invalid "missing grpc-server target"._

---

santiago-praetorian:

**PR:** https://github.com/praetorian-inc/vespasian/pull/164
**Linear:** https://linear.app/praetorianlabs/issue/LAB-4012/vespasian-run-the-live-test-suite-in-github-actions-ci
**Head SHA:** 0498c5eb951e0c0443b49da768d96f8f1f66116a
**Reviewed at:** 2026-07-09

**Recommendation:** APPROVE

|---|---|---|---|---|---|
| Count | 0 | 0 | 0 | 0 | 1 |

> Re-review at head `0498c5e`. The engineer's latest commit ("pin ubuntu runner, strengthen test assertions") resolved the three actionable findings from the prior review: runners pinned to `ubuntu-24.04` (SEC-BE-001), test assertions now check flag values + error messages (TEST-002), and `NewBrowserManager`'s error-propagation branch is now covered by `TestNewBrowserManager_InvalidProxyReturnsError` (TEST-001). All 5 LAB-4012 acceptance criteria remain met. Only one nit remains.

<details>

<details>

<details>
<summary><code>QUAL-001</code> · nit · <code>.github/workflows/live-tests.yml:52</code></summary>

**Issue:** The offline-tests and live-tests jobs each repeat identical SHA-pinned actions/checkout and actions/setup-go steps (same pins, same persist-credentials/go-version-file config). This is only 2 occurrences (below the Rule-of-Three threshold), so extracting a composite action isn't strictly warranted yet, but it does mean any future SHA bump (Dependabot/security update) must be applied in two places instead of one.

**Recommendation:** Optional: if a third job is ever added, extract the checkout+setup-go steps into a local composite action (.github/actions/setup-go-env) to keep the SHA pins in one place. Not blocking for this PR.

</details>

</details>

</details>

---

_Multiple prior findings dismissed as already addressed, fixed by commit 0498c5e, or invalid — see `.review/addressed.json`._

---

santiago-praetorian (approved):
Approved via Praetorian capability-pr-review at head 0498c5e.

All 5 LAB-4012 acceptance criteria met; the latest commit resolved every actionable finding from the prior review (ubuntu-24.04 pin, NewBrowserManager error-path test, strengthened flag/error assertions). No blockers, no critical/high/medium findings — one nit remains (duplicated checkout+setup-go steps across the two workflow jobs, below Rule-of-Three, non-blocking).

Egress-hardening (SEC-BE-002) is risk-accepted for this PR (limited blast radius: contents:read token, SHA-pinned actions, persist-credentials:false, local-only targets) and tracked as a defense-in-depth follow-up under LAB-41.

PR #167, opened 2026-07-10
ci: remove third-party setup-chrome action (LAB-4747)

eli-wald:
## Summary

- Removes the `browser-actions/setup-chrome` third-party action from the `live-tests` job
- The `ubuntu-24.04` runner already ships `google-chrome-stable` at `/usr/bin/google-chrome-stable`, which go-rod discovers automatically via its `LookPath` search
- Updates CLAUDE.md to reflect that Chrome comes from the runner image
- **Switches live-tests PR gating from opt-in to opt-out** (LAB-4759): live tests now run by default on every PR; add the `skip-live-tests` label to bypass them. Previously the `live-tests` label was required to run them on a PR.

The Chrome change is a supply-chain surface reduction — it removes a third-party dependency while maintaining identical Chrome-provisioning behavior. The gating change (LAB-4759) is a deliberate behavior change: live tests now execute on every PR unless explicitly skipped.

Depends on #164 (which adds `live-tests.yml` and pins `ubuntu-24.04`).

Closes [LAB-4747](https://linear.app/praetorianlabs/issue/LAB-4747/vespasian-live-tests-ci-blocked-browser-actionssetup-chrome-not-on)
Implements [LAB-4759](https://linear.app/praetorianlabs/issue/LAB-4759/vespasian-switch-live-tests-ci-gating-from-opt-in-to-opt-out-label)

## Test plan

- [x] Verified on eli-wald/vespasian fork: all 5 intentionally-broken test PRs produce identical pass/fail results with and without the `setup-chrome` step
- [x] Live-tests job completes successfully using runner-provided Chrome
- [x] go-rod `LookPath` finds `/usr/bin/google-chrome-stable` (item #7 in its Linux search list)


santiago-praetorian (changes requested):

**Recommendation: REQUEST CHANGES** — 5 blockers, 3 non-blockers.

> **Note on scope — job consolidation is intentional.** Collapsing the previously-separate `offline-tests` and `live-tests` jobs into a single `test` job is a deliberate design choice by the author/team, **not** a finding. Every recommendation below keeps the tests consolidated in one job — none propose reverting to two jobs. The findings are about making that single consolidated job correct (docs, failure ordering, timeout budget, target-list integrity).

Reviewed against **LAB-4747** (Chrome/allowlist). LAB-4747's four acceptance criteria are met (setup-chrome removed, runner Chrome used, tests attested green, CLAUDE.md touched), with the AC4 caveat in QUAL-008.

| Totals | high | medium | low | nit |
|---|---|---|---|---|
| Count | 2 | 3 | 1 | 2 |
| Blockers | 2 | 3 | 0 | 0 |

✅ _Already addressed: the LAB-4759 gating inversion is now disclosed in the PR description and linked to the ticket._

---

<details>
<summary><strong>Blockers (5)</strong></summary>

<details>
<summary><code>QUAL-008</code> · high · <code>CLAUDE.md:135</code></summary>

**Issue:** The CI section still reads "**Two parallel jobs — offline-tests … and live-tests …**", but this PR collapses both into one `test` job. The doc now contradicts the workflow it documents — in the same commit that edited this paragraph — and this is precisely AC4 of LAB-4747 ("the approach is documented in CLAUDE.md's CI section").

**Recommendation:** Rewrite the bullet to describe the single `test` job that runs all 25 targets (offline + live) with runner-provided Chrome and opt-out `skip-live-tests` gating.

```diff
- - **live-tests.yml**: Two parallel jobs — **offline-tests** (...) and **live-tests** (...). Runs on every PR by default; add the `skip-live-tests` label to bypass.
+ - **live-tests.yml**: A single **test** job runs all 25 targets (offline imports/generates/classifier/spec-edge + live crawl-based tests), using the stable Chrome shipped with the ubuntu-24.04 runner. Runs on every PR by default; add the `skip-live-tests` label to bypass. Always runs on push to `main` and `workflow_dispatch`.
```

</details>

<details>
<summary><code>QUAL-005</code> · high · <code>.github/workflows/live-tests.yml:60-64</code></summary>

**Issue:** The `test` job runs `setup-live-targets.sh` (which is `set -euo pipefail` and exits non-zero on a prerequisite/port/service-boot failure) before the `Run tests` step, and `Run tests` has no `if: always()`. So a service-boot or Chrome flake — after the binary has already built — silently skips the 19 deterministic offline targets (imports, generates, classifier-edge, spec-edge) that need no services at all. Those are the tests that catch parameterization regressions per `test/live-test-gaps.md`.

**Recommendation:** Keep the single job, but order it so a live-setup failure can't mask deterministic coverage: build the binary and run the offline/deterministic targets first, then boot services and run the live targets in a later step.

```yaml
      - name: Run offline tests
        run: ./test/run-live-tests.sh --no-start --no-build --targets <offline set>
      - name: Setup live targets
        run: ./test/setup-live-targets.sh
      - name: Run live tests
        run: ./test/run-live-tests.sh --no-build --targets <live set>
```

</details>

<details>
<summary><code>QUAL-004</code> · medium · <code>.github/workflows/live-tests.yml:44</code></summary>

**Issue:** `timeout-minutes: 20` is unchanged, but the single job now does what used to be a 20-minute `live-tests` job **plus** a separate 10-minute `offline-tests` job (previously parallel, each with its own budget). All 25 targets now share one 20-minute window — a built-in timeout-flake risk.

**Recommendation:** Re-evaluate and raise `timeout-minutes` to cover the combined offline + live run under CI load.

```yaml
    timeout-minutes: 20   # now covers all 25 targets in one job — likely too tight
```

</details>

<details>
<summary><code>QUAL-006</code> · medium · <code>.github/workflows/live-tests.yml:64</code></summary>

**Issue:** The PR removed the explicit `--targets` enumerations that used to live in the workflow (19 offline + 6 live) and now calls `./test/run-live-tests.sh --no-build` with no `--targets`, relying on the script's implicit default block (`run-live-tests.sh:2629-2633`). That default and the case-dispatch universe (`run-live-tests.sh:2660-2685`) are two hand-maintained lists with nothing binding them. They match today (25 = 25), but a future target added only to the dispatch would silently never run in CI — and an unknown/typo'd target only logs `Unknown target … (skipping)`, never fails. Target-set changes are also no longer visible in the workflow diff.

**Recommendation:** Restore an explicit `--targets` list in the workflow **or** add a self-check (see TEST-001).

```diff
- run: ./test/run-live-tests.sh --no-build
+ run: ./test/run-live-tests.sh --no-build --targets <explicit 25-target list>
```

</details>

<details>
<summary><code>TEST-001</code> · medium · <code>.github/workflows/live-tests.yml:63</code></summary>

**Issue:** (Same root cause as QUAL-006, coverage angle.) There is no regression guard asserting the script's default target list equals its dispatch universe, so CI coverage can silently drift. This is the concrete mechanism behind "not all possible tests could run" going unnoticed.

**Recommendation:** Add a guard test that fails when the default-target list and the case-dispatch labels diverge, and turn the dispatch `*` branch from `log_warn` into a hard non-zero exit so an unknown/typo'd target fails CI instead of being skipped.

```bash
*)
    log_warn "Unknown target: $target (skipping)"   # → should fail, not warn
    ;;
```

</details>

</details>

<details>
<summary><strong>Non-blockers (3)</strong></summary>

<details>
<summary><code>QUAL-003</code> · low · <code>.github/workflows/live-tests.yml:40</code></summary>

**Issue:** Serializing the previously-parallel offline (~1-2 min) and live jobs into one job means total wall-clock ≈ offline + live, now paid on every PR (opt-out default). Accepted cost of the intended consolidation — noted for awareness, not a request to re-split.

**Recommendation:** If CI turnaround matters, a matrix leg for the offline set would restore parallel fast feedback while staying one workflow. Otherwise no action.

</details>

<details>
<summary><code>QUAL-002</code> · nit · <code>.github/workflows/live-tests.yml:38</code></summary>

**Issue:** One combined job status is less granular in the Checks UI than the old per-category jobs.

**Recommendation:** Name test steps clearly by category so the failing target is still obvious in the log.

</details>

<details>
<summary><code>QUAL-007</code> · nit · <code>.github/workflows/live-tests.yml:70</code></summary>

**Issue:** Uploaded artifact renamed from `offline-test-results` / `live-test-results` to a single `test-results` — a CI-surface contract change with no external references found in-repo.

**Recommendation:** Fine to keep; just be aware if any external tooling downloads by the old artifact names.

</details>

</details>

---

_Review generated by `capability-pr-review` (Praetorian offsec). Recommendations are compatible with keeping the live tests consolidated in one job._

santiago-praetorian (changes requested):

**PR:** https://github.com/praetorian-inc/vespasian/pull/167
**Linear:** [LAB-4747](https://linear.app/praetorianlabs/issue/LAB-4747/vespasian-live-tests-ci-blocked-browser-actionssetup-chrome-not-on) · [LAB-4759](https://linear.app/praetorianlabs/issue/LAB-4759/vespasian-switch-live-tests-ci-gating-from-opt-in-to-opt-out-label)
**Head SHA:** 46b919c6f606dd9d8d99eb5974d459efb43ccb0f
**Reviewed at:** 2026-07-10

**Recommendation:** REQUEST CHANGES — 1 blocker (medium), 2 non-blockers.

> **Re-review.** The prior `capability-pr-review` round raised 5 blockers; the follow-up commits (`address PR review blockers…`, `add stub config…`, `add explicit build step…`) resolved **all 5** — verified against this head SHA: CLAUDE.md now documents the single `test` job (QUAL-008), offline targets run before live setup (QUAL-005), `timeout-minutes` is 30 (QUAL-004), explicit `--targets` lists are restored (QUAL-006), and the unknown-target branch hard-FAILs with a non-zero CI exit (TEST-001). The prior step-naming nit (QUAL-002-old) is also fixed. **Both tickets' acceptance criteria are met** (LAB-4747 AC1–AC4, LAB-4759). The one remaining blocker below is a hardening gap the author can reasonably defer by filing a follow-up ticket.

|---|---|---|---|---|---|
| Count | 0 | 0 | 1 | 2 | 0 |

---

<details>

<details>
<summary><code>QUAL-002</code> · medium · QUAL · <code>.github/workflows/live-tests.yml:74</code></summary>

**Issue:** The workflow's two hand-maintained `--targets` lists (19 offline + 6 live = 25) currently match every case label in `test/run-live-tests.sh`'s dispatch switch (25). Nothing enforces that agreement going forward — the two lists live in different files and are edited independently. The TEST-001 fix (hard-FAIL on an unrecognized target) only catches the *workflow → script* direction (a typo'd/removed target). It does **not** catch the more likely drift: a new case label added to the script's dispatch that is never added to a workflow `--targets` list. That new target then silently never runs in CI — no failure, no warning, no diff-visible symptom, coverage quietly shrinks.

**Note on prior round:** the earlier review offered "restore explicit `--targets` **OR** add a self-check." The author satisfied the OR by restoring explicit `--targets` + hard-FAIL, which is legitimate. This finding is the residual one-directional gap; **filing a follow-up Linear ticket for the guard would correctly declassify this from a blocker** per the review's blocker formula.

**Recommendation:** Add a lightweight guard — e.g. a `--list-targets` mode in `run-live-tests.sh` diffed against the workflow lists in a small CI step, so an added-but-unwired target fails CI loudly instead of being silently skipped. Or track it in a follow-up ticket.

```
# workflow offline (19) + live (6):
--targets import-burp,import-har,import-base64,import-mitmproxy,import-mitmproxy-native,import-unicode,import-duplicates,import-malformed,import-empty,generate-rest,generate-wsdl,generate-wsdl-matrix,generate-graphql,generate-graphql-imports,generate-js-static,generate-merge-slugs,crawl-unreachable,classifier-edge,spec-edge
--targets rest-api,soap-service,graphql-server,concat-spa,edge-cases,crawl-depth
```

</details>

</details>

<details>

<details>

<details>
<summary><code>QUAL-001</code> · low · <code>.github/workflows/live-tests.yml:19</code></summary>

**Issue:** Intended LAB-4759 behavior (opt-in → opt-out), disclosed in the PR description — noted for CI-posture awareness only. Every PR now unconditionally boots REST/SOAP/GraphQL/concat-SPA services and launches headless Chrome, adding the live-suite wall-clock cost (offline + live serialized, up to 30 min) unless the author adds `skip-live-tests`. This is a `pull_request` trigger (not `pull_request_target`) with `contents: read` and `persist-credentials: false`, so **no secrets or elevated permissions are exposed to fork code** — a cost/turnaround consideration, not a security one.

**Recommendation:** No change required. Optionally document in CONTRIBUTING/README that the live suite runs automatically and maintainers can add `skip-live-tests` for draft/WIP PRs.

</details>

<details>
<summary><code>QUAL-003</code> · low · <code>.github/workflows/live-tests.yml:85</code></summary>

**Issue:** The uploaded artifact is renamed from the split names (`offline-test-results` / `live-test-results`) to a single `test-results` as a byproduct of job consolidation — a CI-surface contract change. A repo-wide grep found no other in-repo consumer of these names, so risk is low. (This was prior non-blocker QUAL-007, not part of the addressed set.)

**Recommendation:** Fine to keep. If any downstream automation outside this repo downloads by the old artifact names, update it in the same rollout.

</details>

</details>

</details>

---

_6 prior-round findings dismissed as already addressed — see `.worktrees/capability-pr-review/pr-167/.review/addressed.json`._

---

---

**Update (2026-07-11):** Follow-up ticket **[LAB-4773](https://linear.app/praetorianlabs/issue/LAB-4773/vespasian-derive-ci-live-test-targets-from-run-live-testssh-to-avoid)** has been filed to address the **QUAL-002** target-list drift guard (make `test/run-live-tests.sh` the single source of truth so a newly added target can't be silently skipped in CI). Per this review's blocker formula, tracking the guard in a follow-up **declassifies QUAL-002 from a blocker**. Also linked from the 13T-146 Notes section. Approving.

santiago-praetorian:
**`.github/workflows/live-tests.yml` (L77 offline, L83 live): hardcoded `--targets` lists will silently drift from the runner.**

These two lists together happen to cover all 25 targets in `run-live-tests.sh` today, but nothing keeps them in sync. When someone adds a target to `run-live-tests.sh`, they also have to remember to add it here — and if they don't, CI stays green while silently never running the new target. That's the dangerous kind of failure (quiet, not loud).

Suggest making the runner the single source of truth so a new target is picked up with no workflow edit: define `OFFLINE_TARGETS` / `LIVE_TARGETS` lists in `run-live-tests.sh`, add a `--group offline|live` flag that selects one, and have CI call `--group offline` / `--group live` instead of naming targets. Adding a target then means editing one list in the runner, and drift becomes impossible by construction.

This also mirrors the opt-out default this PR already adopts for the label gate (`skip-live-tests`) — new targets run by default rather than needing to be opted in.

Non-blocking (the pattern predates this PR), but since #167 is restructuring these jobs it's a good moment to fix it.


---

**Update (2026-07-11):** Filed as follow-up ticket **[LAB-4773](https://linear.app/praetorianlabs/issue/LAB-4773/vespasian-derive-ci-live-test-targets-from-run-live-testssh-to-avoid)** (sub-issue of LAB-41) and linked from the 13T-146 Notes section. This tracks the single-source-of-truth fix; approving #167 with the drift guard deferred to that ticket.

santiago-praetorian (approved):
Approving. The one blocker from the prior review (QUAL-002 — CI live-test target-list drift) is now tracked in follow-up ticket LAB-4773 (sub-issue of LAB-41, linked from 13T-146), which declassifies it per the blocker formula. Both tickets' acceptance criteria (LAB-4747, LAB-4759) are met; the two remaining findings are non-blocking. Nice work removing the third-party setup-chrome action and flipping the live-tests gate to opt-out.

PR #168, opened 2026-07-10
fix(grpc): bound aggregate reflection-descriptor memory across targets (LAB-4693)

saullocarvalho:
## Summary

Defense-in-depth hardening for the gRPC reflection probe — a **global cross-target descriptor-memory budget**. Follow-up from PR #159 review finding **SEC-BE-001** (CVSS 5.3 medium; tracked as non-blocking **LAB-4693**).

### The gap
Per-**target** reflection memory was already bounded in #159: `dialGRPC` sets `MaxCallRecvMsgSize` (4 MiB/message), and `walkFileDescriptors` enforces a 64 MiB aggregate serialized-descriptor budget + 1000-file count cap **per target**. But `GRPCProbe.Probe` retains one `*classify.GRPCReflectionResult` per distinct target in `schemasByTarget`, and the fan-out was bounded only by `Config.MaxEndpoints` (default 500). So a capture referencing many distinct, reachable, hostile gRPC `host:port` targets could drive aggregate retained memory to ~`MaxEndpoints × 64 MiB` (~32 GiB) and OOM the scanning host. No global budget existed.

Likelihood is low (needs an unusual capture — hundreds of hostile targets, or attacker-controlled DNS for many names) and impact is confined to the operator's own scanning process, hence this was non-blocking.

### The fix
Add a global budget in `GRPCProbe.Probe`, mirroring the existing per-target loop-break in `runReflection`:
- New `Config.MaxTotalReflectionDescriptorBytes`, default `DefaultMaxTotalReflectionDescriptorBytes = 4 * maxGRPCDescriptorBytes` (**256 MiB**) — a small multiple of the per-target cap.
- `Probe` accumulates each target's retained descriptor bytes (`reflectionRetainedBytes`) and **stops dialing further targets** once the cumulative total reaches the ceiling.
- Ordering is deliberate: the target that *crosses* the threshold keeps and applies its already-retained schema; only **subsequent** targets are skipped.
- `nil` / reflection-unavailable results retain no descriptors, so they never advance the budget.

Worst-case retained memory is bounded to ~256 MiB (plus at most one in-flight target's ≤64 MiB), versus the unbounded ~32 GiB before. The default is far above any real scan (typical targets retain KB), so normal behavior is unchanged.

## Test plan
- [x] `TestGRPCProbe_Probe_AggregateDescriptorBudgetStopsTargets` — two loopback reflection servers, `MaxTotalReflectionDescriptorBytes: 1`, counting dialer; asserts the first target is retained (`ReflectionEnabled`) while the **second is never dialed** (dial count == 1). **Mutation-verified**: removing the break makes the dial count 1→2 and the test fails.
- [x] `go test -race ./...` — full suite green (18 packages).
- [x] `go vet` / `gofmt` clean; `golangci-lint run pkg/probe/...` — 0 issues.

Refs [LAB-4693](https://linear.app/praetorianlabs/issue/LAB-4693/grpc-bound-aggregate-reflection-descriptor-memory-across-targets-in).

### DIFF
--- a/pkg/probe/grpc.go
+++ b/pkg/probe/grpc.go
@@ -86,6 +86,7 @@ func (p *GRPCProbe) Probe(ctx context.Context, endpoints []classify.ClassifiedRe
 	// reflection probe result.
 	schemasByTarget := make(map[grpcTargetInfo]*classify.GRPCReflectionResult)
 	seen := make(map[grpcTargetInfo]bool)
+	totalRetained := 0
 
 	for _, ep := range endpoints {
 		if ep.APIType != "grpc" {
@@ -104,7 +105,14 @@ func (p *GRPCProbe) Probe(ctx context.Context, endpoints []classify.ClassifiedRe
 		}
 		seen[target] = true
 
-		schemasByTarget[target] = p.probeTarget(ctx, target)
+		res := p.probeTarget(ctx, target)
+		schemasByTarget[target] = res
+		totalRetained += reflectionRetainedBytes(res)
+		if totalRetained >= p.config.MaxTotalReflectionDescriptorBytes {
+			slog.DebugContext(ctx, "grpc probe: aggregate descriptor budget exhausted; stopping target enumeration",
+				"targets_probed", len(seen), "retained_bytes", totalRetained)
+			break
+		}
 	}
 
 	// Copy endpoints to avoid mutating the caller's slice.
@@ -360,4 +368,20 @@ func extractService(fd *desc.FileDescriptor, fqn string) (classify.GRPCService, 
 		return gs, true
 	}
 	return classify.GRPCService{}, false
+}
+
+// reflectionRetainedBytes returns the total serialized descriptor bytes a
+// reflection result retains (0 for a nil result or a result without
+// descriptors — e.g. reflection-unavailable or unreachable targets, which
+// therefore never advance the aggregate budget). Used by Probe to enforce
+// Config.MaxTotalReflectionDescriptorBytes across targets.
+func reflectionRetainedBytes(r *classify.GRPCReflectionResult) int {
+	if r == nil {
+		return 0
+	}
+	total := 0
+	for _, b := range r.FileDescriptors {
+		total += len(b)
+	}
+	return total
 }
--- a/pkg/probe/grpc_test.go
+++ b/pkg/probe/grpc_test.go
@@ -652,6 +652,67 @@ func TestGRPCProbe_Probe_AggregateDescriptorBudgetStopsTargets(t *testing.T) {
 	assert.Equal(t, int32(1), dialCount.Load(), "exactly one target (server A) should be dialed; server B must never be dialed")
 }
 
+// TestGRPCProbe_Probe_ZeroByteTargetDoesNotAdvanceBudget pins the complementary
+// invariant to TestGRPCProbe_Probe_AggregateDescriptorBudgetStopsTargets: a nil
+// / reflection-unavailable / unreachable target retains 0 descriptor bytes
+// (reflectionRetainedBytes(nil) == 0) and therefore must NOT advance
+// totalRetained in Probe. The first endpoint targets an unreachable (closed)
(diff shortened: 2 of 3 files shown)


santiago-praetorian:
## Praetorian Capability PR Review + Test Verification

**Head SHA:** `d155f09` · **Linear:** [LAB-4693](https://linear.app/praetorianlabs/issue/LAB-4693/grpc-bound-aggregate-reflection-descriptor-memory-across-targets-in)
**Recommendation:** ✅ **APPROVE** — 0 blockers

Multi-agent review (capability-reviewer / backend-security / test-lead) plus a ticket-alignment pass, followed by static, dynamic, and live test execution against the PR head.

### Review findings

|---|---|---|---|---|---|
| Count | 0 | 0 | 0 | 1 | 1 |

- **Security (SEC-BE): 0 findings.** Budget logic verified correct — `totalRetained` accumulates per unique target, the `>=` break fires *after* storing the crossing target's schema, no integer overflow (~320 MiB peak ≪ int max), and `reflectionRetainedBytes(nil) == 0` so unreachable/reflection-unavailable targets don't advance the budget. Aggregate worst case drops from ~32 GiB → ~256 MiB.
- **Requirements (REQ): all 5 LAB-4693 acceptance criteria met.**
- **Docs:** adequate — both new exported symbols carry thorough godoc; consistent with the repo's godoc-only convention for `pkg/probe` Config tuning fields.

<details>
<summary>Non-blocking findings (2) — optional, safe to defer</summary>

- **QUAL-001 · nit · `pkg/probe/grpc.go:378`** — `reflectionRetainedBytes` re-sums a total `runReflection` already computes and discards. Negligible cost (map capped at `MaxReflectionDescriptors`, default 1000). Optional: stash the already-computed `totalBytes` on the result and reuse it.
- **TEST-001 · low · `pkg/probe/grpc_test.go:666`** — `TestGRPCProbe_Probe_ZeroByteTargetDoesNotAdvanceBudget` only exercises the unreachable→`nil` path; the non-nil reflection-unavailable branch (empty `FileDescriptors` map → 0 bytes, `grpc.go:382-385`) — the exact shape the comment/PR body call out — is not driven through the aggregate budget. Optional: add a direct table test for `reflectionRetainedBytes` (nil→0, non-nil-empty→0, populated→sum).

</details>

### Test verification (against head `d155f09`)

**Static — ✅ PASS**
- `gofmt -l pkg/probe` clean · `go vet ./...` exit 0 · `go build ./...` exit 0 · `golangci-lint run pkg/probe/...` **0 issues**

**Dynamic — ✅ PASS (PR scope)**
- `pkg/probe` green under `-race`; both new tests pass:
  - `TestGRPCProbe_Probe_AggregateDescriptorBudgetStopsTargets` — PASS
  - `TestGRPCProbe_Probe_ZeroByteTargetDoesNotAdvanceBudget` — PASS
- Full `go test -race ./...`: every package green **except** `pkg/crawl/TestConfigureLauncher/sandbox/*`, which is a **pre-existing, container-only artifact unrelated to this PR** — `pkg/crawl` is byte-identical to `main`, and go-rod's `launcher.New()` force-adds `--no-sandbox` when running in a container (`launcher.go:105-106`), so the "flag absent" assertion can't hold in a devcontainer. GitHub's `ci / Test` check is green, confirming the suite passes in canonical CI.

**Live — ✅ PASS**
- `./test/run-live-tests.sh` against real REST/SOAP/GraphQL/gRPC/SPA targets: **26 passed, 0 failed, 0 skipped**, including **`grpc-server` 3/3** (the reflection path this PR modifies).

---
_Automated review + verification via `/capability-pr-review`. Reproducible against the same head SHA + ticket state._

santiago-praetorian (approved):
Approved — 0 blockers from `/capability-pr-review` (0 security findings, all 5 LAB-4693 acceptance criteria met; 2 optional non-blocking findings). Static, dynamic, and live tests pass against head d155f09: lint 0 issues, new aggregate-budget tests green under -race, live suite 26/26 (grpc-server 3/3). See detailed review comment above.

PR #169, opened 2026-07-10
test(grpc): add CLI-boundary coverage for --grpc-insecure-skip-verify (LAB-4694)

saullocarvalho:
## Summary

Closes the CLI-boundary coverage gap for the `--grpc-insecure-skip-verify` flag, and fully satisfies **LAB-4694 AC3**. Follow-up from PR #159 review finding **TEST-001** (low; non-blocking **LAB-4694**).

Started as a test-only change; a small **behavior-neutral production seam** was added (per PR review) so a *dropped* flag assignment is caught at the CLI boundary — see "Production seam (AC3)" below.

### The gap
The flag is declared on both `GenerateCmd` and `ScanCmd` and copied into `pipeline.Options` / `pipeline.ScanOptions` inside each `Run()`. But `cmd/vespasian/main_test.go` had **zero** references to `GRPCInsecureSkipVerify` — the kong-struct wiring was uncovered at the command boundary. Worse, the `Options` structs were built *inline* inside `Run()`, so a dropped `GRPCInsecureSkipVerify: c.GRPCInsecureSkipVerify` assignment would leave the whole suite green (the AC3 concern).

### Production seam (AC3) — behavior-neutral
Extracted the inline options construction into methods that are now the **sole** construction site of each struct, so a boundary test can inspect flag propagation without executing `Run()`:
- `GenerateCmd.options() pipeline.Options` — pure, from `c` fields.
- `ScanCmd.scanOptions(apiType, afterWSDL) pipeline.ScanOptions` — takes the runtime-resolved `apiType` and the `AfterWSDL` closure; the `c`-derived fields (incl. `GRPCInsecureSkipVerify`) live in one inspectable place. The closure still captures `bs.opts.Headers`/`c`/`c.URL` as before.

Both `Run()` methods now call these. The extraction is field-for-field identical to the previous inline literals — runtime behavior, validation, and SSRF-warning ordering are unchanged (verified: pure struct construction, no I/O).

### Tests added
- **`TestGRPCInsecureSkipVerify_ReachesOptions`** *(AC3)* — asserts the flag reaches `pipeline.Options` **and** `pipeline.ScanOptions` in both directions. **Mutation-verified**: deleting `GRPCInsecureSkipVerify: c.GRPCInsecureSkipVerify` from either `options()`/`scanOptions()` turns it red. This catches the dropped-assignment regression AC3 asked for.
- **`TestGRPCInsecureSkipVerify_Embedded`** — asserts both `GenerateCmd` and `ScanCmd` expose the field (settable `true`, defaults `false`); catches a field removal/rename. Mirrors `TestSlugOptions_Embedded` / `TestDangerousAllowPrivate_ScanCmd`.
- **`TestGRPCInsecureSkipVerify_GenerateCmd`** — writes a REST capture to a temp dir (spec → temp file, not stdout) and runs `GenerateCmd.Run()` with the flag set and `Probe: false` (hermetic — no network), asserting the flag is accepted through the real `Run()` path. Mirrors `TestDangerousAllowPrivate_GenerateCmd`.

The deeper semantic behavior (skip=true vs verify-by-default) remains covered one layer down by `internal/pipeline/pipeline_test.go:TestClassifyProbeGenerate_GRPCInsecureSkipVerify`.

## Test plan
- [x] `go test -race ./cmd/vespasian/...` — pass, incl. the three new tests.
- [x] `go test -race ./...` — full suite green (18 packages).
- [x] `go vet` / `gofmt` clean; `golangci-lint run cmd/vespasian/...` — 0 issues.
- [x] Mutation check: dropping the flag assignment from either `options()`/`scanOptions()` fails `TestGRPCInsecureSkipVerify_ReachesOptions`.

Refs [LAB-4694](https://linear.app/praetorianlabs/issue/LAB-4694/grpc-add-cli-boundary-test-for-grpc-insecure-skip-verify-flag-wiring).

### DIFF
--- a/cmd/vespasian/main.go
+++ b/cmd/vespasian/main.go
@@ -481,6 +481,23 @@ type GenerateCmd struct {
 // maxCaptureSize is the maximum capture file size (100MB).
 const maxCaptureSize = 100 * 1024 * 1024
 
+// options builds the pipeline.Options for this command from its flags. Extracted
+// so a CLI-boundary test can assert each flag reaches pipeline.Options (catching a
+// dropped assignment) without executing Run().
+func (c *GenerateCmd) options() pipeline.Options {
+	return pipeline.Options{
+		APIType:                c.APIType,
+		Confidence:             c.Confidence,
+		Probe:                  c.Probe,
+		Deduplicate:            c.Deduplicate,
+		AllowPrivate:           c.DangerousAllowPrivate,
+		GRPCInsecureSkipVerify: c.GRPCInsecureSkipVerify,
+		MergeSlugs:             c.MergeSlugs,
+		SlugThreshold:          c.SlugThreshold,
+		Status:                 statusWriter(c.Verbose),
+	}
+}
+
 // Run executes the generate command.
 func (c *GenerateCmd) Run() (err error) {
 	if err := validateSlugThreshold(c.APIType, c.MergeSlugs, c.SlugThreshold); err != nil {
@@ -532,17 +549,7 @@ func (c *GenerateCmd) Run() (err error) {
 
 	warnSSRFDisabled(c.DangerousAllowPrivate, c.Probe)
 
-	spec, err := pipeline.ClassifyProbeGenerate(ctx, requests, pipeline.Options{
-		APIType:                c.APIType,
-		Confidence:             c.Confidence,
-		Probe:                  c.Probe,
-		Deduplicate:            c.Deduplicate,
-		AllowPrivate:           c.DangerousAllowPrivate,
-		GRPCInsecureSkipVerify: c.GRPCInsecureSkipVerify,
-		MergeSlugs:             c.MergeSlugs,
-		SlugThreshold:          c.SlugThreshold,
-		Status:                 statusWriter(c.Verbose),
-	})
+	spec, err := pipeline.ClassifyProbeGenerate(ctx, requests, c.options())
 	if err != nil {
 		return err
 	}
@@ -565,6 +572,28 @@ type ScanCmd struct {
 
 	CrawlOptions
 	SlugOptions
+}
+
+// scanOptions builds the pipeline.ScanOptions for this command. The c-derived
+// fields (incl. GRPCInsecureSkipVerify) are collected here so a CLI-boundary test
+// can assert each flag reaches pipeline.ScanOptions (catching a dropped assignment)
+// without executing Run(). The two runtime-derived inputs — the resolved apiType
+// and the AfterWSDL closure (which captures bs.opts.Headers/c/c.URL) — are passed
+// in by Run().
+func (c *ScanCmd) scanOptions(apiType string, afterWSDL func(ctx context.Context, reqs []crawl.ObservedRequest) []crawl.ObservedRequest) pipeline.ScanOptions {
+	return pipeline.ScanOptions{
+		TargetURL:              c.URL,
+		APIType:                apiType,
+		Confidence:             c.Confidence,
+		Probe:                  c.Probe,
+		Deduplicate:            c.Deduplicate,
+		AllowPrivate:           c.DangerousAllowPrivate,
+		GRPCInsecureSkipVerify: c.GRPCInsecureSkipVerify,
+		MergeSlugs:             c.MergeSlugs,
+		SlugThreshold:          c.SlugThreshold,
+		Status:                 statusWriter(c.Verbose),
+		AfterWSDL:              afterWSDL,
+	}
 }
 
 // Run executes the scan command (crawl + generate pipeline).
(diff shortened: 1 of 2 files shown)


saullocarvalho:
Addressed in `930d824`. Both @coderabbitai and Gemini flagged that `TestGRPCInsecureSkipVerify_GenerateCmd` used `Probe: true` on a REST capture pointing at `https://example.com/api/users`, which made `GenerateCmd.Run()` fire live HTTP probes (OPTIONS) against example.com during the test.

Fix: switched that test to **`Probe: false`**. `GRPCInsecureSkipVerify` is consumed only by the gRPC reflection probe, which a REST capture never reaches — so `Probe: true` added zero coverage of the flag while introducing the live external call. With `Probe: false` the smoke test stays hermetic (no network) and still exercises the full `Run()` path where the flag is assigned into `pipeline.Options` (that assignment runs regardless of `Probe`). The test now completes in ~0.00s with no probe attempt.

Note: the sibling `TestDangerousAllowPrivate_GenerateCmd` keeps `Probe: true` intentionally — that flag's behavior *is* about probing, so the asymmetry is correct.

santiago-praetorian (approved):
## Approve (with a comment)

Test-only change, no production code touched — zero runtime regression risk. `go test`/lint are green and the two prior-round issues (live `Probe:true` network call and stdout pollution) are already fixed in-branch. No blockers; the remaining notes are advisory (2 low, 1 nit).

One thing worth making a conscious decision on before closing **LAB-4694**, not a merge blocker:

- **AC3 is not actually satisfied.** The ticket asked for a test that catches a *dropped `c.GRPCInsecureSkipVerify → pipeline.Options` assignment at the CLI boundary* (`main.go:541` / `:669`). Neither new test does that: `TestGRPCInsecureSkipVerify_Embedded` sets a struct field and asserts it (a Go-language guarantee, not the kong flag wiring), and `TestGRPCInsecureSkipVerify_GenerateCmd` runs `Run()` with `Probe:false` on a REST capture and asserts only no-error — a path that never consumes the flag. `internal/pipeline`'s `TestClassifyProbeGenerate_GRPCInsecureSkipVerify` builds `pipeline.Options` directly, bypassing `main.go`. Net: deleting the boundary assignment would leave the suite green.

The PR is candid about this and correctly scopes the real fix (a testable `options()` seam) out of a test-only ticket, so I'm fine merging as-is. Just don't close LAB-4694 as "AC3 done" — either accept the documented residual gap explicitly, or open a small follow-up for the `options()` seam so a dropped assignment is caught at the boundary.

Nice work on the honest scoping in the test doc comments.

saullocarvalho:
@santiago-praetorian thanks — and good call on AC3. Rather than accept the residual gap, I closed it in `3da9764` with the `options()` seam you suggested.

**Production (behavior-neutral):** the inline `pipeline.Options` / `pipeline.ScanOptions` literals are extracted into methods that are now the *sole* construction site:
- `GenerateCmd.options() pipeline.Options` — pure, from `c` fields.
- `ScanCmd.scanOptions(apiType, afterWSDL) pipeline.ScanOptions` — takes the runtime-resolved `apiType` and the `AfterWSDL` closure; the `c`-derived fields (incl. `GRPCInsecureSkipVerify`) live in one inspectable place. The closure still captures `bs.opts.Headers`/`c`/`c.URL` exactly as before.

Both `Run()` methods just call these now — pure struct construction, no I/O, runtime behavior unchanged.

**Test:** `TestGRPCInsecureSkipVerify_ReachesOptions` asserts the flag reaches `pipeline.Options` and `pipeline.ScanOptions` in both directions. This is the AC3 coverage — **mutation-verified**: deleting `GRPCInsecureSkipVerify: c.GRPCInsecureSkipVerify` from either `options()`/`scanOptions()` turns it RED (confirmed on an isolated copy for both seams). The prior field-exposure and hermetic-`Run()` smoke tests are retained.

`gofmt`/`vet`/`golangci-lint` clean, `go test -race ./...` green (18 pkgs). Heads-up: this PR is **no longer test-only** — it now carries the small production refactor — so I'm re-requesting your review rather than riding the prior approval. LAB-4694 can now be closed with AC3 genuinely satisfied.

santiago-praetorian (approved):
## Approve ✅ — superseding my earlier review

Re-reviewed at head `3da9764`. **My previous approval comment is now out of date and this supersedes it** — that comment was written against an earlier commit and two of its statements no longer hold:

- ~~"Test-only change, no production code touched"~~ — the final commit (`3da9764`) **does** touch production code: it extracts the inline `pipeline.Options` / `pipeline.ScanOptions` construction into `GenerateCmd.options()` and `ScanCmd.scanOptions()`, with both `Run()` methods refactored to call them.
- ~~"AC3 is not actually satisfied"~~ — AC3 **is** now satisfied. `TestGRPCInsecureSkipVerify_ReachesOptions` asserts the flag survives `options()`/`scanOptions()`, and since `Run()` builds the options *only* via those methods (main.go:552 / :700), deleting the boundary assignment turns the suite red. This is a live seam, not the dead one I'd flagged earlier.

The refactor is behavior-preserving — the extracted methods are field-for-field identical to the previous inline literals (`GRPCInsecureSkipVerify`, `AllowPrivate`, and the `AfterWSDL` closure all map the same way), and validation/SSRF-warning ordering in `Run()` is unchanged. `go test`/vet/lint green. No blockers.

**LAB-4694 can be closed as AC3-done** (not as a documented residual gap).

Two non-blocking notes, neither needs action to merge:
- `TestGRPCInsecureSkipVerify_ReachesOptions` pins the `options()`/`scanOptions()` method boundary, not the `Run()`→method call edge; a future re-inline of the options construction into `Run()` would bypass it. Fully covered one layer down by `internal/pipeline`'s `TestClassifyProbeGenerate_GRPCInsecureSkipVerify`.
- `TestGRPCInsecureSkipVerify_Embedded` is redundant with `ReachesOptions` + the compiler, but it satisfies the AC1/AC2 field-exposure ask and follows the `TestDangerousAllowPrivate_*` convention — fine to keep.

@saullocarvalho — one housekeeping ask (not a merge blocker): the PR description still says **"Test-only — no production changes"** and describes the `options()` seam as *"intentionally out of scope for this test-only ticket."* Both are now stale — this commit added exactly that seam. Mind updating the Summary / "Honest scope" section so the description matches the diff? Thanks for the clean, well-documented tests.

saullocarvalho:
@santiago-praetorian thanks for the thorough re-review and for explicitly superseding the earlier one — appreciate you calling out the two now-stale statements precisely.

Done: **updated the PR description** to match the diff — dropped the "Test-only — no production changes" line and the "intentionally out of scope for this test-only ticket" note, and added a **"Production seam (AC3) — behavior-neutral"** section documenting `options()`/`scanOptions()` as the sole construction site, plus the `ReachesOptions` mutation-verified boundary test. Summary now reflects that AC3 is satisfied rather than a residual gap.

On your two non-blocking notes:
- Agreed the boundary test pins the `options()`/`scanOptions()` method boundary, not the `Run()`→method call edge. A future re-inline into `Run()` would bypass it — but that's exactly the kind of change your eye (and the `internal/pipeline` end-to-end coverage) would catch, so I've left it as-is rather than over-engineer.
- Kept `TestGRPCInsecureSkipVerify_Embedded` for the AC1/AC2 field-exposure ask and `TestDangerousAllowPrivate_*` convention, as you suggested.

I'll close **LAB-4694 as AC3-done** once this merges (keeping it In Progress until then). Thanks again.

PR #176, opened 2026-07-13
refactor: derive CI live-test targets from runner groups (LAB-4773)

eli-wald:
## Summary

- Define `OFFLINE_TARGETS` and `LIVE_TARGETS` arrays in `test/run-live-tests.sh` as the single source of truth for which targets belong to each CI phase
- Add `--group offline|live|all` flag to select a predefined target group
- Replace hardcoded `--targets` lists in `.github/workflows/live-tests.yml` with `--group offline` and `--group live`
- Adding a new target now means editing one array in the runner — the workflow needs no change, and drift becomes impossible by construction

## Acceptance criteria

- [x] `live-tests.yml` no longer contains a hardcoded per-target list — CI names a group, not individual targets
- [x] Adding a target to `run-live-tests.sh` causes it to run in CI with no change to the workflow
- [x] All 25 existing targets still run, with offline targets executing before live-service setup (current ordering preserved)

## Test plan

- [x] `bash -n test/run-live-tests.sh` — syntax OK
- [x] `--group offline` produces exactly 19 targets matching the old workflow offline list
- [x] `--group live` produces exactly 6 targets matching the old workflow live list
- [x] Default `--group all` produces all 25 targets with `edge-cases` and `crawl-depth` always included
- [x] `TARGETS_SETUP` config override still works (e.g. adds `grpc-server`) with deduplication
- [x] `--targets` still works and takes precedence over `--group`

Closes [LAB-4773](https://linear.app/praetorianlabs/issue/LAB-4773/vespasian-derive-ci-live-test-targets-from-run-live-testssh-to-avoid)


santiago-praetorian (changes requested):

**PR:** https://github.com/praetorian-inc/vespasian/pull/176
**Linear:** https://linear.app/praetorianlabs/issue/LAB-4773/vespasian-derive-ci-live-test-targets-from-run-live-testssh-to-avoid
**Head SHA:** 12bb9c9dd8ad2b235d5daa368079c8a486083f3f
**Reviewed at:** 2026-07-14T00:00:00Z

|---|---|---|---|---|---|
| Count | 0 | 0 | 4 | 2 | 2 |
| Blockers | 0 | 0 | 4 | 1 | 0 |

---

<details>
<summary><strong>Blockers (5)</strong></summary>

<details>
<summary><code>QUAL-002</code> · medium · QUAL · <code>CLAUDE.md:140</code></summary>

**Issue:** This PR adds a new CI gate — the 'Drift guard (target groups vs dispatch)' step in .github/workflows/live-tests.yml (line 64-65), which runs test/test-runner-args.sh and fails the build if a case-dispatch target is missing from OFFLINE_TARGETS/LIVE_TARGETS/CONFIG_ONLY. This is a meaningful, user-facing contract change (rubric item 6: contract changes need a matching doc entry) — a contributor who adds a new target to the dispatch block without updating a group array will now get a CI failure they can't explain from CLAUDE.md alone, since the CI section (and the job-level comment in live-tests.yml lines 38-40) describes only the --group offline/--group live invocation and says nothing about the drift-guard/consistency check.

**Recommendation:** Add a clause to the live-tests.yml CI bullet in CLAUDE.md (and/or the job comment) noting that test/test-runner-args.sh runs as a CI step that fails fast if a dispatch target isn't reflected in OFFLINE_TARGETS/LIVE_TARGETS, e.g. '...a drift-guard step (test/test-runner-args.sh) fails CI if a dispatch target is missing from these arrays.'

```diff
- **live-tests.yml**: A single **test** job runs all targets via `test/run-live-tests.sh --group offline` then `--group live`. Target groups (`OFFLINE_TARGETS` / `LIVE_TARGETS`) are defined in `test/run-live-tests.sh` — the workflow references groups, not individual targets, so adding a target means editing one array in the runner. Uses the stable Chrome shipped with the ubuntu-24.04 runner. Runs on every PR by default; add the `skip-live-tests` label to bypass. Always runs on push to `main` and `workflow_dispatch`.
```

</details>

<details>
<summary><code>TEST-001</code> · low · TEST · <code>test/run-live-tests.sh:2725</code></summary>

**Issue:** smoke-check is a new target/behavior added by this PR. Its PASS branch is exercised only incidentally (the override test happens to run it against a working binary), and its FAIL branch (binary not functional -> set_test_result FAIL) is never exercised. There is no test asserting smoke-check reports FAIL when the binary is broken/absent, so the one thing this target exists to detect is itself unverified.

**Recommendation:** Add a small test that points VESPASIAN at a non-functional/absent binary and asserts test_smoke_check records a FAIL result, complementing the incidental PASS-path coverage.

```diff
    if "$VESPASIAN" version >/dev/null 2>&1 || "$VESPASIAN" --help >/dev/null 2>&1; then
        set_test_result "smoke-check" "FAIL" "-" "-" "$duration"
```

</details>

<details>
<summary><code>TEST-002</code> · medium · TEST · <code>test/run-live-tests.sh:2868</code></summary>

**Issue:** The runner's group-resolution logic (the `case "${group:-all}"` block in main) is never exercised by a test. test-runner-args.sh re-implements the same logic itself (it sources the OFFLINE_TARGETS/LIVE_TARGETS arrays and rebuilds `all` with its own `join_targets` call at lines 71-73, and re-implements the TARGETS_SETUP dedup pipeline at lines 81-83) instead of running the runner with `--group offline|live|all` and asserting the resolved output. This is the classic simulation anti-pattern: if the production `offline)` branch were changed to point at LIVE_TARGETS, or the `all)` concatenation/dedup broke, every assertion in test-runner-args.sh would still pass. The only place the runner's actual `--group` path is invoked (line 155) passes `--targets` too, so the group branch is bypassed by the precedence rule. There is also no assertion pinning the exact membership or count of the offline (20) or live (6) groups, so the PR/AC claims about which targets each group contains are unverified (and the PR body's "exactly 19 offline / all 25" counts are now stale after smoke-check made it 20/26).

**Recommendation:** Add tests that invoke the runner itself (e.g. `source RUNNER --group offline --no-build --no-start` under a stub CONFIG_FILE, as the override test already does) and assert the resolved target set equals the expected offline / live / all membership. Assert exact content/count, not just absence of duplicates, so the group `case` branches are covered and cannot silently drift.

```diff
all="$(join_targets "${OFFLINE_TARGETS[@]}"),$(join_targets "${LIVE_TARGETS[@]}")"
dup_count=$(echo "$all" | tr ',' '\n' | sort | uniq -d | wc -l | tr -d ' ')
            all)
                targets="$(join_targets "${OFFLINE_TARGETS[@]}"),$(join_targets "${LIVE_TARGETS[@]}")"
```

</details>

<details>
<summary><code>TEST-003</code> · medium · TEST · <code>test/test-runner-args.sh:36</code></summary>

**Issue:** The drift guard is the core mechanism satisfying Linear AC#2 ("adding a target causes it to run in CI with no workflow change") — it asserts every group member has a dispatch entry and every dispatch target is in a group or CONFIG_ONLY. But nothing verifies the guard actually catches drift: there is no fixture/negative case where a dispatch target is intentionally omitted from both groups (and from CONFIG_ONLY) and the guard is asserted to report FAIL / exit non-zero. If the extraction sed at line 25 silently produced an empty DISPATCH_TARGETS array, or a comparison were inverted, the guard would pass vacuously and the "drift becomes impossible by construction" guarantee would be false while still reporting green. A safety check with no test that it trips is unverified.

**Recommendation:** Add a self-test that feeds the guard a synthetic runner (or an in-memory array pair) containing a dispatch target absent from all groups and asserts the guard emits the expected FAIL message and non-zero status. This proves the guard detects the exact drift condition AC#2 depends on.

```diff
# Every dispatch target must be in a group or in CONFIG_ONLY.
        fail "Dispatch target '$target' is not in OFFLINE_TARGETS, LIVE_TARGETS, or CONFIG_ONLY"
```

</details>

<details>
<summary><code>TEST-005</code> · medium · TEST · <code>test/test-runner-args.sh:152</code></summary>

**Issue:** The file header states "Does NOT run actual live tests — only validates that the group arrays stay in sync" (lines 2-3), but the override test sources the full runner and runs main(), which reaches the binary-existence guard (run-live-tests.sh:2907 `exit 1` if `bin/vespasian` is missing) and then executes `"$VESPASIAN" version` inside test_smoke_check, and creates the RESULTS_DIR in the repo. This makes a supposedly static drift guard depend on prior `go build` state and on executing the product binary — precisely why the PR had to move the CI drift-guard step to after the build (commit 12bb9c9). Consequences: the guard is not hermetic (fails locally without a build), has filesystem side effects (writes test/.results/), and the assertion `output != *import-burp*` is a weak proxy — it does not prove the resolved target set is exactly {smoke-check}, only that one unrelated token is absent.

**Recommendation:** Split precedence verification from execution: either add a `--print-targets`/dry-run path to the runner that resolves and prints the target list without requiring a binary or running any test, then assert the resolved list is exactly `smoke-check`; or gate this behavioral test so it is skipped when the binary is absent and update the file docstring to admit it executes the runner. Also assert the full resolved set, not just the absence of `import-burp`.

```diff
override_output=$(env CONFIG_FILE="$tmpconfig" bash -c "source '$RUNNER' --targets smoke-check --group offline --no-build --no-start" 2>&1 || true)
if [[ "$override_output" == *"smoke-check"* ]] && [[ "$override_output" != *"import-burp"* ]]; then
```

</details>

</details>

<details>
<summary><strong>Non-blockers (3)</strong></summary>

<details>

<details>
<summary><code>QUAL-001</code> · nit · <code>.github/workflows/live-tests.yml:65</code></summary>

**Issue:** The new step invokes the script as `bash test/test-runner-args.sh`, while every other script-executing step in the same job (`Run offline tests`, `Setup live targets`, `Run live tests`, `Teardown`) uses the `./test/<script>.sh` form (relying on the executable bit, which is set — the diff shows `new file mode 100755`). Mixing invocation styles in one job is a minor readability/consistency inconsistency.

**Recommendation:** Invoke it the same way as the other steps: `run: ./test/test-runner-args.sh`.

```diff
      - name: Drift guard (target groups vs dispatch)
        run: bash test/test-runner-args.sh
```

</details>

</details>

<details>
<summary>Tests (1)</summary>

<details>
<summary><code>TEST-004</code> · nit · <code>test/test-runner-args.sh:81</code></summary>

**Issue:** The dedup test assigns TARGETS_SETUP as a plain (unexported) shell variable in the test process and never resets it. Today this does not leak into later assertions — the subsequent runner invocations use `env CONFIG_FILE=... bash -c`, which does not inherit the unexported parent variable, and no later in-process code reads TARGETS_SETUP. But it is a latent trap: any future in-process assertion that resolves the `all` group would silently pick up this stale value. Scope it locally or unset it after use.

**Recommendation:** Set TARGETS_SETUP in a subshell for the dedup block, or `unset TARGETS_SETUP` immediately after the assertion, so no global state persists across tests.

```diff
TARGETS_SETUP="rest-api,soap-service,graphql-server,grpc-server,concat-spa"
```

</details>

</details>

<details>

<details>
<summary><code>DOC-001</code> · low · <code>test/README.md:98</code></summary>

**Issue:** run-live-tests.sh gained a user-facing --group offline|live|all flag (the headline feature of this PR), but test/README.md — the canonical 'how to run the suite' doc that CLAUDE.md points to — still documents only --targets at lines 85 and 98. The flag is documented in the script's own --help and in CLAUDE.md, so this is a secondary-doc sync gap, not a blocker.

**Recommendation:** Add --group offline|live|all to the run-live-tests.sh options block in test/README.md (lines ~85 and ~98), mirroring the script usage text.

```diff
  --targets <list>      Comma-separated targets to test (default: all)
```

</details>

</details>

</details>

_3 earlier automated-review findings (all-group ordering, grpc-server config-only, guarded summary pass) were verified as already addressed and filtered from this list._

---

santiago-praetorian (changes requested):

**PR:** https://github.com/praetorian-inc/vespasian/pull/176  
**Linear:** https://linear.app/praetorianlabs/issue/LAB-4773/vespasian-derive-ci-live-test-targets-from-run-live-testssh-to-avoid  
**Head SHA:** caad8ede3a633bf2ca0526147de3a3bdc0bf143e  
**Reviewed at:** 2026-07-14T21:54:02+00:00

|---|---|---|---|---|---|
| Count | 0 | 0 | 2 | 4 | 2 |
| Blockers | 0 | 0 | 2 | 1 | 0 |

---

<details>
<summary><strong>Blockers (3)</strong></summary>

<details>
<summary><code>QUAL-003</code> · medium · QUAL · <code>test/run-live-tests.sh:2856</code></summary>

**Issue:** The new --group -> target-list resolution logic (offline/live/all plus TARGETS_SETUP dedup) is inlined in main(), which is already ~145 lines. Because it lives inside main() rather than a standalone function, the drift-guard test has to reconstruct it by sed-extracting raw source text (source <(sed -n '/^OFFLINE_TARGETS=(/,/^)/p' "$RUNNER"), source <(grep '^join_targets()' "$RUNNER")) and by sourcing the whole script with --dry-run. That extraction is brittle: any reformatting of the array declarations or join_targets one-liner would silently break the sed/grep patterns and could make the drift guard source stale/empty data without an obvious failure signal.

**Recommendation:** Extract the group-resolution block into a small function, e.g. resolve_targets "$group" "$targets", that echoes the resolved comma-separated string. This shortens main() and lets test-runner-args.sh call the function directly instead of scraping source with sed/grep.

```diff
     if [ -z "$targets" ]; then
         case "${group:-all}" in
             all)
                 targets="$(join_targets "${OFFLINE_TARGETS[@]}"),$(join_targets "${LIVE_TARGETS[@]}")"
                 if [ -n "${TARGETS_SETUP:-}" ]; then
                     targets="${TARGETS_SETUP},${targets}"
                     targets="$(echo "$targets" | tr ',' '\n' | awk '!s[$0]++' | paste -sd, -)"
                 fi
                 ;;
```

</details>

<details>
<summary><code>TEST-001</code> · low · TEST · <code>test/run-live-tests.sh:2868</code></summary>

**Issue:** The 'all' group prepends TARGETS_SETUP to OFFLINE+LIVE then deduplicates (run-live-tests.sh:2870-2871). The merge test in test-runner-args.sh asserts only occurrence COUNTS (grpc-server once, rest-api once) — it never asserts the resolved ORDER of the targets= string. AC3 requires ordering be preserved (offline before live setup); an ordering regression in the prepend/dedup logic (or a change to awk '!s[$0]++') would not be caught. This is new public behavior (the TARGETS_SETUP merge) whose ordering contract is unverified.

**Recommendation:** In the TARGETS_SETUP test, additionally assert the exact resolved targets= string for a known TARGETS_SETUP value (via --group all --dry-run) so both dedup and order are pinned, matching the exact-string style already used for the --group offline/live/all resolution tests.

```diff
                     targets="${TARGETS_SETUP},${targets}"
                     targets="$(echo "$targets" | tr ',' '\n' | awk '!s[$0]++' | paste -sd, -)"
```

</details>

<details>
<summary><code>TEST-003</code> · medium · TEST · <code>test/test-runner-args.sh:74</code></summary>

**Issue:** The 'phantom-target' negative case does NOT run the real drift-guard code path (the dispatch-in-group check at lines 46-53). It copies that loop inline (lines 76-82) with its own guard_caught flag, then asserts on the copy. Because the assertion validates a duplicated loop rather than the production guard, a regression in the real guard would go undetected: e.g. changing grep -qx to grep -q, or deleting the else fail branch, would silently disable drift detection while this self-test still passes. If you deleted/broke the real guard, this test would still be green — false regression confidence for the guard that is the entire point of the file.

**Recommendation:** Extract the dispatch-coverage check into a single function that both the positive check and the negative self-test invoke, OR drive the real guard loop with an injected synthetic DISPATCH_TARGETS containing 'phantom-target' and assert the guard actually increments FAIL. The self-test must fail if the real guard code is broken.

```diff
 synthetic_dispatch=("${DISPATCH_TARGETS[@]}" "phantom-target")
 all_grouped_with_config=("${OFFLINE_TARGETS[@]}" "${LIVE_TARGETS[@]}" "${CONFIG_ONLY[@]}")
 for target in "${synthetic_dispatch[@]}"; do
     if ! printf '%s\n' "${all_grouped_with_config[@]}" | grep -qx "$target"; then
```

</details>

</details>

<details>

<details>

<details>
<summary><code>QUAL-001</code> · nit · <code>.github/workflows/live-tests.yml:64</code></summary>

**Issue:** test/test-runner-args.sh only sources bash arrays/functions from run-live-tests.sh and invokes it with --dry-run; it never calls the vespasian binary. Placing the 'Drift guard' step after 'Build vespasian' means a drift failure is only caught after paying for the full Go build, instead of failing fast.

**Recommendation:** Move the 'Drift guard (target groups vs dispatch)' step before 'Build vespasian' so array/dispatch drift is caught immediately without waiting on the build.

```diff
       - name: Drift guard (target groups vs dispatch)
         run: ./test/test-runner-args.sh
```

</details>

<details>
<summary><code>QUAL-002</code> · low · <code>test/run-live-tests.sh:33</code></summary>

**Issue:** The block introducing OFFLINE_TARGETS/LIVE_TARGETS is commented as the 'single source of truth', but the case-dispatch block still has a grpc-server) entry deliberately excluded from both arrays (only reachable via TARGETS_SETUP or explicit --targets). Nothing in run-live-tests.sh itself documents this third, config-only category, so a reader of this file alone would not know why a dispatchable target is missing from the arrays that claim to enumerate all targets.

**Recommendation:** Add a short comment near the OFFLINE_TARGETS/LIVE_TARGETS declaration noting grpc-server (and any future config-only targets) is intentionally omitted from both groups and only runs via TARGETS_SETUP or explicit --targets, mirroring CONFIG_ONLY in test/test-runner-args.sh.

```diff
 # ──────────────────────────────────────────────────────────────
 # Target groups (single source of truth — CI references these
 # via --group instead of maintaining its own target lists)
 # ──────────────────────────────────────────────────────────────
 OFFLINE_TARGETS=(
```

</details>

<details>
<summary><code>QUAL-004</code> · nit · <code>test/run-live-tests.sh:2874</code></summary>

**Issue:** Every other invalid-argument path in main()'s option loop (unknown flag, --group/--targets missing a value) calls usage() before exiting, but the 'Unknown group' case only logs the failure and exits — it doesn't print the option summary. Minor inconsistency in error-path style within the same function.

**Recommendation:** Call usage() before exit 1 in the Unknown group branch for consistency (or intentionally omit usage() from all value-errors — pick one convention).

```diff
             *)
                 log_fail "Unknown group: ${group} (valid: offline, live, all)"
                 exit 1
                 ;;
```

</details>

</details>

<details>
<summary>Tests (2)</summary>

<details>
<summary><code>TEST-002</code> · low · <code>test/test-runner-args.sh:70</code></summary>

**Issue:** The drift guard performs two independent checks: (a) every group member has a case-dispatch entry and (b) every dispatch target is grouped or config-only. Only direction (b) has a negative self-test (phantom-target). Direction (a) has no negative case, so a break in the group-member check would not be caught by any self-test. Edge-case coverage of the guard's failure modes is asymmetric.

**Recommendation:** Add a negative self-test for direction (a) — e.g. a synthetic group array containing a member with no dispatch entry — ideally routed through a shared guard function so both directions are exercised via real code.

```diff
 for target in "${OFFLINE_TARGETS[@]}" "${LIVE_TARGETS[@]}"; do
         fail "Group member '$target' has no case-dispatch entry in run-live-tests.sh"
     pass "Drift guard detects ungrouped dispatch target (phantom-target)"
```

</details>

<details>
<summary><code>TEST-004</code> · low · <code>test/test-runner-args.sh:143</code></summary>

**Issue:** The three argument-validation tests (invalid --group, --group without value, --targets without value) capture combined output with 2>&1 || true, which discards the exit status, then assert only that an error substring is present. The comments promise the runner 'exits non-zero', but that contract is never verified. A regression that still prints the error message but drops exit 1 would leave CI green even though the runner no longer fails.

**Recommendation:** Capture the exit code and assert it is non-zero in addition to the message check, e.g. invalid_output=$(... 2>&1); rc=$? then assert [[ $rc -ne 0 ]] alongside the substring match.

```diff
 invalid_output=$(env CONFIG_FILE="$tmpconfig" bash -c "source '$RUNNER' --group bogus" 2>&1 || true)
 if [[ "$invalid_output" == *"Unknown group"* ]]; then
```

</details>

</details>

</details>

---

eli-wald:
### ✅ Live tests validated

The refactored `--group` workflow in this PR (`3e4c7a6`) has been validated end-to-end:

- **This PR's own `live-tests.yml` run is green** — the `test` job passes ([run 29516098047](https://github.com/praetorian-inc/vespasian/actions/runs/29516098047)).
- **Regression-detection harness — 9/9 as designed.** Because the harness intentionally breaks individual targets (which can't live on this repo), it runs on my fork against the *identical* workflow from this PR. Every scenario behaved exactly as expected:

| Scenario | Expected | Result |
|---|---|---|
| break Burp XML import | fail | ✅ failure |
| break SOAP detection | fail | ✅ failure |
| suppress GraphQL mutation types | fail | ✅ failure |
| alter static-asset filter regex | fail | ✅ failure |
| break numeric path normalization | fail | ✅ failure |
| no `skip-live-tests` label | pass (runs) | ✅ success |
| consolidated run-all | pass | ✅ success |
| `skip-live-tests` label present | `test` job skipped | ✅ `test: skipped` |
| push → `main` | pass | ✅ success |

All five injected regressions — each in a **different** target (Burp import, SOAP, GraphQL, static-asset filter, path normalization) — were caught, confirming `--group offline` / `--group live` exercise the full target set rather than silently skipping any. The clean cases pass, and the `skip-live-tests` gate correctly skips the `test` job (verified at the job level, not just a green run overall).

eli-wald:
### Note: the group-count anchor trade-off (re: TEST-003)

`test/test-runner-args.sh` pins the group sizes as **literals**, deliberately *not* derived from the arrays:

```bash
[[ ${#OFFLINE_TARGETS[@]} -eq 19 ]]
[[ ${#LIVE_TARGETS[@]}    -eq 6  ]]
[[ $group_count           -eq 25 ]]
```

The behavioral `--group` tests already compare the resolved list against `join_targets "${OFFLINE_TARGETS[@]}"` — but that expected value is derived from the same array under test, so a *coordinated* silent drop (remove a target from its group array, its `case` arm, and its test function together) shrinks expected and actual in lockstep and still passes green. The hardcoded literals are the tripwire that AC#3 ("all 25 targets still run") actually relies on.

**What forces the count to be updated when adding a target.** Adding a target is a three-part edit to the runner:

1. add it to `OFFLINE_TARGETS` / `LIVE_TARGETS`,
2. add its `case` dispatch arm,
3. bump the matching literal (`19 → 20`, `25 → 26`).

Do (1) + (2) but skip (3), and the drift-guard step fails fast — it runs *before* the build — with a descriptive message:

```
FAIL: OFFLINE_TARGETS count drifted: expected 19, got 20
```

CI stays red until the literal is updated, so a change to the target count can only happen as an **explicit, acknowledged edit** — never silently. Both directions are covered:

- **silent drop** → array falls to 18, literal still 19 → **fails** (the exact regression AC#3 guards against);
- **silent add** → array rises to 20, literal still 19 → **fails**.

**This does not reopen AC#2** ("adding a target needs no workflow change"): the count literal lives in `test/test-runner-args.sh`, not `live-tests.yml`. The workflow keeps referencing `--group`, untouched. The cost is one extra one-line edit in the test file — the price of making any change to the target set impossible to do silently.

santiago-praetorian (changes requested):

**PR:** https://github.com/praetorian-inc/vespasian/pull/176  
**Linear:** https://linear.app/praetorianlabs/issue/LAB-4773/vespasian-derive-ci-live-test-targets-from-run-live-testssh-to-avoid  
**Head SHA:** 3e4c7a695d31dfb0dcf5c982fde0ac6abca7f977  
**Reviewed at:** 2026-07-16T18:19:30+00:00

|---|---|---|---|---|---|
| Count | 0 | 0 | 1 | 1 | 1 |

---

<details>

<details>
<summary><code>TEST-001</code> · medium · TEST · <code>test/run-live-tests.sh:131</code></summary>

**Issue:** This PR adds a security-hardening allowlist to load_config (new in this PR; base had only format validation): only REST_API_PORT/SOAP_SERVICE_PORT/GRAPHQL_SERVER_PORT/GRPC_SERVER_PORT/CONCAT_SPA_PORT/TARGETS_SETUP may be set from the config file, so a crafted config cannot rebind VESPASIAN, PATH, RESULTS_DIR, or TEST_HOST. No test anywhere exercises the disallowed-key branch (log_warn "Skipping unexpected config key"). If the allowlist were removed or the condition inverted, no test would fail and the hardening would silently regress.

**Recommendation:** Add a test that feeds load_config a config containing a disallowed key (e.g. VESPASIAN=/tmp/evil) and asserts the variable is NOT set/exported and the 'Skipping unexpected config key' warning is emitted, while an allowed key (e.g. REST_API_PORT) IS applied.


```diff
    local allowed_keys=" REST_API_PORT SOAP_SERVICE_PORT GRAPHQL_SERVER_PORT GRPC_SERVER_PORT CONCAT_SPA_PORT TARGETS_SETUP "
            if [[ "$allowed_keys" != *" ${key} "* ]]; then
                log_warn "Skipping unexpected config key: ${key}"
                continue
```

</details>

</details>

<details>

<details>

<details>
<summary><code>QUAL-001</code> · nit · <code>test/run-live-tests.sh:23</code></summary>

**Issue:** run-live-tests.sh now honors a CONFIG_FILE env var to relocate the config path (changed this PR from a plain assignment to CONFIG_FILE="${CONFIG_FILE:-${SCRIPT_DIR}/.live-test-config}"). It is an internal test-harness knob (used by test/test-runner-args.sh's dry-run invocations) and is documented nowhere: not in usage(), test/README.md, or CLAUDE.md. Minor discoverability gap.

**Recommendation:** Optionally note CONFIG_FILE in the run-live-tests.sh usage()/Environment section, mentioning it defaults to test/.live-test-config.


```diff
CONFIG_FILE="${CONFIG_FILE:-${SCRIPT_DIR}/.live-test-config}"
```

</details>

</details>

<details>
<summary>Tests (1)</summary>

<details>
<summary><code>TEST-002</code> · low · <code>test/test-runner-args.sh:71</code></summary>

**Issue:** The drift guard checks two directions: every group member has a case-dispatch arm (direction a), and every dispatch target is grouped or config-only (direction b). It never checks that CONFIG_ONLY members (grpc-server) actually have a dispatch arm. If grpc-server's case arm were deleted while it stayed in CONFIG_ONLY, the guard would still pass green even though '--targets grpc-server' would then hit the unknown-target path. Coverage of the guard's failure modes is asymmetric.

**Recommendation:** Extend the group-member dispatch check (or add a third loop) to also require every CONFIG_ONLY member to have a case-dispatch arm, so config-only targets can't silently lose their dispatch path.


```diff
CONFIG_ONLY=(grpc-server)
report_undispatched_group_members "${OFFLINE_TARGETS[@]}" "${LIVE_TARGETS[@]}"
report_ungrouped_dispatch_targets "${DISPATCH_TARGETS[@]}"
```

</details>

</details>

</details>


_Re-review (round 3): prior capability-pr-review rounds at 12bb9c9 and caad8ede reported 4–5 blockers each; all verified fixed in the current head, and all automated PR review comments (CodeRabbit all-group ordering, unconditional summary pass, grpc-server; Codex TARGETS_SETUP-additive) are resolved. All three Linear acceptance criteria are met by the current code (AC3's Linear checkbox is stale — offline/live groups exactly reproduce the base 25 targets and offline runs before live setup)._

---

santiago-praetorian:

**PR:** https://github.com/praetorian-inc/vespasian/pull/176  
**Linear:** https://linear.app/praetorianlabs/issue/LAB-4773/vespasian-derive-ci-live-test-targets-from-run-live-testssh-to-avoid  
**Head SHA:** bea7ed2f5915e5862278285d83d3af9c66f583b8  
**Reviewed at:** 2026-07-16T00:00:00Z

**Recommendation:** APPROVE

✅ All three LAB-4773 acceptance criteria are met (verified base-vs-head: `OFFLINE_TARGETS`=19 + `LIVE_TARGETS`=6 reproduce the base 25 targets, offline runs before live setup, `grpc-server` was never in base CI and stays config-only). All prior-round blockers are resolved. **The items below are non-blocking (1 nit + 3 low) — optional cleanup, none gate the merge.**

|---|---|---|---|---|---|
| Count | 0 | 0 | 0 | 3 | 1 |

---

<details>

<details>

<details>
<summary><code>QUAL-001</code> · nit · <code>.github/workflows/live-tests.yml:55</code></summary>

**Issue:** The updated job comment has two spaces after "one job." ("Runs all test targets in one job.  Target groups are defined in"), inconsistent with single-space sentence spacing used elsewhere in the same file/PR.

**Recommendation:** Collapse to a single space for consistency.


```diff
  # Runs all test targets in one job.  Target groups are defined in
```

</details>

<details>
<summary><code>QUAL-002</code> · low · <code>test/run-live-tests.sh:2853</code></summary>

**Issue:** main() spans ~137 lines and mixes CLI argument parsing (8-arm case statement), conditional config loading, target-group resolution, dry-run short-circuiting, preflight, build, and test dispatch/summary — well past the ~60-line guideline in rubric item 1, and with roughly a dozen independent branch points (case arms, per-flag missing-value checks, the compound dry-run/load_config condition, the empty-targets/resolve_targets-failure check, and the dry-run early-return) it is in the same complexity band the analogous main() in test/setup-live-targets.sh explicitly calls out and justifies with a comment ("Intentionally longer than the ~60-line guideline: each stage is a distinct sequential step...", lines 773-776 of that file). This PR added the --group/--dry-run arms and the multi-condition load_config gate to this already-long function without a matching justification comment, so a reviewer/future maintainer has no signal that the length is an accepted tradeoff versus an oversight.

**Recommendation:** Either add a short comment (mirroring setup-live-targets.sh's main()) explaining why this orchestration function is intentionally long, or extract the argument-parsing loop into a parse_args()-style helper (as setup-live-targets.sh already does) to bring main() under the guideline.


```diff
main() {
    if [ "$dry_run" != true ] || { [ -z "$targets" ] && [ "${group:-all}" = all ]; }; then
        load_config
    fi
```

</details>

<details>
<summary><code>QUAL-003</code> · low · <code>test/setup-live-targets.sh:765</code></summary>

**Issue:** run_tests_guidance() decides 'full setup' vs 'partial setup' guidance by exact string comparison of the resolved --targets value against the literal ALL_TARGETS constant ("rest-api,soap-service,graphql-server,grpc-server,concat-spa"). PARSED_TARGETS is taken verbatim from the user's --targets flag (line 725: PARSED_TARGETS="$2") with no sorting, dedup, or normalization. If a user runs setup-live-targets.sh --targets with all five targets but in a different order (or with a harmless duplicate), the string comparison fails even though the set is semantically identical to ALL_TARGETS, so the function takes the 'partial setup' branch and prints the misleading 'Run the targets you set up: ./test/run-live-tests.sh --targets <reordered-list>' + offline-only hint instead of the simpler 'Run tests with: ./test/run-live-tests.sh' message. Not functionally breaking (the printed --targets command still works), but the guidance text is wrong for a semantically-full setup.

**Recommendation:** Compare as sets rather than raw strings, e.g. sort+dedup both $targets and $ALL_TARGETS before comparing (tr ',' '\n' | sort -u), or canonicalize PARSED_TARGETS immediately after parsing so downstream comparisons are order-independent.


```diff
    if [ "$targets" = "$ALL_TARGETS" ]; then
                PARSED_TARGETS="$2"
```

</details>

</details>

<details>
<summary>Tests (1)</summary>

<details>
<summary><code>TEST-001</code> · low · <code>test/test-runner-args.sh:475</code></summary>

**Issue:** The load_config allowlist test exercises only two of load_config's three branches: an allowed key applied (REST_API_PORT=8990) and a disallowed key skipped (VESPASIAN=/tmp/evil -> 'Skipping unexpected config key'). All three config lines fed to it are format-VALID, so the third branch — run-live-tests.sh:147-149 'Skipping invalid config line' — is never hit. That else branch is reached only when a line fails the KEY=VALUE format regex (^[A-Za-z_][A-Za-z0-9_]*=[A-Za-z0-9_./:@,+\"\ -]*$), which is the security gate that stops injection-shaped values (containing $, backticks, parentheses, ;, &, |) from ever reaching `declare -g`. The comment above the test explicitly frames this as security hardening ('a crafted or hand-edited config can never rebind security-relevant globals'), yet a regression that loosened the regex to admit metacharacters would pass all current assertions unnoticed. This is an edge-case coverage gap on the security-relevant half of the config parser.

**Recommendation:** Add one assertion to the existing load_config block: write a config line with a shell-metacharacter value (e.g. 'REST_API_PORT=$(touch /tmp/pwn)' or a backtick value) and assert (a) load_config emits 'Skipping invalid config line', and (b) REST_API_PORT retains its sentinel (the malicious value was NOT applied). Reuses the same subshell-echo pattern already in place at lines 477-486.


```diff
if printf '%s\n' "$allowlist_out" | grep -q "Skipping unexpected config key: VESPASIAN"; then
printf '%s\n' "REST_API_PORT=8990" "VESPASIAN=/tmp/evil" "TARGETS_SETUP=" > "$allowlist_cfg"
```

</details>

</details>

</details>

_5 findings from earlier rounds (round-3 allowlist test, CONFIG_ONLY drift check, CONFIG_FILE doc; CodeRabbit all-group ordering + guarded summary pass) were verified as already addressed and filtered from this list._

---

_Review generated by `capability-pr-review`._

santiago-praetorian (approved):
Approving. All three LAB-4773 acceptance criteria are met and every prior-round blocker is resolved. The 4 items in the review comment above are non-blocking (1 nit + 3 low) — optional cleanup, at your discretion.

PR #178, opened 2026-07-16
ci(live-tests): add harden-runner egress monitoring in audit mode (LAB-4732)

santiago-praetorian:
## Summary

Defense-in-depth follow-up to **SEC-BE-002** from the [PR #164 review](https://linear.app/praetorianlabs/issue/LAB-4012) ([LAB-4732](https://linear.app/praetorianlabs/issue/LAB-4732)).

Adds [`step-security/harden-runner`](https://github.com/step-security/harden-runner) as the first step of the `preflight-selftest` and `test` jobs in `.github/workflows/live-tests.yml`, in **`egress-policy: audit`** mode.

## Why

These jobs check out code and pull Go/Node/npm dependencies with unrestricted network egress. A compromised third-party action or transitive dependency could exfiltrate data or reach unexpected hosts. Existing controls already limit blast radius (`permissions: contents: read`, SHA-pinned actions, `persist-credentials: false`, local-only test targets); this adds outbound-traffic monitoring on top.

## Approach

- **Audit mode (non-blocking):** logs all egress to the StepSecurity dashboard without enforcing an allowlist. Cannot break CI. Matches the ticket's *"audit mode first, then block mode"* guidance.
- **`check-label` job skipped:** it performs no checkout and no network I/O.
- **Follow-up path documented in-file:** the expected block-mode allowlist (`github.com`, `proxy.golang.org`, `sum.golang.org`, `storage.googleapis.com`, `objects.githubusercontent.com`, `registry.npmjs.org`) is listed in a comment so flipping to `block` is a one-line change once telemetry confirms the set.
- Action pinned to a full 40-char commit SHA (`v2.20.0`), consistent with repo convention.

> **Note:** the workflow was refactored since the ticket was filed — `browser-actions/setup-chrome` is gone and the job layout is now `preflight-selftest` / `check-label` / `test` (no separate `offline-tests`/`live-tests` jobs). Hardening is applied to the two jobs that actually check out code and hit the network.

## Enterprise Actions allowlist compatibility

The `praetorian-inc` GitHub enterprise restricts workflows to an Actions allowlist. This previously **blocked** `browser-actions/setup-chrome` and forced its removal in [#167](https://github.com/praetorian-inc/vespasian/pull/167) ([LAB-4747](https://linear.app/praetorianlabs/issue/LAB-4747)). That policy accepts an action if it meets **any** of:

1. Owned by the enterprise (`praetorian-inc/*`)
2. Created by GitHub (`actions/*`)
3. **Published by a GitHub Marketplace verified creator**
4. Matches the explicit allowlist (`aws-actions/*`, `1password/install-cli-action`, `currents-dev/playwright-last-failed`, plus specific pinned SHAs)

`step-security/harden-runner` is **not** enterprise-owned, `actions/*`, or on the explicit list — it clears the policy via **path 3**: StepSecurity is a GitHub-verified Marketplace creator (*"GitHub has manually verified the creator of the action as an official partner organization"*). This is exactly the path `browser-actions/setup-chrome` lacked (browser-actions is not a verified creator).

**This is confirmed empirically, not just by policy reading** — on this PR's CI run the enterprise allowed the action and both jobs went green:

```
test  Set up job         Download action repository 'step-security/harden-runner@bf7454d0…'
test  Pre Harden runner  ##[group]Run step-security/harden-runner@bf7454d0…
test  Pre Harden runner    egress-policy: audit
test  Pre Harden runner  [harden-runner] pre-step
```

A blocked action would fail the job at startup with *"...is not allowed in `praetorian-inc/vespasian`..."*; instead `preflight-selftest`, `test`, and `Verify Action Pins` all pass.

> **Reviewer judgment call:** LAB-4747 showed a team preference to avoid third-party actions entirely (it ranked "request an allowlist addition" last, preferring to drop `setup-chrome` rather than allowlist it). harden-runner is allowed *and* passing, but its eBPF egress monitor has no runner-native / plain-`run:` equivalent, so it can't be eliminated the way `setup-chrome` was. Flagging in case the maintainer prefers a different tradeoff.

## Verification

- ✅ CI green on this PR: `preflight-selftest`, `test` (full live suite, ~7m40s), and `Verify Action Pins` all pass **with the harden-runner step in place**
- ✅ Action downloaded and executed by the enterprise-restricted runner (log above) — no "action not allowed" failure
- `actionlint` clean; YAML validates; workflow structure unchanged apart from the added step

## Org-level note

The ticket frames egress hardening as an org-wide decision. If Labs standardizes a policy across GitHub Actions (e.g. in `praetorian-inc/public-workflows`), this workflow should adopt the same policy in place of this one-off.

Closes LAB-4732


eli-wald (commented):

**PR:** https://github.com/praetorian-inc/vespasian/pull/178
**Linear:** https://linear.app/praetorianlabs/issue/LAB-4732
**Head SHA:** ae05dc505c12248a15b84de6aa05bc001fdd53e2
**Reviewed at:** 2026-07-17T18:19:53Z

**Recommendation:** COMMENT

|---|---|---|---|---|---|
| Count | 0 | 0 | 0 | 0 | 1 |

---

<details>

<details>

<details>
<summary><code>QUAL-001</code> · nit · <code>.github/workflows/live-tests.yml:33</code></summary>

**Issue:** Optional extra hardening (Gemini's pending suggestion): add `disable-sudo: true` to both `harden-runner` blocks. This is strictly additive, not a defect — the shipped config is already correct. Verified: no CI-executed script runs `sudo`/`apt` (only doc/log strings mention them); local test servers bind non-privileged ports; Chrome runs sandbox-less (`VESPASIAN_NO_SANDBOX="true"` → `--no-sandbox`), which avoids the setuid-sandbox path; and `setup-go`/`setup-node` write to the runner tool cache without sudo. So dropping sudo removes an unused privilege-escalation vector with no risk to the Go/Node/Chrome setup.

**Recommendation:** Optionally add `disable-sudo: true` under `with:` in BOTH harden-runner steps (preflight-selftest ~line 33 and test ~line 84). Revisit only if a future CI step introduces `sudo`/`apt` (e.g. installing a browser instead of using the runner's stock Chrome). No change required to approve.

```diff
       - name: Harden runner
         uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920  # v2.20.0
         with:
           egress-policy: audit
+          disable-sudo: true
```

</details>

</details>

</details>

_AC met: the ticket's egress-policy option (`harden-runner`, audit mode first) is implemented on the two jobs that check out code and hit the network; `check-label` is correctly excluded. Verified during review: the pin `bf7454d0…` **is** `step-security/harden-runner` `v2.20.0` (checked against the upstream git tag), all actions are SHA-pinned, and audit mode is non-blocking so it cannot break CI (both jobs are green on the PR). Bot-review items were checked and dismissed: the "confirm the SHA" note (Claude) is now confirmed; the `v2.20.0`-is-a-typo claim (Gemini) is a false positive (v2.20.0 is a real release); and `id-token`/private-repo telemetry (Gemini) is N/A — the repo is public and `contents: read` suffices._

---

eli-wald (approved):
`disable-sudo` added to runner config. Approving PR.

PR #186, opened 2026-07-21
feat: route probe, JS-replay, gRPC & sourcemap traffic through --proxy (LAB-4993)

eli-wald:
## Summary

`--proxy` / `--proxy-insecure` previously routed **only crawl traffic** through the intercepting proxy. Every stage after the crawl built its own HTTP client and bypassed the proxy: the active **probe** stage (OPTIONS / JSON-schema / WSDL fetch / GraphQL introspection / gRPC reflection / grpc-gateway), the **WSDL discovery** client, the **JS-replay** stage, and (per scoping decision below) the **jsstatic sourcemap** fetches. In proxy-centric pentest workflows that made a large fraction of Vespasian's outbound traffic invisible to Burp / mitmproxy and could send requests outside the intended proxy path.

Resolves **LAB-4993**.

## Approach

New stdlib-only leaf package **`pkg/httpx`** (imports only `golang.org/x/net/proxy` beyond stdlib; depends on none of crawl/probe/pipeline/ssrf → no import cycle):

- `ProxyConfig{URL, Insecure}` + `BuildHTTPClient(...)` — mirrors `pkg/crawl.newHTTPClient`'s proxy branch exactly.
- `ProxyDialer(...)` — HTTP `CONNECT` tunnel (http/https) and SOCKS5 dialer for the gRPC reflection path, which has no `http.ProxyURL` equivalent.

A single `Proxy httpx.ProxyConfig` field is threaded through `probe.Config`, `crawl.JSReplayConfig`, `pipeline.Options`, `pipeline.ScanOptions`, `pipeline.AugmentOptions`, and `jsstatic.Options`. 5 of 6 probe strategies inherit proxying for free via `Config.Client`; only gRPC needed the bespoke dialer. `--proxy`/`--proxy-insecure` were added to the `generate` command (it runs probe + JS-replay independently). **No new dependency.** `pkg/crawl.newHTTPClient` / `ValidateProxyAddr` are untouched (no regression to the proven crawl path).

## Scope note

`pkg/analyze/jsstatic` sourcemap fetching (`--fetch-sourcemaps`, on by default for crawl/scan) is a third post-crawl active-HTTP stage not enumerated in the ticket. It was **included** so `scan --proxy` doesn't silently leak `.js.map` fetches, consistent with the ticket's intent that everything after the crawl be proxied.

## Acceptance criteria

- [x] With `--proxy`, probe **and** JS-replay (and WSDL, gRPC, jsstatic) traverse the proxy — verified by recording-proxy tests that assert a hit counter.
- [x] `--proxy-insecure` applies to probe/replay TLS the same as the HTTP crawl path; **socks5 keeps verification**.
- [x] Docs updated (CLI `--proxy` help, `pkg/crawl/doc.go`, `CLAUDE.md`, `README.md`) to reflect probe/replay/WSDL/gRPC/jsstatic are now proxy-aware.

## Preserved security invariants

- socks5 **always** verifies TLS; `InsecureSkipVerify` only for http/https proxies under explicit `--proxy-insecure`.
- When proxied, the dial-time SSRF pin is dropped (we dial the proxy, not the target) but URL-level SSRF validators stay enforced — a proxied private target still requires `--dangerous-allow-private`.
- Proxy embedded-credential rejection enforced at the CLI boundary for all three commands.
- `CONNECT` dialer rejects CR/LF in the target address and bounds the handshake by the caller's context deadline.

## Testing & verification

- `make check` (gofmt + vet + golangci-lint + `go test -race ./...`) — clean.
- `go test ./... -count=1` — all packages pass, 0 failures. `golangci-lint run` — 0 issues.
- Recording-proxy end-to-end proofs for each stage; scheme-gated TLS tests; SSRF-via-redirect / CRLF / CONNECT-deadline regression tests.
- Coverage on the new `pkg/httpx` = 77.6% (SOCKS5 forwarding + https-proxy CONNECT paths covered via in-test socks5 + self-signed TLS proxy servers).

## Review

Independent code + security review performed. One MEDIUM (SSRF-via-redirect on the proxied sourcemap client — it followed redirects while jsstatic only scopes the initial URL) was found and fixed (`noFollowRedirects`, matching the other proxied clients), plus two LOW hardening items (CONNECT CRLF guard, CONNECT I/O deadline) — all regression-tested.


santiago-praetorian:

**PR:** https://github.com/praetorian-inc/vespasian/pull/186
**Linear:** https://linear.app/praetorianlabs/issue/LAB-4993/vespasian-probe-and-js-replay-stages-ignore-proxy-only-crawl-traffic
**Head SHA:** b58431c2ceb9386d459fb84d725d43a90cb78ff1
**Reviewed at:** 2026-07-22T00:48:35Z

|---|---|---|---|---|---|
| Count | 0 | 0 | 1 | 2 | 2 |

---

<details>

<details>
<summary><code>QUAL-002</code> · medium · QUAL · <code>pkg/httpx/proxy_client.go:156</code></summary>

**Issue:** `connectDialer`'s doc comment claims the watcher goroutine "is stopped (`close(done)`) the instant the handshake finishes … so it can never close the tunnel conn we hand back." That invariant does not hold: the watcher's `select { case <-ctx.Done(): conn.Close(); case <-done: }` and the main goroutine's `close(done)` (immediately after `connectHandshake` returns) are racing. If `ctx` is canceled at (or just before) the moment `connectHandshake` succeeds — but before `close(done)` executes — the watcher can observe `ctx.Done()` as the only ready case and call `conn.Close()` on the connection that was just successfully established. `handshakeErr` is `nil` at that point, so `connectDialer` returns the now-closed conn as a success. The caller (gRPC reflection dial) then receives a seemingly-healthy `net.Conn` that fails on first Read/Write with a generic "use of closed connection" error instead of a clear cancellation error at dial time.

**Recommendation:** After `close(done)`, check `ctx.Err()` before returning success (e.g. `if ctx.Err() != nil { conn.Close(); return nil, ctx.Err() }`), or synchronize the watcher and the success path with a shared atomic/mutex so `conn.Close()` cannot race past a successful handshake return. Add a regression test that cancels `ctx` concurrently with a slow-but-successful CONNECT handshake to confirm the returned conn (or error) is well-defined.

```
// The watcher is stopped (close(done)) the instant the handshake
// finishes — success or error — so it can never close the tunnel conn we
// hand back.
br, handshakeErr := connectHandshake(ctx, conn, addr)
close(done)
```

</details>

</details>

<details>

<details>

<details>
<summary><code>SEC-BE-001</code> · low · <code>pkg/analyze/jsstatic/sourcemap.go:286</code></summary>

**Issue:** On every proxied stage (probe, WSDL, JS-replay, gRPC reflection, jsstatic sourcemap) the dial-time SSRF pin is intentionally dropped because the client dials the proxy, not the target. The URL-level validator resolves the target hostname at validation time, but the operator's proxy re-resolves it independently at connect time. A hostile scanned target can use DNS rebinding (public IP at validation, private IP at proxy connect) to make the operator's proxy reach an internal host and relay the response back, defeating the private-target guard even without `--dangerous-allow-private`. Impact is bounded: requires operator to have explicitly enabled `--proxy` (their own trusted Burp/mitmproxy), the attacker to control target DNS and win the rebinding race, and the fetched content is relayed only into the operator's own tooling. Documented as an intentional tradeoff; matches the pre-existing accepted crawl proxy path. CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:L/I:N/A:N (3.7, Low).

**Recommendation:** Accept as a documented design tradeoff (recommended given the operator-controlled-proxy threat model), OR enforce SSRF destination policy at proxy-connect time (connect to a validated IP while preserving original Host/SNI) and narrow the guarantee wording in `pkg/httpx/doc.go`. At minimum, ensure operator-facing docs make the proxied-path DNS-rebinding residual explicit so operators do not assume `--proxy` preserves the non-proxy dial-time DNS-rebinding mitigation.

```
return httpx.BuildHTTPClient(proxy, 10*time.Second, noFollowRedirects)
// Route through the proxy: no dial-time SSRF pin (we dial the proxy, not
// the target). This is the production path (pipeline leaves HTTPClient
```

</details>

</details>

<details>

<details>
<summary><code>QUAL-001</code> · low · <code>cmd/vespasian/main.go:500</code></summary>

**Issue:** `GenerateCmd.resolveJSReplayConfig()` already calls `parseProxyConfig(c.Proxy, c.ProxyInsecure)` (fail-fast, validated) and stores the result on the returned `JSReplayConfig.Proxy`. `GenerateCmd.Run()` then calls `c.options()`, which independently re-parses the same `--proxy`/`--proxy-insecure` flags via `parseProxyConfigOrEmpty` instead of reusing `jsReplayCfg.Proxy`. This is the exact anti-pattern the PR's own head commit (b58431c) fixed for `ScanCmd`, left in place for `GenerateCmd`. Currently harmless (pure function of the same immutable inputs), but it duplicates `url.Parse` + `ValidateProxyAddr` work per generate invocation and diverges from the single-parse pattern the PR establishes elsewhere.

**Recommendation:** Thread the already-parsed `jsReplayCfg.Proxy` into `GenerateCmd.options()` (pass it as a parameter, mirroring how `ScanCmd.scanOptions` now takes a `proxy httpx.ProxyConfig` argument) instead of re-parsing `c.Proxy` inside `options()`.

```
func (c *GenerateCmd) options() pipeline.Options {
	proxy := parseProxyConfigOrEmpty(c.Proxy, c.ProxyInsecure)
	proxy, err := parseProxyConfig(c.Proxy, c.ProxyInsecure)
```

</details>

</details>

<details>
<summary>Tests (2)</summary>

<details>
<summary><code>TEST-001</code> · nit · <code>internal/pipeline/augment_test.go:353</code></summary>

**Issue:** The forwarding recording-proxy handler is re-implemented inline in `augment_test.go` (`TestAnalyzeJS_ForwardsProxy`) and `resolve_generate_test.go` (`TestResolveAndGenerate_ForwardsProxy`) even though `pipeline_test.go` already defines a reusable `newRecordingProxy` helper in the same package. The three copies differ only in whether they forward the response body/headers. Cross-package copies (httpx, probe, crawl, jsstatic) are justified (unexported helpers); this is only about the avoidable within-package duplication.

**Recommendation:** Extend `newRecordingProxy` (`pipeline_test.go`) with an option to forward the response body/headers and reuse it from `augment_test.go` and `resolve_generate_test.go`. Low priority — readable test setup can outweigh strict DRY.

```
proxy := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
	hits.Add(1)
	outReq, err := http.NewRequestWithContext(r.Context(), r.Method, r.RequestURI, nil) //nolint:gosec // test proxy forwards the received request URI
```

</details>

<details>
<summary><code>TEST-002</code> · nit · <code>pkg/httpx/proxy_client_test.go:707</code></summary>

**Issue:** `TestProxyDialer_ConnectHandshakeRespectsContextCancel` sequences cancellation with a fixed `time.Sleep(100ms)` in a separate goroutine before calling `cancel()` — a wall-clock timing dependency in an otherwise deterministic suite. In practice robust (the stub proxy never writes a status line, so the dial cannot complete before cancel), but the pattern relies on scheduling rather than a synchronization signal, so under heavy CI load the 100ms window is technically not guaranteed to land while the handshake is in-flight.

**Recommendation:** Signal from the stub proxy's `Accept` (close a channel once the connection is accepted) and cancel on that signal instead of after a fixed sleep. Optional — the current test is bounded by the 2s select ceiling and is not flaky in normal conditions.

```
go func() {
	time.Sleep(100 * time.Millisecond)
	cancel()
}()
```

</details>

</details>

</details>

_8 findings dismissed as already addressed (prior-review fixes: gosec TLS MinVersion/InsecureSkipVerify/unhandled-errors, CONNECT tunnel drain, SSRF-via-redirect on sourcemap, CONNECT CRLF guard, CONNECT I/O deadline, ScanCmd double-parse QUAL-001) — see `.review/addressed.json`._

---

santiago-praetorian (changes requested):
REQUEST CHANGES — 1 blocker (QUAL-002, medium): `connectDialer` in pkg/httpx/proxy_client.go:156 has a race between the ctx.Done() watcher goroutine and close(done); on cancellation at handshake success it can close the just-established tunnel yet still return it as a successful dial, so the gRPC reflection caller gets a conn that fails on first use instead of a clean cancellation error. Fix: check ctx.Err() after close(done) and add a concurrent-cancel regression test. Full findings (SEC-BE-001, QUAL-001/002, TEST-001/002) and the acceptance-criteria assessment are in the review roll-up comment above.

santiago-praetorian:

**PR:** https://github.com/praetorian-inc/vespasian/pull/186
**Linear:** https://linear.app/praetorianlabs/issue/LAB-4993/vespasian-probe-and-js-replay-stages-ignore-proxy-only-crawl-traffic
**Head SHA:** 441bec792b21eb8bb389293d4157d76e239a0573
**Reviewed at:** 2026-07-27T16:45:00Z

|---|---|---|---|---|---|
| Count | 0 | 0 | 3 | 9 | 5 |
| Blockers | 0 | 0 | 3 | 0 | 0 |

---

<details>
<summary><strong>Blockers (3)</strong></summary>

<details>
<summary><code>SEC-BE-001</code> · medium · SEC-BE · <code>cmd/vespasian/main.go:709</code></summary>

**Issue:** This PR threads the parsed ProxyConfig (which carries --proxy-insecure) into every post-crawl stage: jsstatic sourcemap fetch (crawl + scan + generate), probe, WSDL discovery and JS-replay. All of those stages are pure net/http and run regardless of --headless. The CrawlOptions.ProxyInsecure help text (main.go:240) and the matching README blocks for `scan` and `crawl` were NOT updated and still scope the flag to "an http/https intercepting proxy (Burp/mitmproxy MITM) on the net/http backend (--headless=false) ... no effect on socks5 or the headless backend". Concretely, `vespasian crawl --proxy https://p:8080 --proxy-insecure` in the (default) headless mode now builds an InsecureSkipVerify:true client for sourcemap fetches, and `vespasian scan --headless=true --proxy https://p:8080 --proxy-insecure` does so for the OPTIONS/schema/WSDL/GraphQL probes and JS-replay (which forwards -H auth headers same-origin). Operator risk: an operator who reads the flag help and believes it is inert in headless mode gets certificate validation silently turned off on several outbound stages, so an on-path attacker between the proxy and the target can substitute a certificate and tamper with (or capture credentialed) probe/replay traffic without any warning. The invariant itself is implemented correctly in code (socks5 always verifies, http/https only under the explicit opt-in) — the defect is that the security control's documented blast radius is now narrower than the real one.

**Recommendation:** Update the CrawlOptions.ProxyInsecure struct-tag help (main.go:240) and the two README `--proxy-insecure` blocks to state that the flag disables TLS verification for every net/http stage routed through an http/https proxy — crawl (only with --headless=false), plus jsstatic sourcemap fetch, probe, WSDL discovery, JS-replay and the gRPC reflection dial — regardless of --headless. GenerateCmd.ProxyInsecure (main.go:485) already has accurate wording; mirror it. Optionally also emit a one-line stderr notice when --proxy-insecure is set, next to warnSSRFDisabled, listing the stages that will skip verification.

```
	scanProxy := parseProxyConfigOrEmpty(c.Proxy, c.ProxyInsecure)
		Proxy:           scanProxy,
	crawlProxy := parseProxyConfigOrEmpty(c.Proxy, c.ProxyInsecure)
	ProxyInsecure   bool          `name:"proxy-insecure" help:"Disable TLS certificate verification for an http/https intercepting proxy (Burp/mitmproxy MITM) on the net/http backend (--headless=false). Off by default; no effect on socks5 or the headless backend (there, trust the proxy CA via the OS trust store instead)."`
```

</details>

<details>
<summary><code>SEC-BE-004</code> · medium · SEC-BE · <code>pkg/probe/grpc.go:241</code></summary>

**Issue:** dialGRPC now ORs the proxy's Insecure flag into the target's tls.Config.InsecureSkipVerify whenever the proxy scheme is http/https. That correctly covers the MITM case (Burp terminating the CONNECT tunnel and re-signing), but it is unconditional: a plain http/https proxy used as a pure CONNECT tunnel (no interception configured for that host/port, an egress proxy, or a Squid-style forward proxy) presents the real target certificate through the tunnel, and in that case the operator loses gRPC certificate validation for no reason. Vespasian already exposes a dedicated, separately documented opt-in for exactly this (--grpc-insecure-skip-verify, default: verify); --proxy-insecure now overrides it implicitly, and neither the scan/crawl nor the generate help text for --proxy-insecure mentions gRPC at all. Operator risk: an attacker on the path between the operator's proxy and the gRPC target can present any certificate, MITM the reflection exchange, and feed forged FileDescriptorProtos into the generated .proto (integrity of the assessment artifact) as well as observe the reflection dial. Note the socks5 half of the gate is correct and is covered by TestGRPCProbe_ProxyInsecureTLSGating/socks5_proxy_insecure_does_not_skip_verify.

**Recommendation:** Either (a) keep the coupling but document it in both --proxy-insecure help strings and the README ("also skips TLS verification for the gRPC reflection dial when tunneled through an http/https proxy"), or (b) drop the implicit OR and require the operator to pass --grpc-insecure-skip-verify explicitly for the gRPC stage, so each TLS-verification opt-out remains independently auditable. (b) is preferable because the flags already exist and gRPC-over-CONNECT is frequently not intercepted.

```
	insecureSkip := p.config.GRPCInsecureSkipVerify
	if p.config.Proxy.Enabled() && p.config.Proxy.Insecure {
		if s := p.config.Proxy.URL.Scheme; s == "http" || s == "https" {
			insecureSkip = true
			InsecureSkipVerify: insecureSkip, // #nosec G402 -- opt-in only (default verifies); SSRF is enforced by the URLValidator preflight + Dialer
```

</details>

<details>
<summary><code>TEST-011</code> · medium · TEST · <code>pkg/probe/types.go:152</code></summary>

**Issue:** AC-2 requires test-verifiable proof that --proxy-insecure applies to probe/replay TLS the same way as the HTTP crawl path. At head, `Insecure: true` appears in only four test call sites: pkg/httpx/proxy_client_test.go:79 and :94 (BuildHTTPClient leaf unit tests, incl. the socks5-always-verifies case) and pkg/probe/grpc_test.go:1152 and :1179 (the gRPC ProxyDialer path, which does not go through BuildHTTPClient). No test sets Proxy.Insecure=true on probe.Config, crawl.JSReplayConfig, jsstatic.Options, pipeline.Options or pipeline.ScanOptions. So nothing pins that the Insecure bit survives the withDefaults -> httpx.BuildHTTPClient hop for the five HTTP probe strategies, for JS-replay (pkg/crawl/jsreplay.go:149), or for sourcemap fetching (pkg/analyze/jsstatic/sourcemap.go:296). A regression that passed `httpx.ProxyConfig{URL: cfg.Proxy.URL}` instead of `cfg.Proxy` at any of those three sites would silently restore TLS verification against an intercepting MITM proxy — probe/replay would start failing against Burp/mitmproxy under --proxy-insecure — and the entire suite would still be green. AC-2 for the HTTP paths currently rests on inference from the leaf test plus struct-forwarding, not on any assertion. The three existing tests already inspect `tr.TLSClientConfig`/`tr.Proxy` in exactly this style, so the guard is one extra subtest each.

**Recommendation:** Add an Insecure=true case to TestConfig_WithDefaults_ProxyClient (pkg/probe/proxy_internal_test.go), TestJSReplayConfig_WithDefaults_ProxyClient (pkg/crawl/jsreplay_test.go) and TestDefaultSourcemapClient_ProxyVsSSRF (pkg/analyze/jsstatic/sourcemap_test.go) asserting `tr.TLSClientConfig.InsecureSkipVerify == true` for an http/https proxy URL, plus a socks5 subtest asserting it stays false — mirroring TestBuildHTTPClient_InsecureOnlyForHTTPScheme / TestBuildHTTPClient_SOCKS5AlwaysVerifies one layer up.

```
		if cfg.Proxy.Enabled() {
			cfg.Client = httpx.BuildHTTPClient(cfg.Proxy, cfg.Timeout, probeRedirectPolicy)
```

</details>

</details>

<details>
<summary><strong>Non-blockers (14)</strong></summary>

<details>
<summary>Security (2)</summary>

<details>
<summary><code>SEC-BE-002</code> · nit · <code>pkg/httpx/proxy_client.go:140</code></summary>

**Issue:** connectDialer guards the CONNECT request line with a denylist (strings.ContainsAny(addr, "\r\n")) and then interpolates addr twice into the request line and the Host header. CR/LF is the only smuggling primitive that matters, so the guard is sufficient for the request-splitting case, and at the single call site in this PR (grpcDialer -> grpc.WithContextDialer) addr is the gRPC resolver's host:port so it is well-formed. It remains a denylist on a hand-built protocol line in an exported helper: a value containing SP/TAB, or an absolute-form URI, still reaches the wire and is interpreted by a lenient proxy (e.g. as a forward-proxy absolute-form request rather than a tunnel target). A positive shape check costs one line and converts the guard from "reject the known-bad byte" to "accept only host:port".

**Recommendation:** Add net.SplitHostPort(addr) (and a numeric-port check) alongside the existing CR/LF rejection, returning an error when it fails, so only a host:port literal can be interpolated into the CONNECT request line and Host header.

```
		if strings.ContainsAny(addr, "\r\n") {
			return nil, fmt.Errorf("httpx: invalid proxy target address %q: contains CR or LF", addr)
	req := fmt.Sprintf("CONNECT %s HTTP/1.1\r\nHost: %s\r\n\r\n", addr, addr)
```

</details>

<details>
<summary><code>SEC-BE-003</code> · low · <code>pkg/httpx/proxy_client.go:246</code></summary>

**Issue:** connectHandshake reads the proxy's CONNECT reply with http.ReadResponse over a bare bufio.NewReader(conn). Unlike the server side, net/http applies no MaxHeaderBytes on this path — textproto.ReadMIMEHeader is called without a limit — so the reply's status line and header block are read into memory until EOF. A proxy that answers a CONNECT with an unbounded header stream (a compromised/misbehaving proxy, or an attacker injecting into the plaintext leg when --proxy points at a non-loopback http-scheme proxy) can drive vespasian's heap arbitrarily. The only backstop today is the caller's context deadline propagated via conn.SetDeadline; with a cancel-only context there is no byte bound at all, only the ctx.Done() watcher. This mirrors what net/http's own Transport does for CONNECT, and the operator-controlled-proxy trust model makes it low risk, but it is cheap defense-in-depth on the one leg of this PR that parses attacker-reachable protocol bytes by hand.

**Recommendation:** Wrap the conn in an io.LimitReader (e.g. 64 KiB) for the handshake read before constructing the bufio.Reader, or read the reply through a bufio.Reader whose buffer is bounded and fail closed if the header block exceeds a fixed cap. Keep handing the caller the same reader afterwards so pipelined tunnel bytes are still preserved.

```
	br := bufio.NewReader(conn)
	resp, err := http.ReadResponse(br, &http.Request{Method: http.MethodConnect})
	if err != nil {
		return nil, fmt.Errorf("httpx: reading CONNECT response: %w", err)
```

</details>

</details>

<details>

<details>
<summary><code>QUAL-001</code> · low · <code>internal/pipeline/wsdl_probe.go:45</code></summary>

**Issue:** This PR adds two new, functionally-identical `CheckRedirect` helpers — `pipeline.noFollowRedirects` (internal/pipeline/wsdl_probe.go:45-47) and `probe.probeRedirectPolicy` (pkg/probe/types.go, added in this PR) — each doing nothing but `return http.ErrUseLastResponse`. Two more equivalent copies already exist in the packages this PR touches: `jsstatic.noFollowRedirects` (pkg/analyze/jsstatic/sourcemap.go:255) and `crawl.noRedirect` (pkg/crawl/jsreplay.go:179). All four packages already import the new shared leaf package `pkg/httpx` for `BuildHTTPClient`/`ProxyDialer`, so the redirect policy could be centralized there instead of continuing to copy-paste the same 2-line function (with three different names: noFollowRedirects, noRedirect, probeRedirectPolicy) into every package that builds an HTTP client.

**Recommendation:** Add a single exported `httpx.NoFollowRedirects` (or similar) `CheckRedirect` func and have `pkg/probe`, `internal/pipeline`, `pkg/analyze/jsstatic`, and `pkg/crawl` reference it instead of each defining/naming their own copy. Low priority — purely a maintainability/DRY nit, no behavioral difference.

```
func noFollowRedirects(*http.Request, []*http.Request) error {
	return http.ErrUseLastResponse
}
```

</details>

</details>

<details>
<summary>Tests (10)</summary>

<details>
<summary><code>TEST-001</code> · low · <code>cmd/vespasian/main_test.go:3403</code></summary>

**Issue:** TestParseProxyConfig_ValidatesAndParses covers all four branches of parseProxyConfig, but nothing covers the error branch of parseProxyConfigOrEmpty (cmd/vespasian/main.go:867-873), which swallows the parse error and returns the zero ProxyConfig. That branch is a fail-open: on an unparseable --proxy the post-crawl stages run with proxying disabled and send traffic direct instead of through the operator's intercepting proxy — precisely the leak LAB-4993 exists to close. Its doc comment argues the branch is unreachable because each command fail-fast validates earlier, but that reachability claim is itself untested, so a future refactor that moves or drops the fail-fast validation would turn a hard error into silent proxy bypass with no test failing. This helper is invoked at three production call sites (main.go:408, :709, and GenerateCmd.options).

**Recommendation:** Add a subtest asserting `parseProxyConfigOrEmpty("ftp://127.0.0.1:21", true)` returns a config with `Enabled() == false`, so the fail-open contract is pinned deliberately rather than incidentally.

```
func TestParseProxyConfig_ValidatesAndParses(t *testing.T) {
	t.Run("rejects unsupported scheme", func(t *testing.T) {
		_, err := parseProxyConfig("ftp://127.0.0.1:21", false)
```

</details>

<details>
<summary><code>TEST-002</code> · low · <code>cmd/vespasian/main_test.go:3465</code></summary>

**Issue:** TestGenerateCmd_Options_CarriesProxy and TestScanCmd_ScanOptions_CarriesProxy both stop at `require.NotNil(t, opts.Proxy.URL)`. TestBuildJSReplayConfig_SetsProxy likewise asserts only `cfg.Proxy.URL`. None of the three constructs the command with ProxyInsecure: true, so the security-relevant half of the CLI boundary — that --proxy-insecure reaches pipeline.Options.Proxy.Insecure, pipeline.ScanOptions.Proxy.Insecure and crawl.JSReplayConfig.Proxy.Insecure — is asserted nowhere. These tests were written specifically to close the CLI-boundary gap (their doc comments cite TestGRPCInsecureSkipVerify_ReachesOptions, which does test both the true and false cases), so the omission is inconsistent with the pattern they claim to mirror.

**Recommendation:** Set `ProxyInsecure: true` in each of the three tests and assert `opts.Proxy.Insecure` / `cfg.Proxy.Insecure` is true, plus a false case — matching the two-sided true/false assertions in TestGRPCInsecureSkipVerify_ReachesOptions.

```
	cmd := &GenerateCmd{Proxy: "http://127.0.0.1:8080"}
	require.NotNil(t, opts.Proxy.URL, "GenerateCmd.Proxy must reach pipeline.Options.Proxy")
	require.NotNil(t, opts.Proxy.URL, "scanOptions must carry the passed proxy into pipeline.ScanOptions.Proxy")
```

</details>

<details>
<summary><code>TEST-003</code> · low · <code>internal/pipeline/pipeline_test.go:457</code></summary>

**Issue:** TestClassifyProbeGenerate_ProxyAndAllowPrivate_ProxyWins has exactly the same inputs and assertions as TestClassifyProbeGenerate_ProxyReachesProbe 27 lines above it: same optionsOrigin, same newRecordingProxy(t, false), same Options{APIType: REST, Confidence: 0.5, Probe: true, AllowPrivate: true, Proxy: {URL: proxyURL}}, same require.NoError / require.NotEmpty / assert.NotZero(hits). Because ProxyReachesProbe already passes AllowPrivate:true, the second test cannot distinguish the precedence contract its doc comment claims to pin — it is a verbatim duplicate that doubles runtime and gives false confidence that a Proxy-vs-AllowPrivate interaction is covered. The genuinely uncovered combination is Proxy set with AllowPrivate=false (the SDK's configuration), where the URL-level validator must reject the loopback target and the proxy must never be contacted.

**Recommendation:** Either delete TestClassifyProbeGenerate_ProxyAndAllowPrivate_ProxyWins, or change it to the case that actually differs: AllowPrivate=false with Proxy set, asserting the loopback target is rejected and hits.Load() == 0 (the pattern already used by TestSourcemap_ProxiedFetch_AllowPrivateGate in pkg/analyze/jsstatic/sourcemap_test.go).

```
func TestClassifyProbeGenerate_ProxyAndAllowPrivate_ProxyWins(t *testing.T) {
		AllowPrivate: true, // private/loopback target — must still be reachable
	assert.NotZero(t, hits.Load(), "proxy must win over AllowPrivate's permissive-client branch")
```

</details>

<details>
<summary><code>TEST-004</code> · low · <code>internal/pipeline/wsdl_probe_test.go:364</code></summary>

**Issue:** TestProbeWSDLDocument_RoutesThroughProxy hand-rolls a forwarding recording proxy (counter, http.NewRequestWithContext on r.RequestURI, DefaultTransport.RoundTrip, WriteHeader, io.Copy) that is a near-verbatim copy of newRecordingProxy(t, true) added by this same PR at internal/pipeline/pipeline_test.go:381 — the same package (pipeline_test), so the helper is directly callable. The copy also silently diverges: newRecordingProxy(forwardBody=true) relays upstream response headers, this one relays only the body, so the two proxies behave differently for any assertion touching Content-Type. Two copies of a test proxy in one package means a fix to one (e.g. adding CONNECT support for https origins) will not reach the other.

**Recommendation:** Replace the inline proxy with `proxy, hits := newRecordingProxy(t, true)` and assert on `hits.Load()`, deleting the local `proxied` counter and handler.

```
func TestProbeWSDLDocument_RoutesThroughProxy(t *testing.T) {
	var proxied atomic.Int64
		resp, err := http.DefaultTransport.RoundTrip(outReq)
		io.Copy(w, resp.Body) //nolint:errcheck,gosec // test best-effort
```

</details>

<details>
<summary><code>TEST-005</code> · nit · <code>pkg/analyze/jsstatic/sourcemap_test.go:777</code></summary>

**Issue:** The doc comment on TestSourcemap_ProxiedFetch_AllowPrivateGate states "today NOTHING validates mapURL at the URL level for the proxied path" and "(the developer will add probe.ValidateProbeURL(mapURL) gated on !allowPrivate)". Both statements are false at head — pkg/analyze/jsstatic/sourcemap.go:104 calls `probe.ValidateProbeURL(mappingURL)` and the test itself asserts the rejection succeeds. Left in place, the comment tells a future auditor that the proxied sourcemap path has no URL-level SSRF gate, which is the opposite of what the test proves.

**Recommendation:** Rewrite as a present-tense regression guard: the proxied path relies on probe.ValidateProbeURL(mappingURL) at sourcemap.go:104 because BuildHTTPClient installs no dial-time pin; the recording proxy's zero hit count proves the validator runs before any network I/O. Drop "the developer will add".

```
// today NOTHING validates mapURL at the URL level for the proxied path. A
// private/loopback-host mapURL must still be rejected when AllowPrivate=false
// (the developer will add probe.ValidateProbeURL(mapURL) gated on
```

</details>

<details>
<summary><code>TEST-006</code> · low · <code>pkg/analyze/jsstatic/sourcemap_test.go:794</code></summary>

**Issue:** The same forwarding recording-proxy handler is written twice in this one file: inline at line 666 (TestRecoverSourcemap_RoutesThroughProxy) and again as a function-local closure `newRecordingProxy := func(t *testing.T) ...` at line 794 inside TestSourcemap_ProxiedFetch_AllowPrivateGate. The closure form is worse than a package-level helper: it cannot be reused by the two other proxy tests in the file, it shadows nothing but reads like the package-level helper of the same name that exists in internal/pipeline, and it returns a `stop func()` that duplicates what t.Cleanup already does elsewhere in this file. Divergence risk is concrete — the line-666 copy has no `stop` and relies on `defer origin.Close()` ordering.

**Recommendation:** Extract one package-level `newRecordingProxy(t *testing.T) (proxyURL *url.URL, hits *atomic.Int64)` helper in this file that registers its own t.Cleanup, and call it from both TestRecoverSourcemap_RoutesThroughProxy and TestSourcemap_ProxiedFetch_AllowPrivateGate.

```
	newRecordingProxy := func(t *testing.T) (proxyURL *url.URL, hits *atomic.Int64, stop func()) {
		var proxied atomic.Int64
			resp, err := http.DefaultTransport.RoundTrip(outReq)
```

</details>

<details>
<summary><code>TEST-007</code> · nit · <code>pkg/crawl/jsreplay_test.go:3173</code></summary>

**Issue:** TestReplayJSExtracted_RoutesThroughProxy inlines the same counter+RoundTrip+io.Copy forwarding proxy that already exists a few files away in the same package at pkg/crawl/http_crawler_test.go:205 (which its own comment cites as the model). Counting the new copies added by this PR — internal/pipeline/pipeline_test.go:381 (the one named helper), internal/pipeline/wsdl_probe_test.go:367, pkg/analyze/jsstatic/sourcemap_test.go:666 and :796, pkg/httpx/proxy_client_test.go:135, pkg/crawl/jsreplay_test.go:3173, plus pkg/probe/proxy_test.go:51 — the same ~18-line handler is now written seven times. Cross-package copies are unavoidable in Go, but the two within pkg/crawl are not.

**Recommendation:** Extract a `newRecordingProxy(t *testing.T) (*httptest.Server, *atomic.Int64)` helper in pkg/crawl (e.g. in a shared *_test.go) and use it from both jsreplay_test.go and http_crawler_test.go.

```
	var proxied atomic.Int64
		resp, err := http.DefaultTransport.RoundTrip(outReq)
	assert.NotZero(t, proxied.Load(), "JS-replay must route its fetches through the configured proxy")
```

</details>

<details>
<summary><code>TEST-008</code> · nit · <code>pkg/httpx/proxy_client_test.go:76</code></summary>

**Issue:** TestBuildHTTPClient_InsecureOnlyForHTTPScheme asserts `tr.TLSClientConfig.InsecureSkipVerify` is true but never asserts `tr.TLSClientConfig.MinVersion`. The explicit `MinVersion: tls.VersionTLS12` in proxy_client.go was added specifically to clear the GHAS G402 "TLS MinVersion too low" alerts, and it is the only thing keeping that scanner quiet on this hand-built tls.Config. Nothing in the suite fails if it is dropped, so the alert will silently re-fire on the next edit to that branch.

**Recommendation:** Add `assert.Equal(t, uint16(tls.VersionTLS12), tr.TLSClientConfig.MinVersion)` to the Insecure=true subtest.

```
			require.NotNil(t, tr.TLSClientConfig,
			assert.True(t, tr.TLSClientConfig.InsecureSkipVerify,
				"InsecureSkipVerify must be true for a %s proxy with Insecure=true", scheme)
```

</details>

<details>
<summary><code>TEST-009</code> · low · <code>pkg/httpx/proxy_client_test.go:415</code></summary>

**Issue:** Three new test doc comments describe the production code as still unguarded, which is false at head and actively misleads the next reader of this security-sensitive file. (a) Line 415-427 states connectDialer builds the CONNECT line "with no validation that addr is free of \r\n" and that "Once addr validation is added, ProxyDialer must reject the target" — the guard shipped in this same PR at pkg/httpx/proxy_client.go:140 (`if strings.ContainsAny(addr, "\r\n")`). (b) Line 459 states the CONNECT response read "is never wired to ctx — no SetReadDeadline derived from ctx, no goroutine closing conn on ctx.Done()" — proxy_client.go:167 has the ctx.Done() watcher. (c) Line 671-672 states "connectDialer only calls conn.SetDeadline(ctx.Deadline()) — a deadline-based bound" and that nothing unblocks a cancel-only context — also fixed by the same watcher. A future reader auditing whether CRLF injection is possible will read these comments as current fact and conclude the vulnerability is live.

**Recommendation:** Rewrite the three comments in the present tense as regression guards describing the shipped invariant (e.g. "connectDialer rejects CR/LF in addr before writing the CONNECT request line (proxy_client.go:140); the naive stub proxy accepts anything, so a nil error here would mean the guard was removed"), and drop the future-tense "once ... is added" / "the developer will" phrasing.

```
// fmt.Sprintf("CONNECT %s HTTP/1.1\r\nHost: %s\r\n\r\n", addr, addr) with no
// validation that addr is free of "\r\n". An addr containing CRLF lets an
// exists. Once addr validation is added, ProxyDialer must reject the target
// the CONNECT response (http.ReadResponse(br, ...)) is never wired to ctx —
// for TEST-001/SEC-BE-002: connectDialer only calls
// conn.SetDeadline(ctx.Deadline()) — a deadline-based bound. A cancel-only
```

</details>

<details>
<summary><code>TEST-010</code> · low · <code>pkg/httpx/proxy_client_test.go:770</code></summary>

**Issue:** Three of the new stub proxies manage connection lifetime with wall-clock sleeps or unbounded blocks instead of test-scoped signals. (a) Line 770: `time.Sleep(200 * time.Millisecond)` is used as synchronization — the comment says "Keep the connection open briefly so the client can read before we exit" — so TestProxyDialer_ConnectPreservesPipelinedBytes flakes on a loaded CI box if the client's dial + type assertion + io.ReadFull does not complete inside 200ms. (b) Line 697: `<-make(chan struct{})` blocks that goroutine forever; its `defer conn.Close()` never runs, so the goroutine and its accepted fd leak for the lifetime of the test binary. (c) Line 483: `time.Sleep(5 * time.Second)` keeps a goroutine alive ~4.7s after the 300ms test finishes. None of these are the assertion under test, so they are pure flake/leak surface in a package that otherwise reaches for deterministic signals (see the `accepted` channel at line 683).

**Recommendation:** Replace all three with a `stop := make(chan struct{})` that the test closes via t.Cleanup, and have the stub goroutines block on `<-stop` (and select on it) instead of sleeping or blocking on a channel nobody closes. That removes the 200ms race, the leaked fd, and the 5s straggler while keeping the same behavior.

```
		// Keep the connection open briefly so the client can read before we exit.
		time.Sleep(200 * time.Millisecond)
		// Stay silent forever; never write a CONNECT response status line.
		<-make(chan struct{})
		time.Sleep(5 * time.Second)
```

</details>

</details>

<details>

<details>
<summary><code>DOC-001</code> · nit · <code>internal/pipeline/wsdl_probe.go:86</code></summary>

**Issue:** Three exported helpers changed signature to take a new `proxy httpx.ProxyConfig` argument -- ProbeWSDLDocument (L86), ProbeAndAppendWSDLRequest (L154) and ResolveWSDLType (L190). None of their godoc blocks mention the parameter, even though each block documents its other arguments in prose (ProbeWSDLDocument explicitly explains `status is an optional io.Writer for progress messages; pass nil or io.Discard to suppress them`). A caller reading godoc alone cannot tell that passing the zero ProxyConfig means unproxied, nor that the proxied path drops the dial-time SSRF pin while keeping URL-level validation -- a security-relevant nuance that every other new exported Proxy field in this PR (probe.Config.Proxy, crawl.JSReplayConfig.Proxy, jsstatic.Options.Proxy, pipeline.Options.Proxy, pipeline.AugmentOptions.Proxy) does document inline. Impact is limited to internal/pipeline, which is not importable outside the module.

**Recommendation:** Add one sentence to each of the three godoc blocks stating that proxy routes the WSDL fetch through an intercepting proxy when set, that the zero value means unproxied, and that the proxied client carries no dial-time SSRF pin (URL-level validation unchanged) -- mirroring the wording already used on probe.Config.Proxy.

```
// for progress messages; pass nil or io.Discard to suppress them.
func ProbeWSDLDocument(ctx context.Context, targetURL string, allowPrivate bool, proxy httpx.ProxyConfig, status io.Writer) []byte {
```

</details>

</details>

</details>

---

santiago-praetorian (changes requested):
REQUEST CHANGES — 3 blockers (all medium), round 4 at head 441bec7. Full roll-up: https://github.com/praetorian-inc/vespasian/pull/186#issuecomment-5094499648

All three LAB-4993 acceptance criteria verified MET against the code rather than the PR description — proxy wiring reaches every ticket-named stage, the `--proxy-insecure` gate is the same predicate as the crawl path, socks5 keeps verification everywhere, and no stale "probe and JS-replay traffic is not proxied" wording survives in tracked source (4 occurrences at base, 0 at head). Zero REQ findings.

The three blockers cluster on one theme: **`--proxy-insecure`'s real blast radius is now wider than what's documented or separately opted into.**

- **SEC-BE-001** (medium, `cmd/vespasian/main.go:709`) — the shared `--proxy-insecure` help still scopes itself to the `--headless=false` net/http backend, but threading `ProxyConfig` into the post-crawl stages means it now disables TLS verification on probe, JS-replay and sourcemap fetches regardless of headless mode. Help text and README understate the reach.
- **SEC-BE-004** (medium, `pkg/probe/grpc.go:241`) — `dialGRPC` lets `--proxy-insecure` implicitly override `--grpc-insecure-skip-verify`, collapsing a separate explicit opt-in into an unrelated flag. The more substantive of the pair.
- **TEST-011** (medium, missing-tests, `pkg/probe/types.go:152`) — `Insecure: true` is only ever asserted at the `httpx.BuildHTTPClient` leaf, so nothing pins that the bit survives the `withDefaults` → `BuildHTTPClient` hop. A regression reconstructing `ProxyConfig{URL: ...}` without `Insecure` would leave the suite green while silently breaking Burp interception — and AC #2 is precisely the criterion that requires this to hold.

14 non-blockers (2 security, 1 quality, 10 test, 1 doc) are in the roll-up: an unbounded `http.ReadResponse` on the CONNECT reply, an untested fail-open branch in `parseProxyConfigOrEmpty`, a vacuous precedence test (`ProxyWins` is byte-identical to `ProxyReachesProbe`), stale RED-phase comments asserting shipped guards don't exist, and further recording-proxy duplication.

One caveat on the DRY findings (QUAL-001, TEST-004/006/007): the Rule of Three is met on the counts, but readable test setup legitimately outweighs strict DRY — feel free to reject TEST-007 on test-clarity grounds rather than treating it as a defect. All four are non-blocking.

Prior-round findings (QUAL-001/002, SEC-BE-001, TEST-002 from round 2/3, plus the resolved gosec/GHAS threads) were excluded as already addressed and are not re-reported.

santiago-praetorian:

**PR:** https://github.com/praetorian-inc/vespasian/pull/186
**Linear:** https://linear.app/praetorianlabs/issue/LAB-4993/vespasian-probe-and-js-replay-stages-ignore-proxy-only-crawl-traffic
**Head SHA:** ca5e875890e43b5fe7904f6530e548b761fa1746
**Reviewed at:** 2026-07-28T22:41:00Z

|---|---|---|---|---|---|
| Count | 0 | 0 | 1 | 5 | 2 |

---

<details>

<details>
<summary><code>TEST-002</code> · medium · TEST · <code>pkg/crawl/jsreplay_test.go:3212</code></summary>

**Issue:** JS-replay is the only proxy-aware stage in this PR whose proxied path has no AllowPrivate=false test. withDefaults' new proxied branch builds the client via httpx.BuildHTTPClient, which deliberately installs no dial-time SSRF pin, so canFetchURL's ssrf.ValidateURLContext (pkg/crawl/jsreplay.go) becomes the ONLY remaining SSRF guard for replay fetches/probes when proxied. Every added proxy test for this stage passes AllowPrivate: true (grep of 'Proxy' in jsreplay_test.go shows only withDefaults transport-shape tests plus TestReplayJSExtracted_RoutesThroughProxy with AllowPrivate: true), so a regression that skipped URL validation when a proxy is configured -- the exact 'the proxy relaxes SSRF' mistake the proxied branch invites -- would leave the suite green. Both sibling stages got this coverage: pkg/analyze/jsstatic/sourcemap_test.go's TestSourcemap_ProxiedFetch_AllowPrivateGate and internal/pipeline/pipeline_test.go's TestClassifyProbeGenerate_ProxiedLoopback_RequiresAllowPrivate each assert the recording proxy sees ZERO hits when AllowPrivate=false, which is what proves the URL gate runs before any network I/O.

**Recommendation:** Add a TestReplayJSExtracted_ProxiedFetch_AllowPrivateGate mirroring TestSourcemap_ProxiedFetch_AllowPrivateGate: same loopback origin + newRecordingProxy(t), Client left nil, Proxy set, AllowPrivate: false. Assert (a) no js-extract entries are appended and (b) hits.Load() == 0 -- the proxy must never be contacted for a private/loopback target. Keep the existing AllowPrivate: true case as the positive half.

```
func TestReplayJSExtracted_RoutesThroughProxy(t *testing.T) {
	cfg := JSReplayConfig{
		Proxy:        httpx.ProxyConfig{URL: proxyURL},
		AllowPrivate: true,
	assert.NotZero(t, hits.Load(), "JS-replay must route its fetches through the configured proxy")
```

</details>

<details>
<summary><code>TEST-003</code> · low · TEST · <code>pkg/httpx/proxy_client.go:313</code></summary>

**Issue:** connectHandshake documents and implements a deliberate behavior -- conn.SetDeadline(time.Time{}) after a successful CONNECT reply, 'so it does not apply to tunneled traffic' -- that no test in pkg/httpx/proxy_client_test.go exercises. TestProxyDialer_HTTPConnectTunnel and TestProxyDialer_HTTPSProxy_ConnectTunnel both dial with a 5s ctx deadline and then read/write immediately, so they pass identically whether or not the deadline is cleared. If the clear were dropped (or moved before the pipelined-bytes read), every proxied tunnel would silently die at the caller's ctx deadline; for the gRPC reflection dial that ctx carries p.config.Timeout, so a reflection stream through a proxy would be cut short with no failing test. This is the one documented invariant of the new CONNECT path with no coverage.

**Recommendation:** Add a deterministic test using startRecordingCONNECTProxy + startTCPEchoServer: dial with a short-deadline ctx (e.g. 300ms), require the dial succeeds, then block on <-ctx.Done() (no sleep) and afterwards Write/ReadFull through the returned conn, requiring both to succeed. That fails with an i/o timeout if the handshake deadline ever leaks into the tunnel.

```
	// Clear the handshake deadline so it does not apply to tunneled traffic.
	if err := conn.SetDeadline(time.Time{}); err != nil {
		return nil, fmt.Errorf("httpx: clearing proxy CONNECT deadline: %w", err)
```

</details>

</details>

<details>

<details>
<summary>Security (2)</summary>

<details>
<summary><code>SEC-BE-001</code> · nit · <code>pkg/probe/grpc.go:297</code></summary>

**Issue:** proxyTLSVerifyMismatch gates only on t.useTLS && Proxy.Enabled() && Proxy.Insecure && !GRPCInsecureSkipVerify — it never inspects Proxy.URL.Scheme. Every other consumer of ProxyConfig.Insecure in this PR is scheme-gated (httpx.BuildHTTPClient at pkg/httpx/proxy_client.go:114 checks `p.URL.Scheme == "http" || p.URL.Scheme == "https"`; httpx.dialProxy at :327 only honors Insecure on the https branch; the socks5 branch of ProxyDialer at :140-151 ignores Insecure entirely). Consequently, with `--proxy socks5://127.0.0.1:1080 --proxy-insecure` against a TLS gRPC target, Proxy.Insecure is completely inert — SOCKS5 is a transparent TCP tunnel, so the gRPC dial performs TLS directly against the real target and its genuine certificate verifies normally — yet probeTarget (grpc.go:222-224) still emits WARN 'a TLS gRPC target behind an intercepting proxy will fail verification — pass --grpc-insecure-skip-verify to enumerate it'. The warning is a false positive that instructs the operator to disable gRPC target-certificate verification in a configuration where verification would have succeeded; acting on it turns an authenticated reflection exchange into an unauthenticated one, letting an on-path attacker between the SOCKS5 proxy and the target substitute a certificate and inject forged FileDescriptorProtos into the generated .proto. The new table test TestGRPCProbe_ProxyTLSVerifyMismatch (pkg/probe/proxy_internal_test.go:7026-7049 in the diff) covers only an http-scheme proxy, so no test pins the socks5 case. Note this is advisory text only — the code invariant itself is correct (socks5 always verifies).

**Recommendation:** Add the scheme gate to proxyTLSVerifyMismatch so it matches the rest of the ProxyConfig.Insecure handling, e.g. `s := p.config.Proxy.URL.Scheme; return t.useTLS && p.config.Proxy.Enabled() && p.config.Proxy.Insecure && (s == "http" || s == "https") && !p.config.GRPCInsecureSkipVerify`. Add a `socks5 proxy insecure is not a mismatch` row (want: false) to TestGRPCProbe_ProxyTLSVerifyMismatch so the gate is pinned.

```
func (p *GRPCProbe) proxyTLSVerifyMismatch(t grpcTargetInfo) bool {
	return t.useTLS && p.config.Proxy.Enabled() && p.config.Proxy.Insecure && !p.config.GRPCInsecureSkipVerify
		slog.WarnContext(ctx, "grpc probe: --proxy-insecure does not relax gRPC target-cert verification; a TLS gRPC target behind an intercepting proxy will fail verification — pass --grpc-insecure-skip-verify to enumerate it", "target", t.hostPort)
```

</details>

<details>
<summary><code>SEC-BE-002</code> · nit · <code>pkg/probe/types.go:167</code></summary>

**Issue:** withDefaults warns 'probe: Proxy configured but ignored — an injected Client owns its transport; probe traffic will BYPASS the proxy' whenever Config.Client is non-nil and Config.Proxy is enabled. The claim is over-broad: withDefaults leaves cfg.Proxy populated, GRPCProbe stores the withDefaults result (pkg/probe/grpc.go:72 `return &GRPCProbe{config: cfg.withDefaults()}`), and grpcDialer (grpc.go:307-314) branches solely on `p.config.Proxy.Enabled()`. So in that configuration the five HTTP strategies go direct while the gRPC reflection dial still tunnels through the proxy and — for an https-scheme proxy — still honors Proxy.Insecure on the CONNECT leg (httpx.dialProxy, proxy_client.go:326-339). An operator or integrator reading the warning would conclude no probe traffic reaches the proxy and would not expect a CONNECT tunnel (or a proxy-cert-verification opt-out) on the gRPC stage, i.e. the security control's actual blast radius is wider than the warning states — the same class of defect as the already-addressed SEC-BE-001 help-text finding. Reachability note: the only in-repo production constructor, internal/pipeline.buildProbeConfig (pipeline.go), deliberately leaves cfg.Client nil when proxied, so today this configuration is reached only by tests (pkg/probe/proxy_internal_test.go:116) and by external library callers of the exported probe.Config.

**Recommendation:** Scope the message to the stages it actually describes, e.g. 'an injected Client owns its transport; the HTTP probe strategies will BYPASS the proxy (the gRPC reflection dial still tunnels through it)'. Alternatively make grpcDialer honor the same opt-out (skip the proxy dialer when a Client was injected) so both halves of Config behave consistently, and assert the chosen semantics in TestConfig_WithDefaults_WarnsWhenClientInjectedWithProxy.

```
	} else if cfg.Proxy.Enabled() {
		// SEC-BE-004: an injected Client owns its transport, so a configured Proxy
		// is silently ignored here. Warn loudly rather than bypass the proxy
		slog.Warn("probe: Proxy configured but ignored — an injected Client owns its transport; probe traffic will BYPASS the proxy")
```

</details>

</details>

<details>
<summary>Tests (3)</summary>

<details>
<summary><code>TEST-001</code> · low · <code>internal/pipeline/augment_test.go:378</code></summary>

**Issue:** The require.Greater message claims the enriched slice contains 'sourcemap-recovered endpoints', but nothing in this test can produce one. The helper is called as newRecordingProxy(t, false), and with forwardBody=false the pipeline_test.go helper relays only w.WriteHeader(resp.StatusCode) -- response headers and body are NOT copied -- so the .js.map body never reaches the client, json.Unmarshal fails, and stats.SourcemapsRecovered stays 0 (stats.SourcemapFetchFails is incremented instead). The assertion passes solely because jsluice extracts /api/x from the bundle body's own fetch("/api/x") literal supplied in the capture. Even if the body were forwarded, mapBody's sourcesContent is "export const x = 1;", which contains no API path, so the length assertion could still never distinguish sourcemap recovery from plain bundle extraction. The hits.Load() assertion on the next line is the only part that actually proves AugmentOptions.Proxy reached jsstatic.Options.Proxy; the length assertion is a false signal a future reader will trust.

**Recommendation:** Switch to newRecordingProxy(t, true) so the map body is relayed, give the sourcemap a sourcesContent containing a distinct API path (e.g. fetch("/api/from-sourcemap")), and assert a request with Source == crawl.SourceStaticJSSourcemap (or the specific recovered URL) is present in out. Otherwise drop the misleading message and assert only what the test proves.

```
	proxy, hits := newRecordingProxy(t, false)
	mapBody := []byte(`{"version":3,"sources":["app.ts"],"sourcesContent":["export const x = 1;"]}`)
	require.Greater(t, len(out), len(captured), "expected enriched slice with sourcemap-recovered endpoints")
	assert.NotZero(t, hits.Load(), "sourcemap fetch must route through the configured proxy")
```

</details>

<details>
<summary><code>TEST-004</code> · low · <code>pkg/probe/proxy_internal_test.go:41</code></summary>

**Issue:** TestConfig_WithDefaults_ProxyClient asserts Transport.Proxy, the absent DialContext, the redirect policy and the Insecure gate, but never asserts a timeout -- the word 'Timeout' does not appear anywhere in this file. That matters here specifically: the non-proxy branch of withDefaults uses DefaultTransport(), which carries per-stage TLSHandshakeTimeout/ResponseHeaderTimeout bounds, while the proxied branch (httpx.BuildHTTPClient) clones DefaultTransport, clears DialContext and sets no stage caps, so http.Client.Timeout is the ONLY thing bounding a proxied probe request. pkg/probe/types.go's own comment states this ('A client-level Timeout stands in for the stage timeouts that DefaultTransport() carries'). A refactor that passed 0, or that reordered withDefaults so cfg.Timeout is still zero at client-construction time, would produce a proxied probe client with no timeout and no stage caps -- an unbounded hang against a slow proxy -- and every test in this file would still pass. The WSDL stage received exactly this guard (internal/pipeline/wsdl_probe_internal_test.go's 'proxied client carries the per-stage caps' subtest); the probe stage did not.

**Recommendation:** In TestConfig_WithDefaults_ProxyClient, build the config with an explicit non-zero Timeout and assert cfg.Client.Timeout equals it, plus a second case with Timeout unset asserting cfg.Client.Timeout equals the withDefaults default (non-zero). Consider the same one-line assertion in pkg/crawl/jsreplay_test.go's TestJSReplayConfig_WithDefaults_ProxyClient for parity.

```
	cfg := Config{Proxy: httpx.ProxyConfig{URL: proxyURL}}.withDefaults()
	assert.NotNil(t, tr.Proxy, "proxied client must set Transport.Proxy")
	assert.Nil(t, tr.DialContext, "proxied client must NOT install the SSRF dial pin (no target pin when proxied)")
```

</details>

<details>
<summary><code>TEST-005</code> · low · <code>pkg/probe/proxy_test.go:83</code></summary>

**Issue:** TestProbe_OptionsStrategy_RoutesThroughProxy throws away the enriched endpoints (the result is assigned to _) and asserts only that the proxy counter moved. Compounding this, the inline test proxy relays just w.WriteHeader(resp.StatusCode) and never copies response headers, so the origin's 'Allow: GET, POST' cannot reach the probe -- meaning the test structurally cannot verify that a proxied OPTIONS probe still produces usable output. A regression where the proxied client routes correctly but the response is mangled (headers dropped, wrong method, body truncated) would keep this test green. Every other stage's end-to-end proxy test asserts a real payload alongside the hit counter (WSDL asserts the document contains 'Calculator', jsstatic asserts the recovered source string, JS-replay asserts the discovered API URL, pipeline asserts a non-empty spec); the probe stage's own end-to-end test is the outlier.

**Recommendation:** Copy resp.Header into w.Header() before WriteHeader in the test proxy handler, then assert on the enriched result: capture 'enriched, err := p.Probe(...)' and require enriched[0].AllowedMethods to equal []string{"GET", "POST"} in addition to the existing proxied.Load() assertion.

```
	_, err = p.Probe(context.Background(), endpoints)
	require.NoError(t, err)
	assert.NotZero(t, proxied.Load(), "OPTIONS strategy must route its request through the configured proxy")
		w.WriteHeader(resp.StatusCode)
```

</details>

</details>

<details>

<details>
<summary><code>DOC-001</code> · low · <code>CLAUDE.md:75</code></summary>

**Issue:** This PR introduces a new public leaf package, pkg/httpx (ProxyConfig, BuildHTTPClient, ProxyDialer, NoFollowRedirects), and CLAUDE.md carries a package inventory at lines 73-89 with exactly one bullet per package. Every other directory under pkg/ (analyze, classify, crawl, generate, importer, mediatype, probe, sdk, ssrf) plus cmd/vespasian, internal/pipeline and internal/grpcwire has its own bullet; pkg/httpx is the only one that does not. The single-line CLAUDE.md edit in this PR appended the LAB-4993 proxy behavior to the existing pkg/crawl bullet and mentions pkg/httpx.BuildHTTPClient/ProxyDialer only inline inside that sentence. CLAUDE.md is the agent-facing map of the codebase, so a new package that owns the proxy-client contract for five stages is discoverable only by reading the pkg/crawl bullet to its end — the next contributor (human or agent) looking for where proxy clients are built will not find it in the inventory.

**Recommendation:** Add a `- **pkg/httpx**: ...` bullet to the inventory (adjacent to pkg/ssrf, since both are stdlib-only leaf packages) stating that it builds proxy-aware *http.Client values and CONNECT/SOCKS5 dialers for every post-crawl stage, imports only golang.org/x/net/proxy beyond stdlib, and deliberately depends on none of crawl/probe/pipeline/ssrf to stay cycle-free.

```
- **pkg/ssrf**: SSRF protection — blocks private IP ranges, cloud metadata endpoints, and DNS rebinding
- **pkg/analyze**: JavaScript endpoint extraction from captured JS bundles
```

</details>

</details>

</details>

---

santiago-praetorian (changes requested):
REQUEST CHANGES — 2 blockers, round 7 at head `ca5e875`. Full roll-up: https://github.com/praetorian-inc/vespasian/pull/186#issuecomment-5110605831

All three LAB-4993 acceptance criteria verified MET against the code rather than the PR description: the flag is parsed once per command and threaded into every ticket-named stage, the `--proxy-insecure` predicate is textually identical to the crawl path's (`pkg/httpx/proxy_client.go:114` vs `pkg/crawl/http_crawler.go:79`), socks5 always verifies, and no stale "probe and JS-replay traffic is not proxied" text survives anywhere in the tree. Both design constraints (PF1/PF2) are satisfied. Zero REQ findings, zero QUAL findings.

**Blockers**

- `TEST-002` (medium) — `pkg/crawl/jsreplay_test.go`: JS-replay is the only proxy-aware stage whose proxied path has no `AllowPrivate=false` test. All 29 `AllowPrivate` references in the file are `true`. Because the proxied client drops the dial-time SSRF pin, `canFetchURL`'s `ssrf.ValidateURLContext` is the sole remaining guard there — and both sibling stages got exactly this coverage (`TestSourcemap_ProxiedFetch_AllowPrivateGate`, `TestClassifyProbeGenerate_ProxiedLoopback_RequiresAllowPrivate`), each asserting the recording proxy sees zero hits. This is the substantive gap.
- `TEST-003` (low) — `pkg/httpx/proxy_client.go:313`: the documented `conn.SetDeadline(time.Time{})` clear has no coverage. Verified by mutation: deleting the clear entirely leaves `go test ./pkg/httpx/` green.

**Non-blockers** — two `nit` advisory-text accuracy items (`SEC-BE-001` socks5 false-positive warning, `SEC-BE-002` over-broad bypass warning), three `low` items forming one cluster (*three new proxy tests assert less than they appear to*: a misleading assertion message, an unasserted `Client.Timeout`, a discarded probe result), and one `low` doc gap (`pkg/httpx` is the only `pkg/*` directory without a `CLAUDE.md` inventory bullet).

**Two calibration notes on my own findings, so you can weigh them fairly:**

1. `SEC-BE-002` is arguably out of scope by my own review rules. Its text concedes the configuration is reachable only from tests and external library callers — `pipeline.buildProbeConfig` leaves `Client` nil whenever a proxy is set. The wording inaccuracy is real, but no in-repo caller reaches it. Reasonable to close as won't-fix.
2. `TEST-003` blocks on a technicality: any `missing_tests` finding is promoted to blocker regardless of severity. A regression here would hang a tunnel, not create a security hole. If you'd rather land this and file the test as follow-up, I won't re-raise it.

`TEST-002` is the one I'd genuinely hold the PR for. Every other finding is cheap and non-urgent.

eli-wald:
## Adjudication — LAB-4993 / QUAL-001 @ head `6d417c42`

**Change under review:** deduplicate the three byte-identical `augmentOptions` builders in `cmd/vespasian/main.go` into a package-level `buildAugmentOptions(...)` helper that each command's `augmentOptions` delegates to (QUAL-001), plus a one-word doc-comment spelling fix (`Centralised` → `Centralized`). Behavior-preserving; the three CLI-boundary proxy tests (`TestCrawlCmd/GenerateCmd/ScanCmd_AugmentOptions_CarriesProxy`) pass unchanged.

**Premise gate (`reviewing-the-premise`): SKIP — recorded.** The trigger is "builds non-trivial machinery"; this is a single-file, behavior-preserving extract-helper refactor (no new layer, abstraction, or subsystem) plus a spelling fix. The complement condition — no non-trivial machinery — applies, so the gate does not fire.

**CI @ `6d417c42`: green.** `ci/{Build, Lint, Test, Module integrity, preflight}`, `security/{Gosec, Govulncheck}`, `scan/Titus Secrets Scan`, `preflight-selftest` all pass; `drift-check` skips by repo config.

**Reviewer coverage @ `6d417c42`:**

| Reviewer | Kind | State | Evidence |
| --- | --- | --- | --- |
| CodeRabbit | GitHub App | substantive (clean) | head commit status `CodeRabbit = success`, "Review completed"; no new inline/review findings on the incremental diff |
| GitHub Advanced Security | GitHub App | substantive (clean) | head check-runs `gosec` / `govulncheck` / `titus` all `success` — "No new alerts in code changed by this pull request" |
| codex-code | Workflow | not-applicable (fork) | delegates to the hardened `public-workflows` reusable workflow; same-repo guard false on fork head `eli-wald/vespasian`; zero runs associated with this PR |
| gemini-code | Workflow | not-applicable (fork) | caller gate `if: head.repo.full_name == github.repository` false on the fork head; zero runs associated with this PR |
| claude-code | Workflow | not-applicable (fork) | delegates to the hardened `public-workflows` reusable workflow; same-repo guard false on the fork head; zero runs associated with this PR |

The three in-repo AI review workflows are structurally excluded on fork PRs and additionally trigger only on `opened` / `ready_for_review`, never `synchronize`, so a push never re-dispatches them. The `chatgpt-codex-connector[bot]` review dated 2026-07-21 is the standalone Codex GitHub App, anchored to the PR's early commit `1aa82cf`, and is superseded.

**Findings adjudicated @ head: none open.** No automated reviewer produced a finding anchored to `ccd8937` / `6d417c42`. The historical GHAS gosec and Codex-connector notes are all anchored to the PR's pre-`7f24a16` state and are superseded — GHAS reports "No new alerts in code changed by this pull request" at the current head.

**Reviewer items collected through:** 2026-08-04T23:37:47Z

Adjudication complete; CI is green and every reviewer at the current head is accounted for. Merge remains a human decision — I am not merging.

santiago-praetorian:
## `capability-pr-review` — head `6d417c42`

**Recommendation: REQUEST CHANGES** → to be resolved by **deferral**, see below.

|---|---|---|---|---|---|
| Count | 0 | 0 | 4 | 19 | 4 |
| Blockers | 0 | 0 | 4 | 0 | 0 |

Exhaustive sweep: 20 agent dispatches (4 passes each across quality / backend-security / test-lead), ~18 code mutations, every production and test file in the PR read at least once. `frontend-security` not dispatched (no frontend files).

**No `critical` or `high` findings, and no functional bug.** The proxy implementation itself verified correct under adversarial review: socks5 always verifies TLS (and that guarantee is pinned — deleting the scheme gate fails 3 packages), there is no silent direct-connect fallback on any of the five newly-proxied stages, the SSRF/origin gates compose correctly, and the CONNECT tunnel holds up against a hostile proxy. All three LAB-4993 acceptance criteria were verified against the code rather than the PR description: AC1 and AC2 **MET**, AC3 **partially met** (see REQ-002).

### Deferring the four blockers

The following four are agreed to be **deferred to a follow-up Linear ticket** (should be created under LAB-41). None is a functional defect; three are documentation or assertion-strength gaps.

| ID | Sev | Location | Issue |
|---|---|---|---|
| `DOC-001` | medium | `docs/grpc.md:115-124` | The "SSRF and TLS" section still states the gRPC dial target is "re-checked at connect time … closing the DNS-rebinding TOCTOU window" and that "SSRF is still enforced by the dialer regardless". Both were true at base, but `grpcDialer` (`pkg/probe/grpc.go:315`) now returns `httpx.ProxyDialer` and never reaches `p.config.Dialer` when `--proxy` is set. Same stale claim also appears at `README.md:335-336` and `pkg/probe/doc.go:46-51` — one fix covers all three surfaces. |
| `REQ-002` | medium | `pkg/probe/doc.go`, `internal/pipeline/doc.go`, `pkg/analyze/jsstatic/doc.go` | AC3 shortfall. Each package gained public proxy surface (`probe.Config.Proxy`; three exported pipeline signatures now taking `proxy httpx.ProxyConfig`; `jsstatic.Options.Proxy`) but none has a `doc.go` update. `CONTRIBUTING.md:312/328` lists this as a PR checklist item, and the PR does follow it for `pkg/crawl/doc.go` and `pkg/httpx/doc.go`. |
| `TEST-006` | medium | `internal/pipeline/wsdl_probe_internal_test.go` | The proxied WSDL client's no-follow-redirect guard (`wsdl_probe.go:59`) is **deletable**: setting the third argument to `nil` leaves `go test ./...` green repo-wide, while all three sibling stages kill the identical mutation. That policy is the SSRF-via-redirect backstop for the proxied path, which carries no dial-time SSRF pin. |
| `TEST-002` | medium | `cmd/vespasian/main.go` | The CLI carriage tests (`Test{Crawl,Generate,Scan}Cmd_*_CarriesProxy`) construct the `httpx.ProxyConfig` in the test body and pass it to the helper, so they pin only that the helper copies its own argument. The six `Run()` call sites that actually supply it are mutation-survivable. |

**Next step:** a follow-up Linear ticket covering these four, linked here. Once it is created and referenced in this PR, all four become non-blocking under the review's blocker formula (`severity ∈ {critical, high, medium} AND no follow-up ticket`), and the recommendation moves to APPROVE.

### Caveats worth recording

- **This audit is incomplete.** The quality and test lanes hit the 4-pass cap without two consecutive dry passes, so no convergence watermark was written and a future run will repeat the full sweep. Backend-security did converge.
- **Two findings were retracted after a premise check**, both with disproven mechanisms: a claim that `bufferedConn` promotes `WriteTo`/`ReadFrom` (it embeds the `net.Conn` *interface*, so no such promotion occurs — verified by type assertion and end-to-end byte check), and a claim about a test-file naming convention that 14 of 93 test files in this repo do not follow.
- **This run is not a superset of the previous one.** Relative to the round-8 review, 24 findings were previously missed, but 8 of that round's findings are absent here — 6 of them in code unchanged since. Read the two rounds together.
- The nine resolved GitHub Advanced Security threads on `pkg/httpx/proxy_client.go` all carry `isOutdated=true`, so their line anchors point at a superseded revision and were not used to suppress anything.

The remaining 23 findings are `low`/`nit` and non-blocking — mostly assertion-strength improvements and three cyclomatic-complexity measurements just over a threshold of 10.

<sub>Generated by <code>capability-pr-review</code> against head <code>6d417c422e7f058f543d81388fa3ca9fa45f39e9</code> and LAB-4993.</sub>

eli-wald:
Follow-up ticket for the four deferred `medium` blockers: **[LAB-5610](https://linear.app/praetorianlabs/issue/LAB-5610)** — *vespasian --proxy follow-ups: doc drift + doc.go coverage + proxied-path test hardening*.

It captures all four, each with its location and an acceptance-criterion checkbox:

- **DOC-001** — stale SSRF/TLS docs for the proxied gRPC dial path (`docs/grpc.md`, `README.md`, `pkg/probe/doc.go`)
- **REQ-002** — missing `doc.go` proxy-surface updates (`pkg/probe`, `internal/pipeline`, `pkg/analyze/jsstatic`); closes LAB-4993 AC3
- **TEST-006** — untested proxied-WSDL no-follow-redirect guard (`wsdl_probe.go:59`)
- **TEST-002** — mutation-survivable proxy carriage tests in `cmd/vespasian/main.go`

Per the review's blocker formula (`severity in {critical, high, medium} AND no follow-up ticket`), all four now have a linked follow-up and are non-blocking — this should move the recommendation to **APPROVE**.

Filed under epic LAB-41, sibling of LAB-4993 (this PR's ticket).

santiago-praetorian (approved):
Approved

PR #187, opened 2026-07-22
test(live-tests): harden coverage & assertion gaps (LAB-3890)

eli-wald:
## Summary

Hardens the live-test suite (`test/run-live-tests.sh` + friends) by closing real coverage holes and replacing assertions that were too soft to fail. Implements **T1–T4** of [LAB-3890](https://linear.app/praetorianlabs/issue/LAB-3890/vespasian-harden-live-test-suite-coverage-and-assertion-gaps).

The deterministic layer (byte-exact spec comparison) and importer breadth were already strong; this PR targets the gaps below.

## T1 — Real-parser validators (gap B1)

The three "structure" validators used grep/substring matching, so a malformed spec that merely *contained* the right strings passed:

| Validator | Before | After |
|-----------|--------|-------|
| `validate_openapi_structure` | greps for `openapi:`/`info:`/`paths:` | `@apidevtools/swagger-parser` — OpenAPI 3.0 schema + `$ref`/duplicate-`operationId` checks |
| `validate_soap_operations` | `if op not in content` (substring) — `GetUser` ⊂ `GetUserList` false-passed | `xmllint` well-formedness + XPath **exact** portType operation-set match |
| `validate_graphql_structure` | checks for the literals `type Query {` + `}` | `graphql-js` `buildSchema` + `validateSchema` |

Parsers live in a new hermetic `test/spec-validators/` npm package (pinned lockfile; offline after `npm ci`). Node-based because CI already provisions Node and ships `graphql-js`, and the anticipated egress allowlist includes `registry.npmjs.org` (not PyPI). A new `test/validate_test.sh` regression proves the fix: **10 assertions pass; the 6 prior false-passes are eliminated.** CI installs the package and runs the regression before the offline suite.

## T2 — `scan` E2E + write-method fixtures (gaps A1, A2)

- **A1:** new `scan-rest` live target drives the full `scan` command pipeline (crawl → `analyze.ExtractForms` → classify → generate) in one invocation. The rest-api test server now serves a static HTML `<form action="/api/subscribe">` that is never linked or fetched — so `/api/subscribe` appearing in the generated spec proves `ExtractForms` ran.
- **A2:** added PUT/PATCH/DELETE requests on `/api/users/{id}` to `reference-capture.json` and regenerated `expected-spec.yaml` (purely additive). `generate-rest` now covers all five verbs.

## T3 — Hard assertions (gaps B2, B3)

- **B2 (`crawl-depth`):** depth and max-pages limits now **hard-fail** instead of `log_warn`. Also corrected an off-by-one in the depth boundary — since depth is hop-count (seed `/deep/1` = depth 0), `--depth 2` legitimately reaches `/deep/3`, so the check now flags `/deep/4+`.
- **B3 (`import-malformed`):** a Go `panic`/goroutine stack trace is no longer accepted as a "graceful" non-zero exit — each import's output is scanned for panics.

## T4 — SSRF + auth coverage (gaps A4, A5)

- **A4:** new `ssrf-rejection` target asserts the SSRF gate **rejects** a private/loopback target (`http://127.0.0.1:9`) when `--dangerous-allow-private` is absent (every other target only ever *bypasses* the gate).
- **A5:** new `auth-capture` target asserts an `Authorization` header is captured and preserved through `import`; fixed the false "auth header tests" claim in `test/README.md`.
  - **Known gap (follow-up):** the generator does not yet emit OpenAPI `securitySchemes` from captured auth. Asserting *captured-auth → securitySchemes* needs that generator feature — out of scope for this test-hardening PR; worth a separate capability ticket.

## Verification

- Offline group: **21/21 PASS** (incl. new `ssrf-rejection`, `auth-capture`, panic-aware `import-malformed`).
- Live: `scan-rest`, `rest-api`, `crawl-depth` all **PASS**.
- `validate_test.sh`: **10/10** (RED against old validators was 4/6; the 6 false-passes now rejected).
- Drift guard (`test-runner-args.sh`): **32 passed, 0 failed** (anchors 21 offline + 8 live = 29).
- Negative tests confirm the new hard assertions fire on violations (a `/deep/5` capture fails; a simulated panic fails) and don't false-positive on clean graceful errors.

## New test targets

`scan-rest` (live), `ssrf-rejection` (offline), `auth-capture` (offline).

## Review follow-up

Blocking findings from the `/capability-pr-review` are tracked in **[LAB-5611](https://linear.app/praetorianlabs/issue/LAB-5611/vespasian-pr-187-follow-up-close-live-test-assertion-gaps-in)** — a sub-issue of [LAB-41](https://linear.app/praetorianlabs/issue/LAB-41/api-enumeration-tool-vespasian), same de-block route as [LAB-5332](https://linear.app/praetorianlabs/issue/LAB-5332). It covers all 7 blocking findings: **TEST-004, TEST-013, TEST-020, TEST-005, TEST-006, TEST-017, QUAL-002**.


santiago-praetorian:

**PR:** https://github.com/praetorian-inc/vespasian/pull/187
**Linear:** https://linear.app/praetorianlabs/issue/LAB-3890/vespasian-harden-live-test-suite-coverage-and-assertion-gaps
**Head SHA:** f4cd842693a85c512fe9a2d61e500b4c86571553
**Reviewed at:** 2026-07-22T00:54:07Z

|---|---|---|---|---|---|
| Count | 0 | 0 | 1 | 5 | 1 |

---

<details>

<details>
<summary><code>TEST-002</code> · medium · TEST · <code>test/run-live-tests.sh:1909</code></summary>

**Issue:** T1's validate.sh parsers got a dedicated regression harness (validate_test.sh) that proves the hardened validators reject the prior false-passes. The T3 hard-assertions added to run-live-tests.sh got no equivalent. check_panic() is pure string logic (a regex over captured output) yet is only exercised implicitly by the live import-malformed target, which passes as long as no panic happens to occur. Nothing proves the regex actually fires on a real Go panic/goroutine trace, so a broken pattern (e.g. a typo in 'goroutine [0-9]+ \[running\]') would silently downgrade the check to a no-op and re-introduce gap B3. The PR body claims 'Negative tests confirm the new hard assertions fire ... a simulated panic fails' but those negative tests are not in the diff. Same exposure for the crawl-depth /deep/4+ and max-pages>15 branches. check_panic in particular is trivially unit-testable in the validate_test.sh style (feed it a canned panic string, assert failures increments).

**Recommendation:** Add a regression case in the validate_test.sh style that sources run-live-tests.sh (or extracts check_panic into validate.sh) and asserts: (a) a canned string containing 'panic:' / 'goroutine 1 [running]:' is rejected, and (b) clean graceful-error output is accepted. This locks the panic-detection regex the same way validate_test.sh locks the parser validators.

```
    check_panic() {
        if printf '%s' "$output" | grep -qiE 'panic:|goroutine [0-9]+ \[running\]'; then
```

</details>

</details>

<details>

<details>
<summary>Ticket alignment (1)</summary>

<details>
<summary><code>REQ-001</code> · low · <code>test/run-live-tests.sh:2048</code></summary>

**Issue:** T4 gap A5 as written requires a fixture that asserts captured auth is surfaced as OpenAPI securitySchemes in the GENERATED spec. The PR's auth-capture target (test/run-live-tests.sh:2021-2062) only asserts the Authorization header is preserved through `import` (l.2041); it explicitly defers the captured-auth -> securitySchemes assertion because the generator does not yet emit securitySchemes (l.2048-2052, and commit message NOTE). SSRF-rejection (A4) and the README false-claim fix are fully done, so only this sub-part of T4 is unmet — leaving the criterion PARTIAL. The deferral is documented and blocked on an unimplemented generator feature, but the exact assertion A5 identified is still absent.

**Recommendation:** Either (a) implement securitySchemes emission in generate/rest and add a fixture asserting captured Authorization -> OpenAPI components.securitySchemes / security, fully closing A5; or (b) if the generator feature is genuinely out of scope, get the ticket's T4/A5 acceptance criterion amended to scope it to import-side auth preservation and formally split the securitySchemes generation into the referenced follow-up ticket, so the AC and the delivered coverage match.

</details>

</details>

<details>

<details>
<summary><code>SEC-BE-001</code> · low · <code>test/spec-validators/validate-openapi.mjs:32</code></summary>

**Issue:** SwaggerParser.validate(specFile) dereferences all $refs; swagger-parser resolves external http/https/file refs by default. A spec containing an external $ref would trigger a fetch/file-read in CI. Not reachable with current inputs (internal refs only; egress-audited runner).

**Recommendation:** Pass { resolve: { external: false } } to SwaggerParser.validate so external refs are not fetched.

</details>

</details>

<details>
<summary>Tests (3)</summary>

<details>
<summary><code>TEST-001</code> · low · <code>test/run-live-tests.sh:183</code></summary>

**Issue:** scan-rest is a live target that drives 'vespasian scan' against the rest-api server, but the REST arm of preflight_test_host matches only rest-api, edge-cases, and crawl-depth. Running '--targets scan-rest' alone therefore skips the /api/health preflight probe. This is fail-safe rather than a false-pass (a down server makes the scan fail, failing the test), but it is inconsistent with the sibling rest-api/crawl-depth targets and turns a clear 'server unreachable' preflight message into an opaque scan failure. In the full '--group live' run the rest-api arm happens to fire, masking the gap.

**Recommendation:** Add scan-rest to the REST case pattern: '*,rest-api,*|*,scan-rest,*|*,edge-cases,*|*,crawl-depth,*)' so the health preflight runs whenever scan-rest is selected.

```
        *,rest-api,*|*,edge-cases,*|*,crawl-depth,*)
            _probe_target_host "${REST_API_PORT:-}" "/api/health" "rest-api" || failed=1
```

</details>

<details>
<summary><code>TEST-003</code> · low · <code>test/run-live-tests.sh:2627</code></summary>

**Issue:** The hardened depth check (T3/B2) counts URLs at /deep/4+ and hard-fails if any exist, correctly catching an over-crawl (--depth not enforced). It does not assert the positive side of the boundary: that the crawl actually reached /deep/2 or /deep/3. A regression where the crawler under-crawls and stops at the seed (/deep/1, depth 0) would capture zero /deep/4+ URLs and pass green. The boundary assertion is therefore one-sided; the ticket frames this as 'hard-assert crawl-depth', which implies pinning both edges.

**Recommendation:** Add a positive assertion alongside the ceiling check, e.g. confirm at least /deep/2 (or /deep/3) appears in shallow.json, so the test fails on both over-crawl and premature-stop regressions.

```
        if [ "$beyond_depth" = "0" ]; then
            log_ok "Depth limit: correctly stopped within depth 2 (no /deep/4+)"
```

</details>

<details>
<summary><code>TEST-004</code> · nit · <code>test/validate.sh:300</code></summary>

**Issue:** The rewritten validator does exact per-name matching (grep -qxF against xmllint-extracted portType operation @name values), which correctly kills the GetUser ⊂ GetUserList substring false-pass the ticket targets. However it only verifies every expected operation is present (expected ⊆ actual); it never flags unexpected/extra operations. The success log ('all N operations found (exact match)') and the PR/commit wording ('XPath exact portType operation-set match') overstate this as full set-equality. Functionally fine for a coverage validator, but the 'exact set' claim is misleading and could lull a future reader into assuming extra operations are caught.

**Recommendation:** Either tighten the phrasing to 'all expected operations present (exact name match)' or, if set-equality is intended, additionally diff actual_ops against the expected set and fail on unexpected operations.

```
        if ! grep -qxF -- "$op" <<< "$actual_ops"; then
    log_ok "SOAP operations: all ${total} operations found (exact match)"
```

</details>

</details>

<details>

<details>
<summary><code>DOC-001</code> · low · <code>test/README.md:126</code></summary>

**Issue:** The PR adds three new user-selectable --targets values (scan-rest, ssrf-rejection, auth-capture — dispatched in run-live-tests.sh:3246/3273/3274) and mirrors the target reference list into the --help output, but the authoritative 'Valid targets' list in test/README.md (lines 126-138) still omits all three. auth-capture and ssrf-rejection were added to the results-directory tree only; scan-rest appears nowhere in the README. A user reading the README target list cannot discover the new targets.

**Recommendation:** Add scan-rest (to the Service/Scan group), ssrf-rejection, and auth-capture to the 'Valid targets' list in test/README.md and to the matching --help block in run-live-tests.sh.

```
                          Crawl:      crawl-depth, crawl-unreachable
                          Edge cases: edge-cases, classifier-edge, spec-edge
```

</details>

</details>

</details>

---

santiago-praetorian (changes requested):
**Recommendation: REQUEST CHANGES** — 1 blocker, 6 non-blockers.

Blocker:
- **TEST-002 (medium)** — T3's new hard-assertions (`check_panic`, crawl-depth/max-pages) got no committed regression test, unlike the T1 validators (`validate_test.sh`). `check_panic` is pure regex logic exercised only implicitly; a broken pattern would silently no-op and re-open gap B3. Add a canned-panic-string regression case in the `validate_test.sh` style.

Ticket alignment: T1–T3 fully met; T4 partial (auth→securitySchemes assertion deferred — REQ-001). Full finding set and evidence in the review comment above.

Strong test-hardening PR overall — the parser swap (T1) and SSRF-rejection/auth-capture targets (T4) are well-built. Once TEST-002 has a regression test, this is good to go.

eli-wald:
Thanks for the review — pushed `46d0baa` addressing it, plus a merge of `main` that resolves the conflicts this branch had picked up.

## Review findings

| Finding | Sev | Resolution |
|---|---|---|
| **TEST-002** | medium (blocker) | Detection extracted from the nested `check_panic` into `validate.sh:assert_no_panic` as a pure predicate; the nested wrapper is now one line and keeps sharing `failures` via dynamic scoping. Four regression cases added to `validate_test.sh` — panic string and goroutine stack trace rejected, clean graceful-error output and empty output accepted. |
| **TEST-003** | low | Depth boundary is no longer one-sided. The same `python3` pass now emits both counts, and the test additionally asserts the crawl actually reached `/deep/2+`, so an under-crawl that stops at the seed fails instead of passing green. |
| **TEST-001** | low | `scan-rest` added to the REST arm of `preflight_test_host`, so `--targets scan-rest` gets the `/api/health` probe instead of an opaque scan failure. |
| **TEST-004** | nit | Took the "tighten the phrasing" option rather than adding set-equality — enforcing it would reject WSDLs with legitimately extra operations. Log now reads `all N expected operations present (exact name match)` and the comment states plainly that extra/unexpected operations are **not** flagged. No behaviour change. |
| **SEC-BE-001** | low | `{ resolve: { external: false } }` passed to `SwaggerParser.validate`. |
| **DOC-001** | low | `scan-rest`, `ssrf-rejection` and `auth-capture` added to the README list. |
| **REQ-001** | low | **Deferring** — see below. |

On **TEST-002** specifically: the point of that finding was that a typo in the regex would silently downgrade the check to a no-op, so I verified the new regression actually has teeth — deliberately broke the pattern (`\[running\]` → `\[runnning\]`), confirmed the goroutine-stack-trace assertion failed, then restored it. It fails on a broken pattern now rather than passing quietly.

## One extra gap found

Reconciling the DOC-001 target lists turned up that **`forms-target` was missing from `usage()` entirely** — it is a real live target (member of `LIVE_TARGETS`, with a dispatch arm, a `/healthz` preflight probe, and a managed service in `setup-live-targets.sh`), so `--help` was not advertising it. Pre-existing, not from this PR, but fixed here since it is the same discoverability bug. The README list, the `usage()` list, and the `OFFLINE_TARGETS`/`LIVE_TARGETS` arrays are now reconciled at 32 names, checked by set-diff rather than by eye.

## REQ-001 — deferring

Leaving the existing documented deferral in place. The criterion wants captured auth asserted as OpenAPI `securitySchemes` in the **generated** spec, but `securitySchemes` exists nowhere in `pkg/generate/`, `internal/`, or `cmd/` — the generator does not emit it yet. Closing REQ-001 properly means building that generator feature, which turns a test-hardening PR into a feature PR and deserves its own review and tests. The deferral is already recorded in three places (the `auth-capture` target itself, the commit `NOTE`, and the PR body), and the import-side half of A5 — `Authorization` header captured and preserved through `import` — is asserted and passing.

Happy to file the follow-up ticket for `securitySchemes` emission and amend the T4/A5 wording to match what actually ships, if you'd prefer the AC and the delivered coverage line up rather than carrying the gap as a note.

## Verification

- `validate_test.sh` — **14/14** (10 pre-existing + 4 new)
- `test-runner-args.sh` drift guard — **32/32**
- Offline group — **21/21 PASS**
- Affected targets end-to-end (`import-malformed`, `generate-wsdl`, `generate-rest`) — PASS
- `check_panic` wrapper proven to still increment `failures` (0 → 1 → 2) after the extraction
- `bash -n` / `node --check` clean; `go build ./cmd/vespasian` OK
- SEC-BE-001 checked A/B: before the change an external `$ref` triggered a real outbound HTTP request; after it, the ref is left unresolved and internal `$ref`s still resolve identically

Not run locally: the live group (needs `setup-live-targets.sh` up) and `go test ./...` — no Go product code changed here, though the branch now carries `main`'s merged Go changes, so CI is the check on that.

## Status

Conflicts are resolved and the branch is mergeable again. **Another review round by me (Eli) is still pending** — please treat this as work-in-progress rather than ready to merge until I have done that pass.

santiago-praetorian:
## Praetorian Capability PR Review — `/capability-pr-review`

**Recommendation: REQUEST CHANGES** · reviewed at head `a4769a2` against [LAB-3890](https://linear.app/praetorianlabs/issue/LAB-3890) · scope: whole PR, 18 files, +2102/-98

| | critical | high | medium | low | nit |
|---|---|---|---|---|---|
| Findings | 0 | 0 | 7 | 23 | 5 |
| Blocking | 0 | 0 | 7 | 0 | 0 |

Nice PR overall — the parser swap is the real thing (I confirmed `swagger-parser` runs with `resolve:{external:false}`, that the WSDL check is now exact set-membership rather than substring, and that `validate_test.sh` runs 55/55 in its own un-gated CI job). The `ssrf-rejection` predicate is genuinely falsifiable: it requires the literal `seed URL rejected by frontier` **plus** the seed URL and rejects a kong usage banner, so a flag or subcommand rename cannot false-pass the gate. That is the right way to build a security assertion.

### Ticket alignment: T1–T4 all met ✅

Verified each against code, using the **amended** wording — WSDL on python3 `xml.etree` instead of `xmllint`, and A5 scoped to the import-side half with the `securitySchemes` generator work tracked in LAB-5332. The documented residuals were treated as accepted, not re-litigated. **Zero `REQ` findings.**

---

### Blocking (7)

Split by what I think each actually warrants, rather than presenting them as an undifferentiated wall.

#### Worth fixing in this PR (3) — small, and each sits on the ticket's own thesis

<details>
<summary><strong>TEST-004</strong> · medium · `test/rest-api/expected-paths.json:16` — test_rest_api reports endpoint vs expected count in PASS row without comparing</summary>

**Issue:** This PR bumps total_paths from 10 to 11 in expected-paths.json, and the README row it also updates advertises `rest-api PASS 11 11`. But test_rest_api reads total_paths only to print it: run-live-tests.sh:420-423 assigns endpoint_count and expected_count, and lines 426-432 pass both straight into set_test_result with no comparison between them. So this changed value can never fail anything for rest-api, and rest-api can emit `PASS 12 11` in the summary table. That is precisely the defect the sibling test's own inline comment names as a review finding and fixes: 'Exact-count: validate_path_coverage only detects MISSING paths, so without this a scan regression that emitted the expected paths plus spurious ones would report PASS with a mismatched pair of numbers in the summary (PR #187 review finding TEST-002)' (run-live-tests.sh:507-514). The fix was applied to test_scan_rest and not to test_rest_api, even though this PR edits both and both consume twin fixtures that are supposed to stay identical. The asymmetry is self-contradictory: either an exact path count is sound over a live crawl of this target — in which case rest-api is missing it — or it is not, in which case scan-rest's version is a flake generator. One of the two is wrong.

**Suggested fix:** Make the two tests consistent. Either add the same `[ "$endpoint_count" != "$expected_count" ]` failure to test_rest_api (the target already crawls the same server with the same --depth 2 --max-pages 50, so if the count is stable enough for scan-rest it is stable enough here), or, if the live crawl makes an exact count unsound, replace it in BOTH tests with an explicit lower bound plus an 'unexpected extra path' check so neither test prints an unasserted number next to PASS.

```
test/rest-api/expected-paths.json:16:  "total_paths": 11,
test/run-live-tests.sh:423:    expected_count=$(json_field "$expected" total_paths)
test/run-live-tests.sh:427:        set_test_result "rest-api" "PASS" "$endpoint_count" "$expected_count" "$duration"
test/run-live-tests.sh:511:    if [ "$endpoint_count" != "$expected_count" ]; then   # scan-rest only
```

</details>

<details>
<summary><strong>TEST-013</strong> · medium · `test/spec-validators/validate-graphql.mjs:76` — validateSchema branch of validate-graphql.mjs has no committed regression case</summary>

**Issue:** The validator's own header states validateSchema() exists specifically to catch 'a missing/empty Query root type that buildSchema alone tolerates'. Both committed reject fixtures short-circuit before that call: I ran graphql-js directly against test/validate_test.sh's bad.graphql fixture and buildSchema THREW ('Unknown type "NoSuchType".'), and junk.graphql is a parse error. So the `const errors = validateSchema(schema)` line and its `errors.length > 0` arm are dead as far as the regression harness is concerned -- deleting both leaves validate_test.sh green. The branch is reachable and does real work: `type Foo { a: String }` yields 'INVALID: Query root type must be provided.' with exit 1. A truncated or root-typeless SDL from pkg/generate/graphql is precisely the generator regression this half of T1 is supposed to catch, and it is currently unlocked.

**Suggested fix:** Add an assert_reject case in validate_test.sh with a fixture that parses cleanly but has no Query root type (e.g. `type Foo { a: String }`) and expected substring 'Query root type must be provided', so the buildSchema path and the validateSchema path are locked independently.

```
  const schema = buildSchema(sdl); // throws on syntax / SDL-structure errors
  const errors = validateSchema(schema); // schema-level errors (e.g. no Query)
  if (errors.length > 0) {
    console.error(`INVALID: ${errors[0].message}`);
    process.exit(1);
  }
```

</details>

<details>
<summary><strong>TEST-020</strong> · medium · `test/validate.sh:871` — assert_max_pages passes on an under-crawl: 0 or 1 page satisfies limit=10</summary>

**Issue:** assert_max_pages only asserts the upper bound (page_count > limit -> fail). It has no lower bound, so a crawl that visits zero or one page passes green. The sole caller (test/run-live-tests.sh:3365) crawls /api/many-links, which serves 20 distinct links to /api/items/1..20, with --max-pages 10; the CI log quoted in the function's own comment shows the correct value is exactly 10 ('visited 10 page(s) (limit=10)'). Verified by sourcing validate.sh and calling the predicate directly: page_count=0 -> PASS, 1 -> PASS, 10 -> PASS, 11 -> FAIL. And count_capture_pages returns "0" (not the "?" sentinel) for a structurally valid empty capture `[]`, so an empty-but-successful crawl reaches the assertion and satisfies it. A regression that neuters link extraction, the DFS frontier, or the rod tab lifecycle therefore reports 'Max-pages limit: visited 1 page(s) (limit=10)' as an [OK]. This is exactly the defect class the ticket exists to remove -- an assertion that passes on both the correct and the incorrect output -- and it is asymmetric with the sibling predicate hardened in this same PR: assert_within_depth got an explicit under-crawl guard ('crawl never got past the seed'), assert_max_pages did not. validate_test.sh actively locks the permissiveness in ('comfortably under the limit (3) is accepted') while committing no case for 0 or 1.

**Suggested fix:** Add a lower-bound arm to assert_max_pages, e.g. accept a required minimum argument and fail when page_count is below it ('crawl visited only N page(s), expected at least M -- under-crawl / premature stop'), mirroring assert_within_depth's reached-past-the-seed guard. For the /api/many-links caller the correct value is deterministic and equal to the cap, so `[ "$page_count" -eq "$limit" ]` is defensible; a conservative `>= 2` still closes the empty-crawl hole. Add committed cases in validate_test.sh for page_count 0 and 1 asserting rejection.

```
    if [ "$page_count" -gt "$limit" ]; then
        log_fail "${label}: visited ${page_count} page(s) (limit=${limit}) — --max-pages not enforced"
        return 1
    fi
    return 0
        if assert_max_pages "Max-pages limit" "$page_count" 10; then
assert_ok "comfortably under the limit (3) is accepted" \
    assert_max_pages "under-limit" 3 10
```

</details>

My reasoning for putting these three in this bucket: `TEST-020` is an assertion that cannot fail in one direction, which is the exact defect class LAB-3890 exists to remove (gap B2) — and its sibling `assert_within_depth` *did* get an under-crawl guard in this same PR, so the asymmetry is visible in adjacent functions. `TEST-004` is already fixed in the twin: `test_scan_rest` compares the counts at `test/run-live-tests.sh:511` under a comment citing this same defect, so one arm of a two-arm pattern was fixed and the other left. `TEST-013` closes with a one-line fixture — I verified that `type Foo { a: String }` reaches the branch and yields `Query root type must be provided`, whereas both committed reject fixtures throw inside `buildSchema` first and never get there.

#### Reasonable to defer to a follow-up (3)

<details>
<summary><strong>TEST-005</strong> · medium · `test/rest-api/scan-expected-paths.json:2` — Twin path fixtures kept in lockstep by prose; no check compares them</summary>

**Issue:** Both fixtures carry a ~1400-character description that shouts 'INTENTIONAL DUPLICATE — EDIT BOTH FILES ... the paths and total_paths are identical on purpose' and then names its own failure mode verbatim: 'updating only one silently diverges the rest-api and scan-rest expectations instead of failing loudly.' Nothing mechanical enforces it — I grepped the whole test tree and the only consumer of either file is validate_path_coverage (subset check) and json_field <file> total_paths; no code ever diffs the two. Two concrete silent divergences follow. (1) total_paths is a hand-maintained duplicate of len(paths) within each file and nothing asserts they agree. total_paths is the sole anchor of scan-rest's exact-count assertion (run-live-tests.sh:511), so bumping total_paths to 12 without adding the 12th path leaves that assertion green (12 == 12) while the new endpoint is never asserted by anyone — the exact-count teeth silently become vacuous. (2) Adding an endpoint to scan-expected-paths.json only (or to expected-paths.json only, see the sibling finding) passes both tests, because validate_path_coverage is a one-way subset check by construction (validate.sh:131-139) and rest-api has no count assertion. This is the same coverage-drift class that test-runner-args.sh guards with literal anchors, applied to fixtures instead of arrays, and the fix is the same size.

**Suggested fix:** Two cheap mechanical guards. (a) In validate_path_coverage's Python body, fail when expected.get("total_paths") != len(expected["paths"]) — this hardens every fixture in the suite (concat-spa, js-static, soap, forms-target) in three lines. (b) Add one assertion to test/test-runner-args.sh (already the drift-guard home, runs in 0.2s with no binary) that loads both rest-api fixtures and fails unless their paths dicts and total_paths are equal, replacing the prose contract with a failing test. Then shorten the two descriptions to point at the guard.

</details>

<details>
<summary><strong>TEST-006</strong> · medium · `test/rest-api/scan-expected-paths.json:4` — ExtractForms proof asserts path only; POST method and form fields unasserted</summary>

**Issue:** The new fixture declares "/api/subscribe": ["POST"], but validate_path_coverage never reads the method array — its Python body iterates keys only (`for path in expected["paths"]`, test/validate.sh:132) and compares path shapes. So the declared POST is inert. The only other proof in either test is `grep -q "/api/subscribe" "$spec_file"` (run-live-tests.sh:495 for scan-rest, :409 for rest-api). Nothing asserts (a) that the operation under /api/subscribe is a POST, or (b) that the form's two inputs (email, name — main.go:703-704) surface as its request-body properties. This matters because the load-bearing invariant behind the whole T2/A1 claim — that /api/subscribe is 'never linked or fetched' — is enforced only by a code comment (main.go:697-700), and main.go registers `mux.HandleFunc("/", handleIndex)` as a catch-all so /api/subscribe returns HTML 200 to any probe. If a future edit adds an href to the form action, or ExtractForms regresses to emitting a GET with no inferred body, the path string still appears in the spec and both tests still report PASS while proving nothing about ExtractForms. The suite already owns the exact scoped helpers needed: assert_post_get_operations (run-live-tests.sh:1084, driven by post_form_paths) asserts a POST exists and no GET was invented on a POST-only form action, and assert_form_body_fields (:867, driven by post_form_body_fields_by_path) resolves each path's own POST requestBody $ref and asserts the expected fields are properties of THAT schema. Both are applied to forms-target and to neither of the fixtures this PR touches.

**Suggested fix:** Add post_form_paths: ["/api/subscribe"] and post_form_body_fields_by_path: {"/api/subscribe": ["email", "name"]} to scan-expected-paths.json and expected-paths.json, then call assert_post_get_operations and assert_form_body_fields in test_scan_rest and test_rest_api alongside the existing grep. A form-derived body schema is something a crawl-discovered link cannot produce, so this also converts the 'never linked or fetched' code comment into a mechanical assertion.

</details>

<details>
<summary><strong>TEST-017</strong> · medium · `test/validate.sh:25` — No committed case for the validator timeout branch that mitigates the YAML alias bomb</summary>

**Issue:** _run_spec_validator's rc==124 branch is the ONLY real mitigation for the hang documented in validate-openapi.mjs (SEC-FE-003: js-yaml has no maxAliasCount, so anchor/alias expansion is unbounded). The 5 MiB SPEC_VALIDATOR_MAX_BYTES cap does not bound that input: I built a 450-byte nested-9-way-alias OpenAPI document and `node validate-openapi.mjs` did not return within 20s (exit 124 from an external timeout), i.e. the size cap is passed straight through and only the shell-side wall clock stops it. That branch has zero committed cases -- grepping test/validate_test.sh for SPEC_VALIDATOR_TIMEOUT, '_run_spec_validator' and 'timed out' returns nothing. Deleting the `if [ $rc -eq 124 ]` arm, or dropping the timeout wrapper entirely, leaves validate_test.sh at 55/55 green while restoring an unbounded hang that surfaces in CI only as an opaque job kill. SPEC_VALIDATOR_TIMEOUT is already an overridable env var, exactly as SPEC_VALIDATOR_MAX_BYTES is, so this is cheap to lock: I confirmed `SPEC_VALIDATOR_TIMEOUT=3 validate_openapi_structure <bomb>` prints '[FAIL] OpenAPI structure: validator timed out after 3s: validate-openapi.mjs' and returns 1. Additionally the no-timeout/no-gtimeout fallback silently runs node unwrapped, removing the backstop with no warning and no test.

**Suggested fix:** Commit a small (sub-1 KiB) YAML alias-bomb fixture plus an assert_reject case using the same subshell-wrapper pattern already used for SPEC_VALIDATOR_MAX_BYTES: (export SPEC_VALIDATOR_TIMEOUT=3; validate_openapi_structure "$bomb") with expected substring 'validator timed out after'. Pair it with an assert_ok under the same generous-timeout override so the case cannot pass by rejecting everything. Also log a warning in the unwrapped fallback branch so a host without timeout/gtimeout is visible rather than silently unprotected.

</details>

One caveat on `TEST-006`: it touches T2/A1's headline claim. The criterion says `scan-rest` proves `analyze.ExtractForms` ran; what it currently proves is that a path string appears in the spec. The `["POST"]` method array in the fixture is inert because `validate_path_coverage` only reads path keys (`test/validate.sh:132`). Deferring is defensible, but the follow-up should say plainly that the criterion shipped weaker than its wording.

#### I'd dismiss rather than track (1)

**QUAL-002** · medium · `.github/workflows/live-tests.yml:66` — Harden-runner/setup-node/npm-ci steps triplicated instead of extracted to composite action

`a4769a2` added *only comments* to this file, so this was a deliberate, documented decision to require lockstep bumps rather than extract — made in response to an earlier review round. Actions has no include mechanism for step blocks short of a composite action. My tool re-raised it because the blocker formula is mechanical about severity; I don't think it should hold the merge. Flagging it so the call is explicit rather than silent.

---

<details>
<summary><strong>Non-blocking (28)</strong> — low/nit, listed for completeness</summary>

| ID | Sev | Location | Summary |
|---|---|---|---|
| `QUAL-001` | nit | `.github/workflows/live-tests.yml:49` | New job comment cites ephemeral PR number instead of the Linear ticket used elsewhere |
| `QUAL-003` | low | `.github/workflows/live-tests.yml:164` | Workflow comment on 'Install spec validators' step wrongly attributes WSDL validation to swagger-parser/graphql-js |
| `QUAL-004` | low | `CLAUDE.md:125` | CLAUDE.md omits python3 as a hard dependency of test/validate_test.sh |
| `QUAL-005` | low | `CLAUDE.md:148` | CLAUDE.md claims generate-wsdl uses the real-parser validators; it doesn't |
| `SEC-BE-001` | nit | `test/run-live-tests.sh:2676` | Note that auth-capture ships an Authorization header into the CI artifact |
| `SEC-BE-002` | low | `test/spec-validators/package-lock.json:20` | Lockfile pins json-schema-ref-parser 14.0.1, affected by CVE-2026-15195 |
| `SEC-BE-003` | low | `test/spec-validators/validate-openapi.mjs:165` | Pass resolve.external:false to the second SwaggerParser call |
| `SEC-BE-004` | low | `test/validate.sh:17` | Validate SPEC_VALIDATOR_TIMEOUT; 0 silently disables the timeout |
| `SEC-BE-005` | low | `test/validate.sh:27` | Warn when _run_spec_validator silently runs without a timeout |
| `SEC-BE-006` | low | `test/validate.sh:59` | Add --ignore-scripts to the npm ci remediation message |
| `SEC-BE-007` | low | `test/validate.sh:344` | Bound the new WSDL XML parse with a size cap and timeout |
| `SEC-BE-008` | low | `test/validate_test.sh:42` | Unchecked mktemp -d in validate_test.sh lets fixtures write to / |
| `TEST-001` | low | `.github/workflows/live-tests.yml:55` | Drift guard stays inside the skip-live-tests gate the new job escapes |
| `TEST-002` | low | `test/README.md:18` | New npm ci prerequisite is undocumented in test/README.md, breaking its Quick Start |
| `TEST-003` | low | `test/README.md:290` | Documented target inventory (README, --help) not machine-checked vs arrays |
| `TEST-007` | nit | `test/rest-api/scan-expected-paths.json:17` | excluded_patterns in new fixture is never read by any validator |
| `TEST-008` | low | `test/run-live-tests.sh:409` | ExtractForms proof is a whole-file grep, not a path-key assertion |
| `TEST-009` | low | `test/run-live-tests.sh:482` | scan-rest path layers all pass on a spec with no path templating at all |
| `TEST-010` | low | `test/run-live-tests.sh:508` | scan-rest asserts an exact path count over a live crawl with probe and analyze-js on |
| `TEST-011` | nit | `test/run-live-tests.sh:2542` | check_panic is defined inside test_import_malformed and leaks into global scope |
| `TEST-012` | low | `test/run-live-tests.sh:3316` | Depth-boundary URL classification (the off-by-one T3 fixed) has no committed case |
| `TEST-014` | low | `test/spec-validators/validate-openapi.mjs:49` | SPEC_VALIDATOR_MAX_BYTES misconfiguration branch untested in both validators |
| `TEST-015` | low | `test/spec-validators/validate-openapi.mjs:176` | OpenAPI schema-violation failures collapse to 'Swagger schema validation failed.' |
| `TEST-018` | low | `test/validate.sh:330` | Three new guard branches in validate_soap_operations have no committed cases |
| `TEST-019` | low | `test/validate.sh:376` | portType-direct-children-only rule and _ensure_spec_validators wiring are unlocked |
| `TEST-021` | nit | `test/validate_test.sh:50` | assert_ok discards validator output and never asserts the success message |
| `TEST-022` | low | `test/validate_test.sh:111` | Four assert_reject cases use 'INVALID:' — the validators' universal failure prefix |
| `TEST-023` | low | `test/validate_test.sh:683` | validate_test.sh has no assertion-count anchor; silent shrink exits 0 |

</details>

<details>
<summary><strong>Coverage, and two caveats I owe you</strong></summary>

| Reviewer | Rubric items walked | Passes | Converged |
|---|---|---|---|
| capability-reviewer | 12/12 | 4 | **no** |
| backend-security | 18/18 | 4 | yes |
| frontend-security | 9/9 | 2 | yes |
| test-lead | 9/9 | 3 | **no** |

**1. This audit is incomplete — treat the finding set as a floor.** The review sweeps each lane until two consecutive passes surface nothing new. `backend-security` and `frontend-security` converged. `capability-reviewer` hit the 4-pass cap while pass 4 was *still* producing new findings, and `test-lead` lost its 4th pass to an infrastructure error. So a re-run may surface more, particularly in the test lane. That's a limitation of my run, not a defect in your code.

**2. One finding I'm withdrawing as out of scope.** `test/test-runner-args.sh:507` — the drift guard reports its pass count without anchoring its own assertion inventory. Real observation, but that summary block predates this PR; the only lines you changed in that file are the count literals at 166–188. Not yours to fix here.

</details>

---

### What unblocks approval

**Create a new sub-issue under [LAB-41](https://linear.app/praetorianlabs/issue/LAB-41) covering the deferred findings, link it in this PR, and I'll approve.** Under the review's blocker formula a finding with a tracked follow-up stops blocking, which is the same route LAB-5332 already took for the `securitySchemes` gap on this PR.

Two practical notes so the link actually registers:

- **Title the sub-issue so it names the files.** The follow-up is only credited to a finding when the referenced issue's title names the finding's file exactly, or shares a ≥10-character contiguous substring with the finding's title. A generic "address review feedback" will not de-block anything and the next run will re-report all of them. Something like *"Vespasian: close live-test assertion gaps in test/validate.sh and test/rest-api fixtures (PR #187 follow-up)"* works.
- **Reference it in the PR description**, not only in a thread, so it is picked up from the PR body.

If you'd rather ticket all seven instead of fixing the three above, that's your call — say so and I'll approve on the same condition. I'd just rather not ship the ticket that removes soft assertions while leaving `assert_max_pages` unable to fail on an empty crawl.

Also worth a one-line follow-up in the PR body itself: it still describes the WSDL validator as "`xmllint` well-formedness + XPath exact match" (shipped code is python3 `xml.etree`) and reports "10 assertions" for `validate_test.sh` (the harness now runs 55). Code is right; the description lags.

santiago-praetorian (approved):
Approved

santiago-praetorian:
Followup ticket: [LAB-5611](https://linear.app/praetorianlabs/issue/LAB-5611/vespasian-pr-187-follow-up-close-live-test-assertion-gaps-in)

PR #188, opened 2026-07-22
fix(crawl,classify,generate): deterministic mapping + stable classification (LAB-4678 Phase 0)

logan-bayless:
## Summary

Phase 0 of LAB-4678. Makes a Vespasian scan's emitted API surface a deterministic function of a
fixed input and decouples the REST-vs-not verdict from response timing, closing the root causes
LAB-4677 localized for Guard's run-to-run inconsistency. No behavior is added beyond determinism and
classification stability; coverage and fidelity gaps are later phases (see Scope).

## What changed

**Deterministic output (`pkg/classify/classifier.go`, `pkg/generate/rest/openapi.go`)**
- Dedup now selects the retained response order-independently (`preferredResponse`: populated beats
  half-captured, stable content-fingerprint tie-break) instead of keeping the first-seen observation,
  whose order was a nondeterministic map iteration.
- The OpenAPI generator sorts each endpoint group before the first-seen-wins response merge, and sorts
  the server list + derived title. A fixed capture now yields a byte-identical spec.

**Page budget (`pkg/crawl/engine.go`)**
- `--max-pages` counts pages visited (URLs), not captured requests — a single SPA page fires many
  XHR/fetch calls, so counting requests truncated the crawl far earlier than "max pages" implies.
- At the budget the crawl no longer hard-cancels the shared browser context. See the drain definition
  below.

**Timing-independent classification (`pkg/classify/rest.go`, `internal/pipeline/pipeline.go`)**
- New request-side signal (Rule 6): an API path plus a JSON/XML `Accept` or request content-type
  classifies the endpoint even when no response was captured, so a JSON GET whose response arrived too
  late still classifies. The confidence threshold is single-sourced as
  `classify.DefaultConfidenceThreshold`.
- `-v` logs the per-endpoint classification reason, listing every contributing signal.

**Review follow-ups (already applied)**
- Accept-header parsing splits into media ranges and treats `text/html`/`application/xhtml+xml` as a
  navigation, so a browser document-navigation Accept no longer misclassifies an HTML page under an
  api-like path (e.g. `/api/docs`, `/graphql`) as REST.
- A populated response schema is adopted when the deterministically-sorted base is empty, instead of
  being dropped when an empty observation sorts first.

## What "drain in-flight" means (precise)

At the page budget, workers stop taking **new** pages and the shared browser context is **not**
cancelled by the budget. Pages already being visited run their **normal, already-bounded** visit to
completion — page load, DOM-stability wait (`StableTimeout`), then a fixed settle — instead of being
killed mid-capture. This is **not** an unbounded wait for every outstanding HTTP response: a response
that has not arrived by the end of that page's own bounded visit is still captured half-finished
(which is exactly why the deterministic response selection and Rule 6 exist). The cap is exact
(a page slot is reserved before the visit), so the crawl never overshoots `--max-pages`. Completion-
driven capture that waits properly for late responses (network-idle bounded by floor/ceiling/per-
request timeout) is Phase 1, not in this PR.

## Scope

This is **Phase 0 of a five-phase ticket**. Deliberately out of this PR and tracked for later phases:
completion-driven capture and passive-request scope filtering (Phase 1); jsstatic Next.js recovery and
interaction-driven crawl (Phase 2); classify-beyond-allowlists / graded-confidence, size-independent
API verdict, response-schema union, and a separate `max_requests` rate control (Phase 3); cross-run
resume (Phase 4).

## Test plan

- `make check` (fmt + vet + lint + `go test -race ./...`) green on the rebased branch.
- New tests: shuffle-input generate determinism, dedup response selection, request-side classification
  + stability (incl. `Accept: */*` staying below threshold and browser-navigation not classifying),
  and the reason listing all signals.

Ticket: LAB-4678 · Branch rebased onto current `main`.


santiago-praetorian (changes requested):
## 🛡️ Praetorian Capability PR Review — `REQUEST CHANGES`

Automated multi-agent review (`/capability-pr-review`) of this PR against **LAB-4678**, plus static/dynamic/live test execution and acceptance-criteria validation against real binary behavior. Head SHA `37d4a75`.

|---|---|---|---|---|---|
| **Count** | 0 | 1 | 2 | 3 | 3 |
| **Blockers** | 0 | 1 | 2 | 0 | 0 |

**3 blockers · 9 findings total · 6 prior review-thread findings verified as already addressed and filtered.**

### Findings

| ID | Sev | Blocker | Location | Summary |
|---|---|---|---|---|
| QUAL-001 | medium | 🚫 YES | `pkg/classify/classifier.go`:282 | preferredResponse's SHA256 tie-break duplicates buildOperation's direct comparator |
| QUAL-002 | low | no | `pkg/classify/classifier.go`:405 | New classify.getHeader duplicates existing generate/rest.getHeader |
| QUAL-003 | low | no | `pkg/classify/rest.go`:178 | Rule 6 re-loops apiPathSegments instead of reusing Rule 3's result |
| QUAL-004 | low | no | `pkg/classify/rest.go`:191 | Rule 6's content-type match loop duplicates Rule 2's matching pattern |
| QUAL-005 | nit | no | `pkg/classify/rest.go`:202 | Rule 6 can duplicate Rule 2's content-type tag in the reason string |
| QUAL-006 | nit | no | `pkg/crawl/engine.go`:184 | crawlCtx/crawlCancel is now a vestigial WithCancel wrapper |
| TEST-001 | medium | 🚫 YES | `internal/pipeline/pipeline.go`:185 | No test for -v classification-reason listing (logClassificationReasons) |
| TEST-002 | high | 🚫 YES | `pkg/crawl/engine.go`:234 | AC12 unmet: no test asserts exact page cap or in-flight-drain emission |
| DOC-001 | nit | no | `CLAUDE.md` | CLAUDE.md crawl section omits new --max-pages page-budget semantics |

### 🧪 Test execution

| Suite | Command | Result |
|---|---|---|
| **Static** | `go build` · `go vet` · `gofmt -l` · `golangci-lint run` | ✅ **pass** — build/vet exit 0, gofmt clean, lint **0 issues** |
| **Dynamic (unit)** | `go test -race ./...` | ✅ **pass** — all **18 packages** `ok` |
| **Live / integration (browser)** | `test/setup-live-targets.sh` + `run-live-tests.sh --group all` | ⚠️ **could not run** in review env — only the Ubuntu `chromium-browser` snap **stub** is present (no functional Chrome); setup prerequisite check failed |

> ⚠️ **Coverage note:** the crawl integration tests that exercise this PR's page-budget change (`TestRodEngine_MaxPages`, `TestCrawlerContract_RespectsMaxPages`) are gated behind `//go:build integration` **and** require real Chrome — so they are **excluded from the default `go test -race` run and the CI coverage gate**, and did not execute here. They run only in the separate `live-tests.yml` job.

### 🎯 Acceptance-criteria validation (Phase 0 scope)

Validated the PR's own claims by driving the built binary against the fixed REST reference capture:

| AC | Status | Evidence |
|---|---|---|
| **AC1** — identical input → identical endpoint set + classification | ✅ **MET** | 3× `generate rest` on the fixed capture produced **byte-identical** output (`sha256=36f73a5d…` all three runs) |
| **AC2** — REST-vs-not stable & explainable via `-v` | 🟡 **PARTIAL** | `-v` emits per-endpoint reasons incl. the new Rule 6 tag, e.g. `reason=…+request-signal:accept:application/json`. Explains **present** endpoints (and missing-response ones via Rule 6); does **not** enumerate below-threshold/rejected endpoints ("explain a missing one"). Acceptable for Phase 0. |
| **AC12** — exact `--max-pages` cap + in-flight drain, asserted by a test | 🟡 **PARTIAL** | Behavior **shipped & correct** on inspection (reserve-before-visit under mutex, `engine.go:234–247`), but **no test asserts it** — `TestRodEngine_MaxPages` only checks `count != 0` and defers to a `contract_test.go` bound that is loose (`≤ maxPages+2`, Concurrency:1) and Chrome-gated. → **TEST-002 (blocker)**, matching the earlier CodeRabbit flag. |
| **AC13** — determinism/stability under `-race` | ✅ **MET** | Determinism/stability tests present and green under `go test -race ./...` |
| **AC3, AC4, AC5–AC11, AC14** | ⏸️ **DEFERRED** | Explicitly out of Phase 0 per the PR's Scope section; tracked in later phases / PR #189. Not findings. |

### 💬 Additional comments

- **Security (SEC-BE):** backend-security walked the full 12-item rubric and found **0 issues**. Rule 6's Accept-header parser is linear (no regex → no ReDoS) and draws tags from a fixed allowlist; `responseFingerprint`'s sha256 is a non-crypto content key; the worker mutex/budget accounting is sound and goroutines are `Concurrency`-bounded; the sole `#nosec G115` cast is the already-resolved gosec item. Frontend-security did not fire (no frontend files).
- **Docs:** README, `pkg/classify/doc.go`, and `pkg/crawl/doc.go` are updated well (Rule 6, determinism guarantee, `-v`, corrected `--max-pages` semantics). No other README needs changes. Only gap is a nit — **DOC-001**: `CLAUDE.md`'s detailed crawl section omits the new `--max-pages` page-budget semantics.
- **On QUAL-001 (blocker by formula, judgment call):** it flags two independent implementations of the same deterministic-ordering problem (`preferredResponse`'s sha256 tie-break vs `buildOperation`'s direct comparator). It's a maintainability/DRY concern, **not a correctness bug** — reasonable to consolidate now or downgrade if you consider the duplication acceptable. Either way the two **TEST** blockers stand on their own.
- **Determinism spot-check** beyond the suite: the byte-identical 3× run above independently confirms the core LAB-4677 root-cause fix.

<details>
<summary><strong>🚫 Blocker details (3)</strong></summary>

#### QUAL-001 · medium · QUAL · `pkg/classify/classifier.go`:282

**Issue:** This PR introduces two independent, differently-implemented solutions to the same problem — deterministically ordering/selecting among duplicate ObservedResponse values so a fixed capture yields a byte-identical result. `preferredResponse`/`responseFingerprint` (this file) hash StatusCode+ContentType+Body with SHA256 and pick the smaller digest. `buildOperation`'s new `sort.SliceStable` in pkg/generate/rest/openapi.go (lines 301-319) instead directly compares StatusCode, ContentType, and Body with plain field comparisons and bytes.Compare — no hashing. Both exist purely to make response ordering order-independent, added in the same PR, yet use unrelated techniques. Beyond being unnecessary complexity (a SHA256 dependency, a manual big-endian encode of the status code with a #nosec suppression) where a direct field comparison would do, having two divergent implementations of "deterministic response ordering" is a maintenance trap: a future change to the ordering rules (e.g. adding a new tie-break field) is likely to be applied to only one of the two, silently reintroducing order-dependence in the other.

**Recommendation:** Replace the SHA256 fingerprint tie-break with a direct field comparison (StatusCode, then ContentType, then bytes.Compare(Body)) matching the comparator already added in buildOperation, and share a single comparator function between the two call sites (or pkg/crawl) so both pieces of the determinism fix are driven by the same ordering rule.

```diff
func responseFingerprint(r crawl.ObservedResponse) [sha256.Size]byte {
	h := sha256.New()
	var status [8]byte
	binary.BigEndian.PutUint64(status[:], uint64(r.StatusCode)) // #nosec G115 -- status is a small non-negative HTTP code; wrap is irrelevant to a hash key
```


#### TEST-001 · medium · TEST · `internal/pipeline/pipeline.go`:185

**Issue:** logClassificationReasons is new user-facing behavior (the -v per-endpoint classification explanation, backing AC2 explainability and AC1 determinism) and has no test. Its testable branches are unexercised: the nil-writer / empty-slice no-op guard (line 186), the URL-parse path fallback (lines 192-194), the empty-reason -> "-" substitution (lines 196-198), the formatted line shape (lines 199-200), and — most importantly for AC1 — the deterministic sort.Strings ordering (line 202) that makes the -v output byte-stable for a given input. The Reason string CONTENT is covered at the classifier level (TestRESTClassifier_ReasonListsAllSignals), but the pipeline-level formatting and deterministic ordering are not, so a regression in the sort or nil-guard would ship undetected.

**Recommendation:** Add a pipeline-level test that calls logClassificationReasons with a bytes.Buffer and a fixed set of ClassifiedRequests in scrambled order; assert the emitted lines are sorted deterministically (identical output across input permutations), that an empty Reason renders as "-", that a URL with no parseable path falls back to the full URL, and that a nil writer / empty slice produce no output.

```diff
func logClassificationReasons(w io.Writer, classified []classify.ClassifiedRequest) {
if w == nil || len(classified) == 0 {
reason = "-"
sort.Strings(lines)
```


#### TEST-002 · high · TEST · `pkg/crawl/engine.go`:234

**Issue:** The PR ships new reserve-before-visit exact-cap logic (engine.go:234-247: "Reserving before the visit ... keeps the cap exact instead of overshooting by up to Concurrency pages") plus in-flight-drain emission (pages already in flight finalize and emit their captured requests instead of being canceled at the budget). AC12 explicitly requires an integration test asserting a saturated crawl visits EXACTLY --max-pages pages (no overshoot/undershoot) AND that in-flight pages still emit their captured requests; CodeRabbit already flagged the missing assertion, and AC12 is listed in-scope for PR188. The only budget integration test, TestRodEngine_MaxPages (engine_integration_test.go:161-207), was reworked to punt on this: it counts emitted ObservedRequests (pageVisits++ per request, not per page) and asserts only `count != 0`, with a comment conceding "Exact page-cap enforcement is covered separately." That separate test (contract_test.go:173 TestCrawlerContract_RespectsMaxPages, which is NOT part of this PR) asserts only `fetches <= maxPages+2` at Concurrency:1 — a loose upper bound that explicitly tolerates overshoot, performs no undershoot check, is not a saturated/concurrent scenario, and never asserts in-flight-drain emission. Net result: the exact-cap concurrency fix that is the entire point of this change is unverified, and the AC12 assertion requirement is not satisfied.

**Recommendation:** Add an integration test (build tag integration) that saturates the crawl with Concurrency > 1 against a fixture that fans out far more than --max-pages links, and count DISTINCT visited page URLs (dedupe document requests by URL, as TestRodEngine_ScopeFiltering already does) rather than raw ObservedRequests. Assert the distinct page count == MaxPages exactly (no overshoot, no undershoot). Separately assert that a page which is still in flight when the budget is reached emits its captured request(s) — e.g. a fixture page that issues an XHR after a short delay, asserting that XHR appears in the results even though the budget was hit. The rewritten test must also check the Crawl return value (currently `err` is assigned at engine_integration_test.go:186 but discarded).

```diff
*pageCount++
// keeps the cap exact instead of overshooting by up to
// count tallies emitted requests, not pages; a single page can emit several
// (document + subresources), so this only asserts the crawl produced output
// and terminated. Exact page-cap enforcement is covered separately.
if count == 0 {
```


</details>

<details>
<summary>Non-blocker details (6)</summary>

#### DOC-001 · nit · DOC · `CLAUDE.md`

**Issue:** The PR changes --max-pages to count pages visited (not captured requests) and adds an exact reserve-before-visit cap with in-flight drain. README.md and pkg/crawl/doc.go document this, but CLAUDE.md's otherwise-detailed pkg/crawl description does not mention the page-budget semantics at all. Low-value: CLAUDE.md never documented --max-pages previously and nothing in it became factually incorrect; flagged only for consistency with the detailed crawl section.

**Recommendation:** Optionally add one line to the CLAUDE.md pkg/crawl bullet noting --max-pages counts distinct pages visited with an exact reserve-before-visit cap and in-flight drain.


#### QUAL-002 · low · QUAL · `pkg/classify/classifier.go`:405

**Issue:** This PR adds `getHeader(headers map[string]string, name string) string` to pkg/classify, a case-insensitive header lookup. An almost identical helper (`getHeader`, case-insensitive lookup with an exact-match fast path) already exists in pkg/generate/rest/form.go and is used at pkg/generate/rest/openapi.go:442. The two packages now maintain two separate implementations of the same primitive. The codebase already has a precedent for sharing exactly this kind of helper across classify/generate-rest via pkg/mediatype (per package docs: "Shared media-type canonicalization... Used by classify and generate/rest where an import cycle prevents direct sharing"), so this is a discoverable, reusable pattern that wasn't applied here.

**Recommendation:** Move the case-insensitive header lookup into a shared leaf package (e.g. alongside or inside pkg/mediatype, which both classify and generate/rest already import) instead of maintaining two independent copies.

```diff
func getHeader(headers map[string]string, name string) string {
	for k, v := range headers {
		if strings.EqualFold(k, name) {
			return v
		}
	}
	return ""
}
```


#### QUAL-003 · low · QUAL · `pkg/classify/rest.go`:178

**Issue:** Rule 3 (lines 129-138) already loops over apiPathSegments to check whether lowerPath matches an API path segment, boosting confidence when it does. Rule 6 (lines 178-184) runs the identical loop again from scratch to compute a separate pathIsAPI boolean, instead of hoisting a single boolean computed once and reused by both rules. This is pure duplication of decision logic within the same function added by this PR.

**Recommendation:** Compute pathIsAPI once before Rule 3 (or have Rule 3 set it) and reuse the same boolean in Rule 6, removing the second loop over apiPathSegments.

```diff
pathIsAPI := false
for _, seg := range apiPathSegments {
	if strings.Contains(lowerPath, seg) {
		pathIsAPI = true
		break
	}
}
```


#### QUAL-004 · low · QUAL · `pkg/classify/rest.go`:191

**Issue:** Rule 6's request-content-type branch (lines 191-200) reimplements the exact charset-stripping + apiContentTypes exact-match loop that Rule 2 already performs against the response content-type (lines 107-126), just against a different header source. The matching loop body (`for _, apiCT := range apiContentTypes { if X == apiCT { ...; break } }`) is duplicated verbatim.

**Recommendation:** Extract a small helper, e.g. `matchAPIContentType(ct string) string`, that strips the charset parameter and returns the matching apiContentTypes entry (or ""), and call it from both Rule 2 and Rule 6.

```diff
reqCT := strings.ToLower(getHeader(req.Headers, "content-type"))
if idx := strings.Index(reqCT, ";"); idx != -1 {
	reqCT = strings.TrimSpace(reqCT[:idx])
}
for _, apiCT := range apiContentTypes {
	if reqCT == apiCT {
		signal = "content-type:" + apiCT
		break
	}
}
```


#### QUAL-005 · nit · QUAL · `pkg/classify/rest.go`:202

**Issue:** When a request has both a matching response Content-Type (Rule 2, e.g. "content-type:application/json") and a matching request Content-Type on an API path (Rule 6, e.g. "request-signal:content-type:application/json"), the reason string concatenates both tags even though they convey the same underlying fact. This was already flagged by the automated Gemini review on this PR ("Redundant signal logging in Rule 6") as functionally harmless but noise in the -v explanation output, and is not in addressed.json, so it appears still unresolved.

**Recommendation:** Skip appending the request-signal:content-type tag when Rule 2 already recorded a content-type signal for the same media type, or otherwise dedupe reason tags before joining.

```diff
if signal != "" {
	if confidence < RequestSignalConfidence {
		confidence = RequestSignalConfidence
	}
	reason = appendReason(reason, "request-signal:"+signal)
}
```


#### QUAL-006 · nit · QUAL · `pkg/crawl/engine.go`:184

**Issue:** Before this PR, crawlCtx was derived via context.WithCancel(ctx) so that hitting the page budget (in the now-removed wrappedOnResult) could call crawlCancel() to stop all workers early. That early-cancel call site was removed as part of this PR's fix (workers now just stop taking new pages under the mutex/pageCount). crawlCancel is now invoked only via the trailing `defer crawlCancel()`, at which point Crawl is already returning — so crawlCtx behaves identically to ctx for the entire body of Crawl. The derived context no longer serves a functional purpose beyond the standard "always call the returned cancel func" hygiene.

**Recommendation:** Either drop the WithCancel wrapping and pass ctx directly to workers (simpler, one less indirection to reason about), or add a short comment at the declaration noting it is now cleanup-only so a future reader doesn't assume it's still used to short-circuit the crawl.

```diff
// crawlCtx still derives from ctx so the parent's timeout/cancellation
// propagates to the workers; the budget no longer cancels it.
crawlCtx, crawlCancel := context.WithCancel(ctx)
defer crawlCancel()
```


</details>

---
<sub>Generated by Praetorian `capability-pr-review` (deterministic multi-agent) + manual static/dynamic/live test execution and binary-behavior validation. Reproducible against the same head SHA + ticket state.</sub>

logan-bayless:
All 9 findings addressed in `d14fa0c`. `make check` green.

- **QUAL-001** — Added `classify.CompareResponses` (StatusCode → ContentType → `bytes.Compare(Body)`) as the single ordering rule; `preferredResponse` and `buildOperation`'s sort both use it. Dropped the SHA256 fingerprint, big-endian encode, and `#nosec`.
- **QUAL-002** — Consolidated the header lookup into `mediatype.Header`; removed both `getHeader` copies.
- **QUAL-003** — `pathIsAPI` computed once, reused by Rule 3 and Rule 6.
- **QUAL-004** — Extracted `matchAPIContentType`, used by Rule 2 and Rule 6.
- **QUAL-005** — Rule 6 skips the request content-type tag when it duplicates Rule 2's matched media type; a differing request CT still fires.
- **QUAL-006** — Dropped the vestigial `context.WithCancel`; workers run under `ctx`.
- **TEST-001** — Added an internal `package pipeline` test for `logClassificationReasons` covering the deterministic sort, empty-reason `-`, unparseable-path fallback, and nil-writer/empty-slice guards.
- **TEST-002** — Rewrote `TestRodEngine_MaxPages`: `Concurrency:3`, fixture fans out past the budget, counts distinct `PageURL` and asserts `== maxPages` exactly (no over/undershoot), asserts every visited page emits its delayed XHR (in-flight drain), and checks the `Crawl` error.
- **DOC-001** — Documented the `--max-pages` page-budget semantics in CLAUDE.md.

santiago-praetorian:

**PR:** https://github.com/praetorian-inc/vespasian/pull/188  
**Linear:** https://linear.app/praetorianlabs/issue/LAB-4678  
**Head SHA:** a753c02d633195497019b1976c2b8ffc58a11019  
**Reviewed at:** 2026-07-23

|---|---|---|---|---|---|
| Count | 0 | 0 | 2 | 2 | 2 |
| Blockers | 0 | 0 | 2 | 0 | 0 |

---

<details>

<details>
<summary><code>QUAL-002</code> · medium · QUAL · <code>pkg/classify/classifier.go:287</code></summary>

**Issue:** CompareResponses is a new exported (public) function documented as "the single deterministic ordering rule for observed responses, shared by preferredResponse ... and buildOperation's group sort" — a comparator contract other packages (pkg/generate/rest) now depend on directly. It is only exercised indirectly: via Deduplicate/preferredResponse tests (which check the *outcome* of preferredResponse, not CompareResponses' own -1/0/+1 contract) and via generate/rest's group-sort tests. There is no test that calls CompareResponses directly to pin its ordering rules (StatusCode, then ContentType, then Body via bytes.Compare) or its edge cases (equal StatusCode+ContentType+Body → 0, ContentType tie-break direction, empty vs non-empty Body ordering).

**Recommendation:** Add a table-driven unit test in pkg/classify/classifier_test.go that calls CompareResponses directly with cases covering: differing StatusCode, equal StatusCode differing ContentType, equal StatusCode+ContentType differing Body, and fully-equal responses (expect 0) — this is now a public API contract that pkg/generate/rest relies on for sort correctness, per rubric item 6 (contract changes require a matching test).


```diff
func CompareResponses(a, b crawl.ObservedResponse) int {
	if a.StatusCode != b.StatusCode {
		if a.StatusCode < b.StatusCode {
			return -1
		}
		return 1
	}
```

</details>

<details>
<summary><code>REQ-001</code> · medium · REQ · <code>internal/pipeline/reasons.go:38</code></summary>

**Issue:** AC2 (in-scope for Phase 0; not in any deferred category of the PR Scope) requires that -v explain both a present endpoint AND a missing one. The implementation only explains PRESENT endpoints. logClassificationReasons ranges over the `classified` slice (reasons.go:38), which is populated by RunClassifiers only with requests whose best confidence >= threshold (classifier.go:93) and passed to the logger at pipeline.go:103. Every request that was classified as non-REST or fell below the confidence threshold is dropped before -v runs, so there is no -v line explaining WHY a given request is absent from the API surface. Rule 6 does cover the narrower 'missing RESPONSE' case (an endpoint whose response arrived too late still classifies and gets a request-signal reason), but a request that is genuinely rejected/omitted from the spec produces no explanation at all. The 'explain a missing one' clause is therefore partially met.

**Recommendation:** Emit a -v explanation line for requests that were considered but NOT included in the API surface (e.g. surface the below-threshold/non-API decisions with their confidence and the reason they were rejected — static-asset exclusion, below-threshold path-only heuristic, navigation Accept, etc.). This requires carrying the rejected/considered requests (or their ClassifyDetail reason) alongside the accepted `classified` slice through RunClassifiers so logClassificationReasons can render both the present and the missing/rejected endpoints deterministically. Add a pipeline-level test asserting a rejected request appears in the -v output with its rejection reason.

</details>

</details>

<details>

<details>

<details>
<summary><code>SEC-BE-001</code> · low · <code>internal/pipeline/reasons.go:39</code></summary>

**Issue:** logClassificationReasons (new file, called from pipeline.go:103 under -v, where opts.Status is os.Stderr per cmd/vespasian/main.go:857-862) writes each classified endpoint's URL path directly to the operator's terminal without neutralizing control or ANSI escape sequences. `path` is set from `u.Path` of the observed request URL, which is derived from crawling an UNTRUSTED target: a target can serve links whose path segments carry percent-encoded control bytes (e.g. `%1B` ESC, `%0A` newline), and url.Parse decodes these into the raw bytes of u.Path. Those raw bytes are then emitted to stderr via the `%s` line, so a hostile target can inject terminal escape sequences into the analyst's session (CWE-150 / CWE-117: log/terminal escape injection) to spoof, hide, or overwrite displayed classification results — misleading the operator reviewing the discovered API surface. The reason/type/method fields are drawn from fixed vocabularies and are safe; the path is the untrusted field. Impact is limited: verbose mode is off by default and the effect is confined to terminal rendering (no code execution in a well-behaved emulator), hence low.

**Recommendation:** Before formatting the line, sanitize the untrusted path (and the fallback full-URL) by stripping or escaping non-printable/control bytes — e.g. replace bytes < 0x20 and 0x7f, or use strconv.Quote on the path, or filter with a printable-rune allowlist. Apply the same neutralization anywhere else crawled URLs are echoed to a terminal.


```diff
		path := c.URL
		if u, err := url.Parse(c.URL); err == nil && u.Path != "" {
			path = u.Path
		lines = append(lines, fmt.Sprintf("  %-6s %s [type=%s confidence=%.2f reason=%s]",
			strings.ToUpper(c.Method), path, c.APIType, c.Confidence, reason))
```

</details>

</details>

<details>

<details>
<summary><code>QUAL-001</code> · nit · <code>pkg/classify/classifier.go:29</code></summary>

**Issue:** The new exported constant's doc comment states it is "single-sourced here and referenced by the SDK and CLI defaults so the threshold is documented rather than a bare literal scattered across entry points". pkg/sdk/capability.go does reference classify.DefaultConfidenceThreshold (verified), but cmd/vespasian/main.go's Kong CLI struct tags for --confidence still hardcode the literal "0.5" in two places (ScanCmd and GenerateCmd Confidence fields) and were not touched by this PR. The comment is inaccurate for the CLI half of its own claim, which undermines the single-source-of-truth intent it documents.

**Recommendation:** Either wire the CLI struct tag defaults to reference the constant (Kong tags are string literals so this may require a small default-resolution shim), or narrow the doc comment to state it is referenced by the SDK only (and note the CLI default is a separately-maintained literal that must be kept in sync).


```diff
// DefaultConfidenceThreshold is the default minimum confidence for a request to
// be classified as an API. It is single-sourced here and referenced by the SDK
// and CLI defaults so the threshold is documented rather than a bare literal
// scattered across entry points (LAB-4678).
const DefaultConfidenceThreshold = 0.5
	Confidence             float64 `default:"0.5" help:"Minimum confidence threshold"`
```

</details>

<details>
<summary><code>QUAL-003</code> · low · <code>pkg/classify/rest.go:252</code></summary>

**Issue:** matchAPIContentType (lines 217-225) is documented as "Shared by Rule 2 (response content-type) and Rule 6 (request content-type) so both apply one matching rule (QUAL-004)", but acceptSignalsAPI (the Accept-header half of Rule 6) does not call it — it re-implements the identical apiContentTypes membership loop inline. This is a 3+ line duplicate of core matching semantics in the same file, immediately below the shared helper whose stated purpose is to prevent exactly this kind of drift. If apiContentTypes matching logic (e.g. wildcard subtype matching) changes in matchAPIContentType later, this inline copy will silently diverge.

**Recommendation:** Replace the inline for-loop in acceptSignalsAPI with a call to matchAPIContentType(mt) (mt is already lowercased and stripped of q-parameters, so mediatype.Base(mt) is a safe no-op on it).


```diff
func matchAPIContentType(ct string) string {
	base := mediatype.Base(ct)
	for _, apiCT := range apiContentTypes {
		if base == apiCT {
			return apiCT
		}
	}
		if match == "" {
			for _, apiCT := range apiContentTypes {
				if mt == apiCT {
					match = apiCT
					break
				}
			}
		}
```

</details>

<details>
<summary><code>QUAL-004</code> · nit · <code>pkg/crawl/engine.go:218</code></summary>

**Issue:** The pre-Pop "stop taking new pages" check (lines 218-223) and the post-Pop "reserve this page's budget slot" check (lines 239-246) both lock mu, test `maxPages > 0 && *pageCount >= maxPages`, and unlock — a 4-line pattern repeated verbatim with only the cleanup-on-hit branch differing (plain return vs. MarkIdle+return, plus the increment). Both instances are well-commented and individually correct (verified: race-free under `go test -race -tags=integration`), so this is a readability/DRY nit rather than a defect.

**Recommendation:** Optional: extract a small helper, e.g. `budgetReached(mu *sync.Mutex, pageCount *int, maxPages int) bool`, called from both sites, to remove the duplicated lock/compare/unlock block.


```diff
		mu.Lock()
		if maxPages > 0 && *pageCount >= maxPages {
			mu.Unlock()
			return
		}
		mu.Unlock()
		mu.Lock()
		if maxPages > 0 && *pageCount >= maxPages {
			mu.Unlock()
			e.frontier.MarkIdle()
			return
		}
		*pageCount++
		mu.Unlock()
```

</details>

</details>

</details>

_9 findings dismissed as already addressed — see `.review/addressed.json` (all PR review threads resolved)._


---

logan-bayless:
**Re: capability-pr-review (a753c02)** — addressed in `ea3765f`. `make check` green (fmt, vet, lint 0 issues, `go test -race ./...`).

- **QUAL-002** — Added `TestCompareResponses` (`pkg/classify/classifier_test.go`): direct table-driven coverage of the StatusCode→ContentType→Body ordering, the fully-equal→0 case, empty-vs-non-empty body, plus an antisymmetry check.
- **SEC-BE-001** — Added `sanitizeForTerminal` in `internal/pipeline/reasons.go`, applied to both the path and the method (the method is equally attacker-influenced from imported captures). Escapes non-printable runes to `\xNN`; regression test added.
- **QUAL-003** — `acceptSignalsAPI` now calls `matchAPIContentType(mt)` instead of the inline `apiContentTypes` loop.
- **QUAL-004** — Extracted a single `pageBudgetReached(mu, pageCount, maxPages, reserve)` helper. Note: a bool-returning `budgetReached` with a separate `*pageCount++` (as suggested) would split the compare and increment across two lock acquisitions and reintroduce a cap-overshoot race at the reserve site; this helper keeps both in one critical section.
- **QUAL-001** — The CLI can't reference the constant in a Kong `default:` tag (tags must be literals; wiring it via `kong.Vars` panics every parser that doesn't supply the var). Narrowed the doc comment to state the SDK references the constant and the CLI keeps a synced literal, and added `TestCLIConfidenceDefaultMatchesConstant` so CI fails if they drift.
- **REQ-001** — Respectfully pushing back: this is Phase 1, not Phase 0. The implementation plan assigns "sub-threshold near-miss logging under `-v`" to Phase 1, and the first-review pass rated AC2 PARTIAL as "Acceptable for Phase 0." Phase 0's `-v` scope is per-endpoint reasons for classified endpoints; enumerating rejected/below-threshold endpoints lands with the Phase 1 work in PR #189. Happy to track it explicitly there.

santiago-praetorian:
**QUAL-001** · `nit` · **not a blocker** · `internal/pipeline/reasons.go:39-43`

`sanitizeForTerminal` shadows its loop variable `r` in the nested loop. The outer scan at :39 binds `r`, and the inner rebuild loop at :43 re-binds the same name over the same string:

```go
for _, r := range s {            // :39 — outer: detect whether escaping is needed
    if !unicode.IsPrint(r) {
        var b strings.Builder
        b.Grow(len(s))
        for _, r := range s {    // :43 — inner: shadows the outer r
            if unicode.IsPrint(r) {
                b.WriteRune(r)
            } else {
                fmt.Fprintf(&b, "\\x%02x", r)
            }
        }
        return b.String()
    }
}
```

**This is correct code, not a defect.** The inner loop deliberately rescans `s` from the start, and the outer `r` is genuinely unused inside the branch, so the shadowing has no behavioral consequence. The fast-path-then-rebuild structure is a reasonable way to avoid allocating a `strings.Builder` for the common all-printable case.

Flagged only as a readability nit: a reader has to confirm the inner `r` isn't meant to be the outer one. Optional fix is renaming the inner binding (e.g. `for _, rr := range s`) or hoisting the rebuild into a small helper.

For the record on why this is non-blocking:

- It is `nit` severity, so it fails the material clause of the blocker formula (`critical`/`high`/`medium`, or a missing-tests `TEST` finding).
- It violates no convention in this repo. The `.golangci.yml` linter set (`gocritic`, `gocyclo`, `gosec`, `misspell`, `revive` — the latter configured with only the `exported` rule) enables no shadow check, which is consistent with `make check` being green.
- No test or behavior change is required.

Take it or leave it — it does not gate merge.

santiago-praetorian (approved):
Approving. Re-reviewed the delta since my last pass (`a753c02` → `4dbefab`): all six findings from that round are genuinely fixed — I verified each landed rather than taking the summary at face value — and the follow-up commit introduces no substantive defect. The only new finding is QUAL-001, a non-blocking readability nit noted separately above.

Also confirmed the four Phase 0 acceptance criteria in LAB-4678 (AC1, AC2, AC12, AC13) are met, and that the determinism/classification-stability tests run untagged under `go test -race ./...`.

PR #197, opened 2026-07-30
test: make live & dynamic (integration) tests runnable in devcontainers (LAB-5064)

santiago-praetorian:
## Description

Makes the **live** and **dynamic (integration)** test layers runnable inside a devcontainer. Closes most of [LAB-5064](https://linear.app/praetorianlabs/issue/LAB-5064/vespasian-make-live-and-dynamic-integration-tests-runnable-in).

A stock Ubuntu devcontainer ships `/usr/bin/chromium-browser` as a snap launcher stub with snapd unavailable in-container, so there is no runnable browser. LAB-3893 made `setup-live-targets.sh` detect that stub and fail loudly but explicitly deferred the fix. Two consequences:

- **Live** — `run-live-tests.sh` could not run at all, because setup hard-failed at the Chrome preflight before writing `.live-test-config`. That took the **offline** group down with it, even though offline targets need no browser whatsoever.
- **Dynamic** — the `//go:build integration` browser lifecycle tests in `pkg/crawl` could not start Chrome.

> [!IMPORTANT]
> **AC1 is only partially met — the remainder is tracked in [LAB-5766](https://linear.app/praetorianlabs/issue/LAB-5766).** The install recipe lives in `test/install-chrome.sh` and is verified working end-to-end, but **nothing invokes it at devcontainer build time** — the devcontainer definition lives outside this repo. A fresh container still comes up browserless until that image layer calls this script. Everything else (AC2–AC5) is complete and verified.

## Changes

- **`test/install-chrome.sh` (new)** — provisions a real, non-snap Chrome. Google's apt repo is added **temporarily** with its signing key pinned by **primary-key fingerprint**, so apt verifies Release → Packages → `.deb` before dpkg runs maintainer scripts as root; the repo and keyring are then removed via an `EXIT` trap.
  - Downloading the `.deb` directly would leave TLS as the only control: `apt-get install ./local.deb` does **not** authenticate a local file, and Google's `.deb` carries no embedded `debsigs` signature (verified — no `_gpgorigin` member). The fingerprint pin is what makes the check meaningful.
  - The package's own permanent apt source and `/etc/cron.daily/google-chrome` pinger are suppressed via `repo_add_once=false` and verified absent, so having Chrome present adds no background egress. In-test browser telemetry is already handled by LAB-4999.
  - amd64 **and arm64** (the ticket assumed amd64-only; Google now publishes both).
- **`run-live-tests.sh`** — loads `.live-test-config` only when a selected target actually talks to a live service. `--group offline` now runs on a fresh checkout with no setup run at all.
- **`setup-live-targets.sh`** — the browser gate is conditional: fatal only when the selection provisions a browser-backed target. `--skip-start` and grpc-server-only setups warn and continue. Selections that *do* need a browser still fail loudly, preserving LAB-3893.
- **`common.sh`** — the browser probe moves here so the installer's "already present" check and the preflight gate cannot drift apart.
- **CI** — `install-chrome-selftest.sh` and `setup-live-targets_test.sh` added to the un-gated `preflight-selftest` job; `test-runner-args.sh` **moved** there from the label-gated job, where `skip-live-tests` was silently skipping the drift guard.

## Testing

- [x] Unit tests pass (`make test`) — 0 failures
- [x] Linter passes (`make lint`) — 0 issues
- [x] Manual verification (below)

Verified in an **arm64** devcontainer:

| Criterion | Result |
|---|---|
| AC2 `--group offline` | **19/19 pass**, no browser, no config, no setup run |
| AC3 `--group live` | **9/9 pass, 0 skipped**; `no-download` confirms system Chrome, no auto-download |
| AC1 integration tests | `go test -tags integration ./pkg/crawl` — launch/kill/close all **run**, not skip |
| AC4 phone-home | apt source, cron pinger, temp repo and keyring all absent post-install |
| AC5 production behaviour | **no Go files changed** on the branch |

Also: Chrome purged and reinstalled end-to-end through the signature-verified path, and the fingerprint pin verified to **fail closed** on a mismatch.

Shell suites went **84 → 317 assertions**, measured against `main`: `preflight-selftest` 12 → 65, `test-runner-args` 32 → 75, `install-chrome-selftest` new → 137, `setup-live-targets_test` 40 (unchanged by this PR).

**`shellcheck`:** this PR does add findings against `main`, all of them warning-level false positives from the derive-from-source idiom the drift guards require — `test-runner-args.sh` 7 → 21 (13 × SC1090 on `source <(sed …)`, 8 × SC2034 on variables `shellcheck` cannot track through that `source`) and `preflight-selftest.sh` 0 → 2 (SC2317 on stub functions invoked indirectly through `main`). The two new files sit at 0 (`install-chrome.sh`) and 40 (`install-chrome-selftest.sh`, dominated by SC2030/SC2031 on the deliberate subshell-scoped `PATH`/`SUDO` overrides that give each case its isolation). `shellcheck` is not a CI gate here, and `install-chrome.sh` — the only file that runs as root — is clean. An earlier revision of this description claimed zero new findings; that was true when written and is no longer.

## Checklist

- [x] Code follows project conventions
- [x] Tests added/updated for new functionality
- [x] Documentation updated — `test/README.md` (Chrome-in-containers, browserless runs, the integration layer) and `CLAUDE.md`

## Review notes

This branch went through seven rounds of automated review. Rounds 1–6 each surfaced at least one test that **passed for the wrong reason**; round 7 found nothing across all three reviewers. Every security-relevant assertion is mutation-verified — disabling the control fails the test. Worth a human eye regardless:

- The **fingerprint pin** (`GOOGLE_KEY_FPR`) is the entire trust anchor. If Google rotates its primary key the script fails loudly and the constant must be updated deliberately.
- `test/fixtures/not-google-signing-key.asc` is a PGP **public** key (Ubuntu's 2012 archive key, no secret material) used only as a "not Google's key" sample.
- The script tracks Chrome *stable* rather than pinning a version — right trade for a test-only layer, but the installed build is not reproducible from the script alone, so the version is logged on success.

Refs LAB-5064; the AC1 remainder (wiring the installer into the devcontainer image) is tracked in LAB-5766. Follow-up to LAB-3893; sibling to LAB-4994; in-test browser telemetry covered by LAB-4999.


test/install-chrome.sh
```diff
@@ -0,0 +1,343 @@
...
+    if ! gpg --dearmor < "$tmp_key" > "$tmp_gpg" 2>/dev/null; then
+        log_fail "Fetched signing key is not a valid PGP key."
+        return 1
+    fi
+
+    # Compare against the primary key's fingerprint only (the `pub` record);
+    # subkey fingerprints rotate and are not the trust anchor.
+    fprs=$(gpg --show-keys --with-colons --with-fingerprint "$tmp_key" 2>/dev/null \
+        | awk -F: '$1=="pub"{want=1} $1=="fpr" && want{print $10; want=0}')
+    if [ "$fprs" != "$GOOGLE_KEY_FPR" ]; then
```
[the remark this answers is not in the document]
santiago-praetorian: Already addressed in f7fa870: `install_pinned_key` now collects the fingerprint of **every** `pub` record in the fetched bundle (`awk` over `--with-colons`) and does a membership check (`grep -qxF`) against `GOOGLE_KEY_FPR`, then exports **only** the pinned key into the apt keyring — a multi-key bundle with the active key present passes, and obsolete keys in the bundle are never installed. Covered by install-chrome-selftest (multi-key-bundle and mismatch cases). Resolving.

.github/workflows/live-tests.yml
```diff
@@ -66,7 +88,16 @@ jobs:
...
+    # preflight-selftest as well as check-label: the drift guard and the shell
+    # self-tests moved into that job, and they used to be the FIRST step of this
+    # one, where a failure stopped the 30-minute suite immediately. Depending on
+    # the job restores that fail-fast: there is no value in building Go, starting
+    # six services and crawling with a browser when the guards have already shown
+    # the target groups have drifted out of sync with the dispatch block.
+    #
+    # preflight-selftest needs no Go, Node or Chrome and finishes in well under a
+    # minute, so serializing on it costs far less than a wasted full run.
+    needs: [check-label, preflight-selftest]
```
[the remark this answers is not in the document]
santiago-praetorian: Skipping this one deliberately. The premise ("if branch protection only requires `test`") does not hold for this repo: there are **no required status checks** on `main` at all (only pull-request review + thread-resolution rules, verified via `gh api repos/praetorian-inc/vespasian/rules/branches/main`) — CI here is advisory and merges are gated on human review. A failed `preflight-selftest` is itself a red check on the PR, and the `needs:` edge already gives the fail-fast behavior the comment asks about (a preflight failure skips the 30-minute suite). An `always()`-gated pass-through job would add workflow complexity to defend a protection rule that does not exist; if required checks are ever introduced, both jobs should simply be listed as required. Resolving.

test/preflight-selftest.sh
```diff
@@ -214,6 +215,288 @@ out_f=$(echo "${result}" | sed -n '2p')
...
+
+    assert_eq "case f2: slow browser passes under the default 2s budget" \
+        "0" "$(probe_slow_browser)"
+    rc_f2=$(CHROME_PROBE_TIMEOUT=0.2 probe_slow_browser)
+    if [ "${rc_f2}" != "0" ]; then
+        echo "PASS: case f2: CHROME_PROBE_TIMEOUT=0.2 kills the slow probe (rc ${rc_f2})"
+        pass_count=$((pass_count + 1))
+    else
+        echo "FAIL: case f2: CHROME_PROBE_TIMEOUT=0.2 was ignored — slow probe still passed"
+        fail_count=$((fail_count + 1))
```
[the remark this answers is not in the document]
santiago-praetorian: Fixed in 6123b04: the kill assertion now requires rc **124** exactly — the status GNU timeout / gtimeout reserve for "I killed it" — instead of any nonzero rc, so an invalid duration (125) or the probe's own failure code can no longer pass as proof the budget was enforced. Verified locally: suite 35/0, and the mutation copy (hardcoded `2` restored) still fails exactly case f2.

test/preflight-selftest.sh
```diff
@@ -214,6 +215,291 @@ out_f=$(echo "${result}" | sed -n '2p')
...
+            chrome_runnable "${SLOW_BROWSER}"
+            printf '%s\n' "$?"
+        )
+    }
+
+    # Pin the budget to its default for the first probe: the README tells
+    # users to export CHROME_PROBE_TIMEOUT, and an ambient value below ~1s
+    # would fail this assertion spuriously. (Empty means "unset" to :-.)
+    assert_eq "case f2: slow browser passes under the default 2s budget" \
+        "0" "$(CHROME_PROBE_TIMEOUT= probe_slow_browser)"
```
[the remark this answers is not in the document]
santiago-praetorian: Fixed in 6123b04: `CHROME_PROBE_TIMEOUT=''` (explicit quoted empty). SC1007 gone; `shellcheck` on the file now reports only the pre-existing SC1091 info notes.

test/install-chrome.sh
```diff
@@ -0,0 +1,594 @@
...
+                printf '[FAIL] VESPASIAN_TEST_ROOT contains characters outside [A-Za-z0-9._/-]: %s\n' \
+                    "$TEST_ROOT" >&2
+                exit 1
+                ;;
+        esac
+        ;;
+    *)  printf '[FAIL] VESPASIAN_TEST_ROOT must be an absolute path: %s\n' "$TEST_ROOT" >&2
+        exit 1
+        ;;
+esac
```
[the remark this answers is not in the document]
santiago-praetorian: Valid, and thanks — this was a real gap in validation added earlier in this same branch, not a pre-existing one. The charset check allows `.` and `/`, so `/tmp/x/../..` satisfied every test and then resolved to `/`, which prefixes nothing and would point `TMP_LIST`, the defaults-file rewrite and the phone-home removal at the real system paths under `$SUDO`. A test root symlinked to `/` reached the same place.

Fixed in `f6d0fc0` with two layers, because neither subsumes the other:
- a string-level rejection of any `..` component, which works on a path that does not exist yet (canonicalization cannot);
- `cd -P` / `pwd -P` resolution for a root that does exist, with a resolved `/` refused outright.

New selftest **case t** covers traversal (`root/../..`, `/tmp/../..`), a root symlinked to `/`, a bare `/`, a relative path and an embedded space, plus the positive case that a normal fixture root is still accepted. Mutation-verified: reverting the guard fails 4 of the 7 assertions. Resolving.

santiago-praetorian:
### Scope note: the devcontainer image wiring lands separately

Flagging this for reviewers so the gap is visible before merge rather than after.

This PR adds `test/install-chrome.sh` and makes the live and dynamic test layers runnable, but it **does not wire the installer into an image build**. Nothing in this repo invokes it automatically: there is no `.devcontainer` definition here, and no Dockerfile, workflow, Makefile, or `postCreateCommand` calls it. A fresh container is therefore still browserless until someone runs the script by hand.

The devcontainer definition lives outside this repository, so that wiring is not a change this PR can make. It is tracked in **LAB-5766**.

What *is* delivered here, verified on this branch:

| Check | Result |
|---|---|
| `./test/run-live-tests.sh --group offline` | 21/21 pass — browserless, no `.live-test-config` |
| `./test/run-live-tests.sh --group live` | 10/10 pass, 0 skipped — rod-backed targets including `no-download` execute rather than SKIP |
| Phone-home artifacts after install | absent; asserted by the new `install-chrome-e2e` job |
| Shell guard suites | 365 assertions, 0 failures (`install-chrome-selftest` 164, `preflight-selftest` 71, `test-runner-args` 90, `setup-live-targets_test` 40) |
| Production `vespasian` behavior | unchanged — 12 files changed, zero `.go` files, nothing under `pkg/`, `cmd/`, or `internal/` |

The same limitation is recorded in the PR description, `CLAUDE.md`, and `test/README.md`, so it stays discoverable after merge. Once LAB-5766 lands, those "manual step" caveats come out.

santiago-praetorian:
## Verification summary — head `8ef6d21`

Twelve rounds of `/capability-pr-review` (four parallel lanes: QUAL, TEST, SEC-BE, SEC-FE), plus a full test run and a mutation battery against every security control this PR exists to pin.

### Acceptance criteria (LAB-5064)

| AC | Status | Evidence |
|---|---|---|
| **AC1** — devcontainer ships runnable non-snap Chrome; `launcher.LookPath()` resolves it; `//go:build integration` launch/kill/close tests run | ⚠️ **Partially met** | Chrome runnable and non-snap, `LookPath` resolves it, integration tests pass 6/6 against real Chrome 150.0.7871.186. **Not met:** nothing in this repo invokes `install-chrome.sh` at image build — the devcontainer definition lives outside the repository, so no change here can satisfy it. Disclosed in `AGENTS.md` and `test/README.md`; tracked by **LAB-5766**. |
| **AC2** — `--group offline` runs to completion without a browser | ✅ **Met** | 21/21 on a deliberately browserless `PATH` (symlink farm omitting every chrome/chromium binary), not merely on a host that happens to lack one. |
| **AC3** — `--group live` runs against the provisioned browser; rod-backed targets incl. `no-download` execute rather than SKIP | ✅ **Met** | 10/10, `no-download` = **PASS**, **0 skipped**. |
| **AC4** — installing Chrome adds no background phone-home | ✅ **Met** (with one gap, below) | 14 assertions drive the real `suppress_permanent_repo` / `remove_phone_home` against a fixture root: both apt-source spellings (`.list` and deb822 `.sources`), `cron.daily/google-chrome`, the keyring, the failed-install path and the idempotent early-exit path. |
| **AC5** — no change to production vespasian behaviour | ✅ **Met** | Every changed path is `test/`, `.github/`, or agent docs. Zero changes to `pkg/`, `cmd/`, `internal/`, `go.mod`, `go.sum`, `Dockerfile`, `Makefile`. |

### Mutation testing — the PR's central claim, measured

The defect class this PR exists to remove is **an assertion that names or counts a control instead of exercising it**. Each control was removed and the suites re-run; every mutation was confirmed applied before its result was believed, and judged by exit code.

| Control | Mutation | Result |
|---|---|---|
| Trust anchor | fingerprint comparison → `if false` | ✅ caught |
| TLS pinning | `--proto/--proto-redir '=https'` dropped | ✅ caught |
| Keyring mode | `install -m 0644` → `0666` | ✅ caught |
| Phone-home suppression | `suppress_permanent_repo` → no-op | ✅ caught |
| Phone-home removal | `remove_phone_home` → no-op | ✅ caught |
| Privileged-call bound | `timeout -k 30 300` off `apt-get update` | ✅ caught |
| Privileged-call bound | `timeout -k 30 900` off `apt-get install` | ✅ caught |
| Lock guard ×5 | symlink / regular-file / hard-link / owner-unreadable / owner-uid → `if false` | ✅ caught (each) |
| Mutual exclusion | flock refusal → degrade-and-continue | ✅ caught |
| Bind-host (Go) | `127.0.0.1` → `0.0.0.0` | ✅ caught |
| Bind-host (Node) | `127.0.0.1` → `0.0.0.0` | ✅ caught |
| Assertion accounting | a whole case deleted | ✅ caught (drift: 243 ≠ 245) |

**15/15.** Remove any control and a suite goes red — because the assertion exercises the behaviour, not because it greps the control's name.

### Test results at this head

| Layer | Result |
|---|---|
| `gofmt` / `go vet` / `go build` / `make build` | clean |
| `golangci-lint` | **0 issues** |
| `go test -race ./...` | 20 packages ok, 0 failures |
| Coverage | **86.4%** (CI gate: 80%) |
| `preflight-selftest.sh` | 90 / 0 / 0 |
| `install-chrome-selftest.sh` | **245 / 0 / 0** |
| `setup-live-targets_test.sh` | exit 0 |
| `test-runner-args.sh` | 130 / 0 / 0 |
| `validate_test.sh` | 98 / 0 |
| `check-docs.py` | exit 0 |
| `run-live-tests.sh --group offline` | 21 / 21 |
| `run-live-tests.sh --group live` | 10 / 10 |
| `run-live-tests.sh` (default) | 32 / 32, **0 skipped** |
| `go test -tags integration ./pkg/crawl` | 6 / 6 vs real Chrome |
| CI at `8ef6d21` | **21 pass, 0 fail** |

Degraded-host behaviour, all reconciling against one pin of 245:

| Host shape | Result |
|---|---|
| equipped | 245 passed, 0 failed, 0 skipped — exit 0 |
| no `timeout(1)` | 223 + credits 22 = 245, 4 skipped — exit 0 |
| no `gpg` | 209 + credits 36 = 245, 5 skipped — exit 1 *(trust-anchor guard, by design)* |
| empty key fixture | 216 + credits 29 = 245, 3 skipped — exit 1 *(same, by design)* |

### Review convergence

Findings per round, all four lanes: **11 → 7 → 7 → 2 → 3 → 4 → 4**. Zero blockers in the last four rounds. From round 8 onward every finding was in the *test pins* rather than the code under test — three consecutive rounds found the lock-ordering pin incomplete one level deeper each time (a guard omitted → anchored on its `stat` assignment instead of its decision → the wrong arm of the one guard with two). That family was closed structurally in `8ef6d21`: all four checks now derive from a single anchor list, and the refusal counter matches the terminal `exit` rather than a diagnostic, so it is independent of how a future guard is spelled.

### Known open items

1. **AC1 / LAB-5766** — the installer is not wired into any image build. Not fixable in this repo.
2. **`install-chrome-e2e` has not run on this branch.** It is `workflow_dispatch`-gated, so the *real privileged install* and its seven-path phone-home assertion are unexercised here. AC4's evidence above is fixture-driven. Running it before merge would close the last gap.
3. **Trailing-comment inflation in the anchor-completeness counter** (`install-chrome-selftest.sh`). The counter matches `exit [0-9]` over `main()`'s body; `fn_code` strips whole-line comments but not trailing ones, so a trailing `# … exit 1 …` inside the lock span inflates the count and produces a *spurious* failure. Measured, still present at this head. Rated `nit`: the direction is safe (it fails loud, never silent), and no such comment exists in the file today — but the failure text says "add the anchor, do not adjust this count", which is the wrong advice in exactly that case.
4. Three items raised by lanes and deliberately **not** filed: the `/tmp` sticky bit is relied upon but never asserted (no `[ -k /tmp ]`); `head -1` is lenient if a guard appears twice (benign — the in-window occurrence still gates); a flock-less host is not green for pre-existing reasons that predate this PR.

AGENTS.md
```diff
@@ -126,13 +126,16 @@ The `test/` directory contains live test targets:
 - **test/grpc-server/**: Go gRPC server with Server Reflection enabled (sample User/Order/Account services, including a streaming method) for reflection-probe testing
 - **test/concat-spa/**: Go HTTP server serving a single-page app whose API endpoints exist only as concatenated strings in an external JS bundle (discovered via post-crawl JS-replay)
 - **test/forms-target/**: Go HTTP server serving an HTML `<form>` page whose POST endpoints are recovered by static form extraction (`analyze.ExtractForms`); a co-located GET search form contributes query parameters to a crawl-linked endpoint
+- **test/internal/target/**: shared listener setup for the Go live-test targets. `Addr()` resolves the bind address (loopback by default, widened only via `BIND_HOST`) and `Server()` applies the read-header and companion timeouts. The four Go targets above each carried a byte-identical copy of both before this was consolidated, so the loopback default — a security control for these unauthenticated fixtures — is now asserted in one place instead of four.
```
logan-bayless: Naming the four HTTP targets and stating why grpc-server is excluded removes the miscount, since grpc-server binds a gRPC listener directly and keeps its own hardcoded loopback literal rather than sharing the consolidated one.

```suggestion
- **test/internal/target/**: shared listener setup for the Go live-test targets. `Addr()` resolves the bind address (loopback by default, widened only via `BIND_HOST`) and `Server()` applies the read-header and companion timeouts. The four HTTP Go targets above (rest-api, soap-service, concat-spa, forms-target) each carried a byte-identical copy of both before this was consolidated. grpc-server is not among them: it binds a gRPC listener directly and keeps its own hardcoded loopback literal, so the loopback default — a security control for these unauthenticated fixtures — is now asserted in one place instead of four.
```

AGENTS.md
```diff
@@ -126,13 +126,16 @@ The `test/` directory contains live test targets:
...
 
+Supporting scripts: `test/install-chrome.sh` provisions a real non-snap Chrome for containers (Ubuntu's `chromium-browser` is a snap stub that cannot run in-container), installing it from Google's apt repo with the signing key pinned by fingerprint and the repo removed again afterwards. It is **not wired into any image build** — nothing calls it automatically, because the devcontainer definition lives outside this repo; a fresh container is browserless until an image layer or `postCreateCommand` invokes it. That wiring is tracked in LAB-5766. `test/preflight-selftest.sh`, `test/test-runner-args.sh`, `test/install-chrome-selftest.sh`, and `test/setup-live-targets_test.sh` are the CI-run regression guards for the setup preflight, the target-group/dispatch wiring (plus the un-gated job's own step list and the browser-target classification), the installer's non-privileged surface, and the setup script's teardown / orphan-process hardening (LAB-2893). All four run in the un-gated `preflight-selftest` job, which the `test` job depends on so a guard failure stops the full suite before it builds anything.
```
logan-bayless: Narrowing the claim to image provisioning makes it true, since `install-chrome-e2e` does invoke the installer unattended on every push to main.

```suggestion
Supporting scripts: `test/install-chrome.sh` provisions a real non-snap Chrome for containers (Ubuntu's `chromium-browser` is a snap stub that cannot run in-container), installing it from Google's apt repo with the signing key pinned by fingerprint and the repo removed again afterwards. It is **not wired into any image build** — nothing provisions it into a container image automatically, because the devcontainer definition lives outside this repo, and the only automated invocation is the opt-in `install-chrome-e2e` CI job; a fresh container is browserless until an image layer or `postCreateCommand` invokes it. That wiring is tracked in LAB-5766. `test/preflight-selftest.sh`, `test/test-runner-args.sh`, `test/install-chrome-selftest.sh`, and `test/setup-live-targets_test.sh` are the CI-run regression guards for the setup preflight, the target-group/dispatch wiring (plus the un-gated job's own step list and the browser-target classification), the installer's non-privileged surface, and the setup script's teardown / orphan-process hardening (LAB-2893). All four run in the un-gated `preflight-selftest` job, which the `test` job depends on so a guard failure stops the full suite before it builds anything.
```

AGENTS.md
```diff
@@ -153,6 +156,20 @@ See `test/README.md` for how to run the suite, including the `TEST_HOST` overrid
...
+- **live-tests.yml**: six jobs — three un-gated guard jobs, an opt-in end-to-end installer job, a label gate, and the full suite.
+
+  **`preflight-selftest`** (un-gated) is the shell regression net. It runs a `bash -n` syntax check and then all four guard suites: `test/preflight-selftest.sh` (Chrome/Chromium detection, LAB-3893), `test/install-chrome-selftest.sh` (the installer's non-privileged surface), `test/setup-live-targets_test.sh` (teardown / orphan-PID hardening, LAB-2893), and `test/test-runner-args.sh` (target-group vs dispatch drift, the un-gated job's own step list, browser-target classification, and TEST-023: every assertion whose message claims one statement comes BEFORE another must contain an operator comparing their two positions). None need Go, Node, or Chrome. `test-runner-args.sh` fails CI if a dispatch target is not covered by `OFFLINE_TARGETS`, `LIVE_TARGETS`, or the config-only set (e.g. `grpc-server`), if this job stops invoking one of the four suites, or if a target in `ALL_TARGETS` is classified neither browser nor non-browser.
```
logan-bayless: Replacing the parenthetical with the two guards that exist removes a documented safety property nothing implements, since `git grep 'comes BEFORE'` over the tree returns only this sentence.

```suggestion
  **`preflight-selftest`** (un-gated) is the shell regression net. It runs a `bash -n` syntax check and then all four guard suites: `test/preflight-selftest.sh` (Chrome/Chromium detection, LAB-3893), `test/install-chrome-selftest.sh` (the installer's non-privileged surface), `test/setup-live-targets_test.sh` (teardown / orphan-PID hardening, LAB-2893), and `test/test-runner-args.sh` (target-group vs dispatch drift, the un-gated job's own step list, browser-target classification, and two further guards: the `test` job is excluded from the gated-job allowlist, so a suite hosted only there cannot satisfy the wiring check while staying skippable via `skip-live-tests`, and `install-chrome-e2e`'s verification steps are pinned by name so deleting one fails CI). None need Go, Node, or Chrome. `test-runner-args.sh` fails CI if a dispatch target is not covered by `OFFLINE_TARGETS`, `LIVE_TARGETS`, or the config-only set (e.g. `grpc-server`), if this job stops invoking one of the four suites, or if a target in `ALL_TARGETS` is classified neither browser nor non-browser.
```

AGENTS.md
```diff
@@ -153,6 +156,20 @@ See `test/README.md` for how to run the suite, including the `TEST_HOST` overrid
...
+  All four real jobs (`preflight-selftest`, `validator-regression`, `docs-check`, `test`) open with a `step-security/harden-runner` step in `egress-policy: audit` mode (LAB-4732 / SEC-BE-002) — defense-in-depth network monitoring that logs (does not block) outbound traffic; `registry.npmjs.org` is part of the anticipated allowlist. The `--ignore-scripts` flag on both npm installs is SEC-BE-001: it blocks npm lifecycle scripts from executing under the non-blocking audit-only egress policy.
```
logan-bayless: Saying "four non-container jobs" and naming the exception makes the count correct and discloses that the one job running as root with live egress carries no harden-runner.

```suggestion
  The four non-container jobs (`preflight-selftest`, `validator-regression`, `docs-check`, `test`) open with a `step-security/harden-runner` step. `install-chrome-e2e` carries none because the action does not support container jobs, which matters because it is the one job running as root with live egress to Google's apt repo in `egress-policy: audit` mode (LAB-4732 / SEC-BE-002) — defense-in-depth network monitoring that logs (does not block) outbound traffic; `registry.npmjs.org` is part of the anticipated allowlist. The `--ignore-scripts` flag on both npm installs is SEC-BE-001: it blocks npm lifecycle scripts from executing under the non-blocking audit-only egress policy.
```

.github/workflows/live-tests.yml
```diff
@@ -131,6 +167,144 @@ jobs:
...
+  # every rejection arm unprivileged but deliberately stops before the first
+  # $SUDO, so this region shipped with no automated coverage at all: a change
+  # that broke the actual install would pass every existing suite.
+  #
+  # Not run on every PR. It needs root, network egress to Google, and it mutates
+  # system state — all of which are fine here because the container is thrown
+  # away, but it is slow and externally-dependent, so it is opt-in via
+  # workflow_dispatch (and on push to main, where a break should not stay hidden).
+  install-chrome-e2e:
+    if: github.event_name == 'workflow_dispatch' || (github.event_name == 'push' && github.ref == 'refs/heads/main')
```
logan-bayless, line 185: Dispatching this job once from the PR head is the only thing that produces CI evidence for the 1371-line privileged installer, since the gate excludes `pull_request` while its first arm matches any ref.

.github/workflows/live-tests.yml
```diff
@@ -131,6 +167,144 @@ jobs:
...
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
+    # Restated at job level even though the workflow default already grants only
```
logan-bayless, line 188: Restating the comment as containment rather than detection stops it promising a signal that a job-level `permissions` block cannot emit.

.github/workflows/live-tests.yml
```diff
@@ -131,6 +167,144 @@ jobs:
...
+        run: |
+          apt-get update -qq
+          apt-get install -y --no-install-recommends ca-certificates curl gnupg
+
+      # Runs as root, so SUDO resolves empty and every privileged branch is
+      # exercised directly.
+      - name: Install Chrome end-to-end
+        run: ./test/install-chrome.sh
+
+      - name: Assert the install actually produced a runnable browser
```
logan-bayless, line 238: Extracting these assertions to a script and adding it to the `bash -n` list at :53 puts them inside the un-gated job's reach, which is the only way a syntax error in them fails before a push to main.

.github/workflows/live-tests.yml
```diff
@@ -131,6 +167,144 @@ jobs:
...
+            echo "recorded build: (no version record written)"
+          fi
+          exit "$fail"
+
+      # Second run must take the idempotent early-exit path and stay clean.
+      - name: Re-run is idempotent
+        run: |
+          set -euo pipefail
+          out=$(./test/install-chrome.sh 2>&1)
+          printf '%s\n' "$out"
```
logan-bayless: Adding the explicit handler preserves the installer's captured output on the failure path, which errexit currently discards before the `printf` runs.

```suggestion
          out=$(./test/install-chrome.sh 2>&1) || { printf '%s\n' "$out"; echo "second install-chrome.sh run failed" >&2; exit 1; }
          printf '%s\n' "$out"
```

test/install-chrome.sh
```diff
@@ -0,0 +1,1371 @@
...
+# it never gated which source apt picked for the package NAME, and the pin
+# file alone raises that source's priority but does not prove apt actually
+# used it. `apt-cache policy` is read-only, so this needs no $SUDO.
+verify_apt_origin() {
+    local policy candidate origin host
+    # `|| policy=""` guards the assignment itself: under `set -euo pipefail`, a
+    # failing apt-cache (a held dpkg lock, a corrupted cache) would otherwise
+    # abort the script here with no diagnostic instead of reaching the
+    # log_fail below.
+    policy=$(apt-cache policy google-chrome-stable 2>/dev/null) || policy=""
```
logan-bayless: Bounding this call with the same timeout its two siblings carry closes the window where a blocking `apt-cache` leaves a fully-trusted Google source live in `/etc` with cleanup unreached.

```suggestion
    policy=$($SUDO timeout -k 5 30 apt-cache policy google-chrome-stable 2>/dev/null) || policy=""
```

test/install-chrome.sh
```diff
@@ -0,0 +1,1371 @@
...
+# Kubernetes) answered false here, so an image built by any of THOSE shipped
+# the exact standing egress AC4 exists to eliminate — silently, since
+# verify_install shares this same predicate for its audit (see its call site).
+# All four probes are reachable from the selftest: the two marker paths are
+# rerooted by VESPASIAN_TEST_ROOT, and REMOTE_CONTAINERS / container are read
+# from the environment.
+in_container() {
+    [ -f "${TEST_ROOT}/.dockerenv" ] && return 0
+    [ -f "${TEST_ROOT}/run/.containerenv" ] && return 0
+    # SEC-BE-003: validate this the way $container is validated below rather than
```
logan-bayless: Dropping the per-round `SEC-BE-###` ids leaves the surrounding prose, which already states the defect and the mutation, so the comment stops citing a label that resolves to four different guards in this file.

test/install-chrome-selftest.sh
```diff
@@ -0,0 +1,3975 @@
+#!/usr/bin/env bash
+# Copyright 2026 Praetorian Security, Inc.
+#
+# Self-test for test/install-chrome.sh (LAB-5064). Plain bash, no framework —
+# same shape as preflight-selftest.sh: source the script (its BASH_SOURCE guard
+# means main() does not run), then exercise the pure helpers.
+#
+# Covered (TEST-004: kept in sync with the actual case letters, not just the
```
logan-bayless: Deriving the Covered list from the `# ── Case X:` headers makes the claim checked rather than believed, which is what case `cr` already does for the credit register; separately, replacing each review-id subject with its durable reason keeps every comment's argument while removing 429 references that resolve to nothing outside the round that produced them.

test/install-chrome-selftest.sh
```diff
@@ -0,0 +1,3975 @@
...
+# match nothing and the count-match check above would pass vacuously at 0-of-0.
+if [ "${f5_aptget_lines}" -ge 1 ]; then
+    echo "PASS: case f: the apt-get call extraction still finds at least one call in the script"
+    pass_count=$((pass_count + 1))
+else
+    echo "FAIL: case f: the script no longer contains a \$SUDO timeout ... apt-get call — the bound-count check above is now vacuous, fix the extraction rather than deleting it"
+    fail_count=$((fail_count + 1))
+fi
+
+# f1: the endpoint returns something that is not a PGP key at all.
```
logan-bayless, line 1027: Adding the two stubbed-gpg cases gives `install_pinned_key`'s empty-fingerprint and failed-export arms their first assertions, so deleting either refusal now reds the suite.

test/install-chrome-selftest.sh
```diff
@@ -0,0 +1,3975 @@
...
+            main
+        ) 2>&1
+    )
+    rc_f4=$?
+    set -e
+    assert_eq "case f: main() aborts on a fingerprint mismatch before any further mutation (rc 1)" \
+        "1" "${rc_f4}"
+    assert_contains "case f: the abort is the fingerprint mismatch, not something else" \
+        "fingerprint mismatch" "${res_f4}"
+    if [ -e "${root_f4}/etc/default/google-chrome" ]; then
```
logan-bayless, line 1191: Asserting that the rerooted defaults path actually holds `repo_add_once=false` replaces case `f4`'s negative check, which is satisfied for any value of the constant.

test/install-chrome-selftest.sh
```diff
@@ -0,0 +1,3975 @@
...
+printf 'repo_add_once=true\n' > "${FIXTURE_DIR}/defaults-existing"
+res_g2=$(run_suppress "${FIXTURE_DIR}/defaults-existing")
+assert_eq "case g: an existing defaults file is accepted (rc 0)" "0" "$(echo "${res_g2}" | sed -n '1p')"
+assert_eq "case g: repo_add_once=true is rewritten to false" \
+    "repo_add_once=false" "$(cat "${FIXTURE_DIR}/defaults-existing")"
+# Mode actually achieved on disk (SEC-BE-003) -- a literal `install -m 0644`
+# in the source is not evidence of what lands; only stat proves it, same
+# reasoning as case j's three keyring/list/pref mode assertions below.
+assert_eq "case g: the rewritten defaults file is world-readable but not writable (0644)" \
+    "644" "$(stat -c '%a' "${FIXTURE_DIR}/defaults-existing" 2>/dev/null)"
```
logan-bayless: A `stat` fallback keeps all 40 case groups running on a BSD userland, where `stat -c` currently aborts the suite after 26.

test/install-chrome-selftest.sh
```diff
@@ -0,0 +1,3975 @@
...
+        # assume. Scoped, this case's gpg homedir starts empty every run,
+        # mirroring round 5's fix for the same hazard in the PATH stub bins.
+        mkdir -p "${FIXTURE_DIR}/j-scratch"
+        out=$(install_pinned_key "${FIXTURE_DIR}/j-scratch" 2>&1)
+        printf '%s\n%s\n' "$?" "${out}"
+    )
+    assert_eq "case j: the real Google key satisfies the pin (rc 0)" \
+        "0" "$(echo "${res_j}" | sed -n '1p')"
+    # The load-bearing assertion: signed-by= is what makes the pin gate apt.
+    assert_contains "case j: the emitted apt source pins signed-by to our keyring" \
```
logan-bayless: Asserting the full emitted line instead of two fragments makes an added `trusted=yes` fail, which containment cannot see and which would retire the fingerprint pin.

test/install-chrome-selftest.sh
```diff
@@ -0,0 +1,3975 @@
...
+
+# The two SUCCESS paths had no coverage at all -- neither is hypothetical, both
+# feed every privileged command in the script. sudo-present: SUDO must become
+# exactly the string "sudo", not merely rc 0 (rc 0 alone is also what a broken
+# resolve_sudo that always "succeeds" would produce).
+mkdir -p "${FIXTURE_DIR}/sudo-bin"
+cp "${FIXTURE_DIR}/empty-bin/id" "${FIXTURE_DIR}/sudo-bin/id"
+printf '#!/bin/bash\nexec "$@"\n' > "${FIXTURE_DIR}/sudo-bin/sudo"
+chmod +x "${FIXTURE_DIR}/sudo-bin/sudo"
+res_l3=$(run_resolve_sudo "${FIXTURE_DIR}/sudo-bin")
```
logan-bayless, line 1931: Giving `sudo-bin/id` a fixed non-zero uid keeps case `l`'s sudo-discovery branch live on a root host, where the copied real-uid stub currently sends `resolve_sudo` down its already-root path.

test/install-chrome-selftest.sh
```diff
@@ -0,0 +1,3975 @@
...
+    (
+        # shellcheck source=install-chrome.sh
+        source "${INSTALL_SCRIPT}"
+        PATH="$1"
+        set +e
+        out=$(require_tools 2>&1)
+        printf '%s\n%s\n' "$?" "${out}"
+    )
+}
+res_m=$(run_require_tools "${FIXTURE_DIR}/empty-bin")
```
logan-bayless, line 1984: A case `m2` driving `require_tools` without `timeout` gives that arm its first assertion, so removing the coreutils check no longer leaves the suite green.

test/install-chrome-selftest.sh
```diff
@@ -0,0 +1,3975 @@
...
+# must now mean "a browser exists AND nothing phones home".
+# main() calls resolve_sudo, which overwrites any SUDO we preset, so a bare
+# override cannot keep these writes unprivileged. A passthrough `sudo` stub on
+# PATH is what does: resolve_sudo finds it and every $SUDO call runs as the test
+# user, leaving no root-owned files in the fixture tree.
+#
+# Own dir (bin-o), not the shared FIXTURE_DIR/bin (TEST-002): case o2 below
+# installs its OWN sudo shim rather than inheriting this one, so the two cases
+# do not silently depend on which order they run in.
+mkdir -p "${FIXTURE_DIR}/bin-o"
```
logan-bayless, line 2287: Stubbing `flock` in the five fixture bins lets the eight `main()`-driving cases run on a flock-less host instead of failing uncredited.

test/install-chrome-selftest.sh
```diff
@@ -0,0 +1,3975 @@
...
+) >/dev/null 2>&1
+kept_u3=1
+for f in "${root_u}/etc/apt/sources.list.d/google-chrome.list" \
+         "${root_u}/etc/cron.daily/google-chrome"; do
+    [ -e "$f" ] || kept_u3=0
+done
+assert_eq "case u: a SUCCESSFUL install (INSTALL_SUCCEEDED=1) leaves main()'s decision alone" \
+    "1" "${kept_u3}"
+
+# Ordering guard, and the reason this case exists at all. INSTALL_SUCCEEDED must
```
logan-bayless, line 2892: Asserting that `INSTALL_ATTEMPTED=1` precedes `apt-get install` pins the ordering that keeps `cleanup_all` able to remove phone-home artifacts after a mid-install failure.

test/install-chrome-selftest.sh
```diff
@@ -0,0 +1,3975 @@
...
+  Candidate: 999.0.0.0-1
+  Version table:
+     999.0.0.0-1 500
+        500 https://dl.google.com/linux/chrome/deb stable/main amd64 Packages
+POLICY
+POLICY_STUB
+    # Simulates dpkg's postinst: the actual moment the phone-home artifacts and
+    # a runnable browser binary appear on a real host. Only "install" plants
+    # them, so `apt-get update` (and anything else) stays a no-op.
+    cat > "${bin_v}/apt-get" <<EOF
```
logan-bayless, line 3053: Parameterising case `v`'s apt-get stub with a fail-step argument gives both failure diagnostics their first assertions and proves the later stage does not run.

test/install-chrome-selftest.sh
```diff
@@ -0,0 +1,3975 @@
...
+assert_eq "case w: a bare-integer version cannot be satisfied by a priority-column collision (rc 1)" \
+    "1" "$(echo "${res_w4b}" | sed -n '1p')"
+assert_contains "case w: the priority-collision fixture is refused as evil.example, not accepted as dl.google.com" \
+    "evil.example" "${res_w4b}"
+
+# An apt-cache that produces nothing (held dpkg lock, corrupted cache) must be
+# refused with a diagnostic rather than aborting the script under errexit — the
+# `|| policy=""` guard exists for exactly this, and without a case the guard's
+# absence would surface as a silent abort rather than a failure (QUAL-006).
+res_w5=$(run_verify_apt_origin_raw '' emptypolicy)
```
logan-bayless, line 3495: Giving the candidate gate its own wording lets case `w5` pin that gate specifically, rather than a message both gates emit.

test/install-chrome-selftest.sh
```diff
@@ -0,0 +1,3975 @@
...
+# looking like protection.
+#
+# What DOES bite is asserting the tolerance is present in the handler's text.
+# The property is a syntactic one — each step is `||`-guarded by something
+# that itself always succeeds (`true`, or `log_warn`'s printf, per SEC-BE-005:
+# a bare `|| true` swallowed a removal failure with no diagnostic at all) — so
+# a source-level check is the honest form, and it is the same derive-from-
+# source idiom the CI-wiring guards in test-runner-args.sh already use.
+cleanup_all_src=$(sed -n '/^cleanup_all() {/,/^}/p' "${INSTALL_SCRIPT}")
+for guarded_step in remove_phone_home cleanup_apt_wiring; do
```
logan-bayless: Counting `cleanup_all`'s steps from its own body makes the tolerance check cover a third step automatically, which enumerating two cannot.

test/preflight-selftest.sh
```diff
@@ -191,6 +305,124 @@ else
...
+# That is not a cosmetic gap. `/usr/bin/chromium-browser` IS the Ubuntu snap
+# launcher stub named in LAB-5064's own Evidence block and in its parent
+# LAB-3893, and it does NOT live under /snap/. Losing that member sends a
+# developer hitting the exact originating symptom to the generic "check
+# permissions, missing shared libraries" hint instead of the actionable snap
+# diagnosis — reintroducing the confusion LAB-3893 was filed to remove.
+#
+# Derived from the source, not hardcoded, so the test cannot drift from the arm
+# it guards.
+snap_glob=$(grep -oE '^\s+\*/snap/\*[^)]*\)' "${SETUP_SCRIPT}" | head -1 | tr -d ' )')
```
logan-bayless: Unioning every arm that emits the snap hint stops a behaviour-preserving arm split from halving the member set while the extraction sentinel still passes.

test/run-live-tests.sh
```diff
@@ -115,9 +135,55 @@ resolve_targets() {
...
+}
+
 # Source shared colors, logging, and validation functions
 # shellcheck source=common.sh
 source "${SCRIPT_DIR}/common.sh"
+
+# Announced HERE, not at the VESPASIAN default above: common.sh (which defines
+# log_warn) is not sourced until this line, so the notice cannot be emitted
+# earlier without calling an undefined function.
+if [ -n "$_vespasian_from_env" ]; then
```
logan-bayless, line 184: Three assertions against the output `test-runner-args.sh` already captures give the ambient-override notice its first coverage, so deleting all three pieces reds the suite.

test/setup-live-targets_test.sh
```diff
@@ -468,12 +511,951 @@ if command -v pgrep >/dev/null 2>&1; then
...
+    if [ "$gql_install_n" -gt 1 ]; then
+        # Blank the extraction too. fail() only increments a counter and returns,
+        # so without this the block fell through and eval'd head -1's arbitrary
+        # pick anyway — loud AND arbitrary, while the comment claimed otherwise.
+        # Empty routes control to the [ -z ] arm below, which reports the
+        # extraction failure instead of acting on a coin-flip.
+        fail "start_graphql_server has ${gql_install_n} install(1) statements targeting .graphql-server.log — the extraction above would pick one arbitrarily; disambiguate rather than trusting head -1 (SEC-BE-003)"
+        gql_install_line=""
+    fi
+    if [ -z "$gql_install_line" ]; then
```
logan-bayless: Crediting the displaced outcomes makes Test 22's extraction failure report one cause instead of also tripping the accounting pin.

test/test-runner-args.sh
```diff
@@ -428,6 +461,465 @@ else
...
+#
+# run-live-tests.sh's parse loop has a catch-all that names the offending option,
+# prints usage and exits 1 — and NO suite asserted it. Replacing the catch-all
+# with `*) shift ;;` left this suite at 110/0, rc 0: a typo'd flag would then be
+# silently ignored and the run would proceed with the WRONG target selection
+# while reporting success, which for `--group offline` vs `--group live` is the
+# difference between running the browser tests and not.
+#
+# Both sibling suites already pin this for their own script
+# (install-chrome-selftest case b, setup-live-targets_test Test 22), so this was
```
logan-bayless: Pointing the reference at Test 14 sends a reader to the test that actually holds the property.

```suggestion
# (install-chrome-selftest case b, setup-live-targets_test setup-live-targets_test Test 14 (parse_args unknown option)), so this was
```

test/test-runner-args.sh
```diff
@@ -428,6 +461,465 @@ else
...
+# has a usable browser.
+#
+# This is a DRIFT guard, in the same spirit as the target-group check above.
+# Both greps below are structural: they are satisfied by a body that calls
+# detect_chrome_binary and throws the answer away (TEST-021), e.g.
+# `detect_chrome_binary >/dev/null 2>&1; return 0`. Make the RESULT
+# load-bearing by extracting chrome_available the same way targets_need_config
+# is above — a sed range over the source, not the whole runner (which calls
+# `main "$@"` unguarded) — and driving it against a fixture CHROME_CANDIDATES.
+chrome_avail_body=$(awk '/^chrome_available\(\) \{/,/^\}/' "$RUNNER")
```
logan-bayless, line 766: A position-based pin requiring the first resolver statement to be the `detect_chrome_binary` call catches the multi-line presence-only fast path that all three text pins miss.

test/test-runner-args.sh
```diff
@@ -510,7 +1013,1002 @@ fi
...
+    # say why in the commit — that edit is exactly the review moment this pin exists
+    # to force.
+    EXPECTED_PR_TRIGGER='{"branches":["main"],"types":["opened","synchronize","reopened","labeled","unlabeled"]}'
+    actual_pr_trigger="$(yq_query '.on.pull_request | sort_keys(..)' -o=json -I=0)"
+    case "$actual_pr_trigger" in
+        __NO_YQ__) fail_no_yq "live-tests.yml's pull_request trigger shape" ;;
+        __YQ_ERROR__) fail_yq_error "live-tests.yml's pull_request trigger shape" ;;
+        "$EXPECTED_PR_TRIGGER")
+            pass "live-tests.yml's pull_request trigger is unnarrowed (no paths/paths-ignore/types/branches filter that would stop a shell-only PR running the guard suites)" ;;
+        *)  fail "live-tests.yml's pull_request trigger shape changed: expected ${EXPECTED_PR_TRIGGER}, got ${actual_pr_trigger} — a paths/paths-ignore/types/branches narrowing switches off every un-gated guard suite for the PRs they exist to police; if the change is deliberate, update EXPECTED_PR_TRIGGER" ;;
```
logan-bayless: Rewording to the property actually pinned removes a false claim, since the trigger carries both a `branches` and a `types` filter.

```suggestion
            pass "live-tests.yml's pull_request trigger carries no paths/paths-ignore narrowing, so a shell-only PR still runs the un-gated guard suites (branches and types are expected and are part of EXPECTED_PR_TRIGGER)" ;;
        *)  fail "live-tests.yml's pull_request trigger shape changed: expected ${EXPECTED_PR_TRIGGER}, got ${actual_pr_trigger} — a paths/paths-ignore/types/branches narrowing switches off every un-gated guard suite for the PRs they exist to police; if the change is deliberate, update EXPECTED_PR_TRIGGER" ;;
```

logan-bayless (changes requested):
## Issues identified

29 defects introduced by this PR, inline below. One has no line to attach to:

- **This PR cannot merge as it stands.** `mergeable: CONFLICTING`, one conflicted file, `AGENTS.md`, from `80f9451` (LAB-4678 / #189) landing on main. `.github/workflows/live-tests.yml` auto-merges cleanly.

## Not this PR — follow-up tickets needed

Four further real defects sit on code this PR did not write. Each needs its own ticket rather than a fix here:

- `AGENTS.md:158` — claims ci.yml runs on all pushes and PRs; its `paths` filter is Go-only. From `c5e8ed6` (ENG-5577).
- `test/setup-live-targets.sh:728` — `orphan_pids_by_name` returns success with no output when `pgrep` is absent, so `--teardown --sweep` reports a clean teardown while unauthenticated targets keep listening. From `de936ea`.
- `test/setup-live-targets_test.sh:152` — `spawn_named` has no liveness premise, so the nine `assert_dead` assertions can pass vacuously. LAB-2893.
- `test/test-runner-args.sh:194` — the drift-guard block emits a variable outcome count, so real drift also trips the accounting pin with misattributing text. LAB-4773.

## Inline suggestions

Eight of the 29 apply with one click: `AGENTS.md` :129, :134, :161, :175; `live-tests.yml` :304-305; `install-chrome.sh` :506; `test-runner-args.sh` :677 and :1316-1317. The remaining 21 need a second edit in a distant location, most often `EXPECTED_ASSERTIONS` at `install-chrome-selftest.sh:85`.

logan-bayless (approved):
## Approved

All 29 findings from review 4975493367 are addressed, and the fixes were verified against `331211f9` file by file rather than taken from the commit messages. The merge conflict is gone.

### Correcting one of my own findings

My comment on `install-chrome-selftest.sh:924` said a `stat` fallback would keep "all 40 case groups running on a BSD userland, where `stat -c` currently aborts the suite after 26." The fallback landed and does fix ten real mode assertions, but the abort claim was a wrong causal attribution on my part. The terminating call is `lock_owner=$(stat -c '%u' …)` in `install-chrome.sh`'s lock guard, reached through `main()` by case `p2`'s `trap_probe=$(run_trap_probe …)` assignment, which takes the non-zero status under the suite's own errexit. Measured on one host: `8ef6d216` recorded 132 outcomes and aborted at case `p`; `331211f9` records 146 and aborts at the same point for the same reason. `install-chrome.sh` provisions a `.deb`, so GNU `stat` inside it is correct, and the suite inherits the dependency by driving `main()`. Nothing owed here.

### Two new items, non-blocking, both macOS-only

Recorded on LAB-5986 rather than held against this PR. Both are green on ubuntu-24.04 and neither affects CI.

1. `install-chrome-selftest.sh:1727` — case `j`'s new equality assertion builds its expectation from the raw `${root_j}`, but `install-chrome.sh:168` canonicalises `VESPASIAN_TEST_ROOT` through `cd -P … && pwd -P` as a confinement control. Where the fixture root carries a symlink component the two disagree: expected `/var/folders/…/root-j/…`, got `/private/var/folders/…/root-j/…`. Resolving `root_j` the same way the installer does closes it. This only surfaced because the review asked for equality; the two `assert_contains` fragments it replaced were insensitive to it.

2. `install-chrome-selftest.sh:412` — `FIXTURE_DIR=$(TMPDIR=/tmp mktemp -d)`. The comment above states that `/tmp`'s sticky bit is "the property being relied on" for a tree of stub executables every case prepends to `PATH`. Darwin's `mktemp -d` with no template implies `-t` and resolves the per-user temp dir, so the pin does not take effect there: `TMPDIR=/tmp mktemp -d` returns `/var/folders/kk/…/T/tmp.jHD0jKP72a`. GNU `mktemp` honours `TMPDIR`, so the control holds in CI. `mktemp -d /tmp/…XXXXXX` or `-p /tmp` makes it hold everywhere.

### Merge state

`security / Gosec` is red, and it is not this PR. The same job fails on `main` at `c0bb1676`, `d2a4e90a` and `f6c0624e` with a gosec internal error, `internal error: package "encoding/xml" without types was imported from "command-line-arguments"` here and `"bytes"` on main. That is what has `mergeStateStatus` at `BLOCKED`.

One note before merging: `install-chrome-e2e`'s green evidence is dispatch run `32389325316` at `c65fcc54`, and `test/assert-chrome-install.sh` changed after it, with the `--dump-dom` call gaining a `timeout -k 5 30` wrapper through `timeout_cmd`. The wrapper is errexit-safe, since `timeout_cmd` returns 0 with empty output when neither binary exists, so the risk is low, but the privileged path at head has not itself been run. One more dispatch would close that.

PR #204, opened 2026-08-02
chore: make AGENTS.md canonical, CLAUDE.md a pointer (ENG-5577)

nsportsman:
## What

Makes `AGENTS.md` the canonical instruction file for this repo and reduces
`CLAUDE.md` to the one-line `@AGENTS.md` import pointer.

- `AGENTS.md` — the previous `CLAUDE.md` content, **byte-identical**.
- `CLAUDE.md` — now the 11-byte pointer `@AGENTS.md` (Anthropic-documented import syntax).
- `.gemini/settings.json` — `context.fileName` opt-in so Gemini reads `AGENTS.md`.

## Why

This repo is a nested clone that carried a **Claude-only** root `CLAUDE.md`.
Codex and Gemini sessions opened here ran **uninstructed**. After this change
all three harnesses load the same instructions:

| Harness | How it reaches the content |
| --- | --- |
| Claude Code | `CLAUDE.md` → `@AGENTS.md` import |
| Codex | reads `AGENTS.md` natively |
| Gemini | `.gemini/settings.json` → `context.fileName` |

The direction (pointer in `CLAUDE.md`, content in `AGENTS.md`) is forced by
[anthropics/claude-code#6235](https://github.com/anthropics/claude-code/issues/6235):
Claude Code does not read `AGENTS.md` natively, so `CLAUDE.md` must remain.

## How

Produced by the scripted, idempotent conversion merged in
[praetorian-inc/palatine#485](https://github.com/praetorian-inc/palatine/pull/485)
— the same conversion applied to guard-core in
[praetorian-inc/guard#7319](https://github.com/praetorian-inc/guard/pull/7319).
Only git-tracked files are touched; the script never merges and refuses on any
conflict. **No instruction text was rewritten** — content moved verbatim.

## Verification

- Byte-exactness confirmed: `origin/main:CLAUDE.md` vs the new `AGENTS.md` compare **identical** via `cmp`.
- The script's read-only `--check` mode passes on the result.

## Premise gate

**SKIP recorded.** The trigger is "builds non-trivial machinery"; this PR builds
none. It is a byte-identical file move plus an 11-byte pointer, generated by an
already-reviewed script — no new layer, no new subsystem, no newly-authored
behavior-defining prose.

## Scope

Part of ENG-5577 (deliberately **non-closing**). That ticket covers 12 nested
clones; this PR converts one of them, so it must not auto-close the ticket while
the others are unlanded. Merge is a human decision.


nsportsman:
## Adjudication — round 1

Every automated finding on this PR is dispositioned below as either **confirmed** (reproduced concretely, then fixed) or **refuted** (with the specific evidence it does not hold). No reviewer is blind-accepted, and none is blind-dismissed.


### REFUTED — "Claude Code does not parse `@AGENTS.md`, so all project context is lost"

This is the most-repeated finding across the 12 conversion PRs, and it is directly testable. A/B run in a converted repo with file tools **disabled**, so the session cannot answer by reading `AGENTS.md` itself.

That last part is load-bearing. My first attempt at this test left tools enabled, and **both** arms answered correctly — by grepping the file. That control was vacuous and proved nothing about context loading, so I re-ran it with `--disallowed-tools`.

Probe: *"Per this repository, which numeric exit code means 'estimate exceeds budget'? Reply with the bare number only, or NONE if you have no repository instructions."* The answer appears only in `AGENTS.md`.

| Arm | `CLAUDE.md` contains | Result |
| --- | --- | --- |
| **A — as shipped** | `@AGENTS.md` | **`3`** (correct) |
| **B — control** | `AGENTS.md` (the `@` stripped) | *"this session has no Read/Grep/Glob/Bash tools available, so I can't open `AGENTS.md` … Rather than guess a number, I'd rather tell you that."* |

Arm A answers with no file access whatsoever, so the import was expanded into context before the turn began. Arm B — the identical file minus one character — cannot. That is the mechanism working, measured rather than assumed.

`@path` is Anthropic-documented import syntax (`docs.claude.com/en/docs/claude-code/memory`, "CLAUDE.md imports"). The direction here is also forced rather than chosen: Claude Code does **not** read `AGENTS.md` natively (anthropics/claude-code#6235), which is exactly why `CLAUDE.md` has to remain as a pointer instead of being deleted.

### REFUTED — "use a git symlink (`ln -s AGENTS.md CLAUDE.md`) instead"

Two independent reasons.

**1. The symlink degrades into precisely the failure this finding warns about.** A mode-`120000` symlink checked out with `core.symlinks=false` — the default on Git for Windows, and the case on some CI runners — materialises as a *regular text file whose entire content is the string* `AGENTS.md`. That is byte-for-byte Arm B of the table above: the arm that loses all context. So the proposal swaps a documented, measured mechanism for one that fails silently on a subset of checkouts, and fails in the exact way the finding is worried about.

**2. The conversion script rejects symlinks deliberately.** It refuses to follow or create them, a guard added after a containment-escape defect was found and fixed during review of praetorian-inc/palatine#485. Hand-adding one here would route around that guard and diverge from the other 11 repos in this batch.

### REFUTED — "`.gemini/settings.json` lists `GEMINI.md`, which does not exist in the repository"

The fact is correct — no `GEMINI.md` exists — but it is not a defect. `context.fileName` is a **candidate list**, not a set of required files: Gemini loads whichever entries exist and ignores the rest.

Measured with `gemini -p "/memory show"`, which makes the CLI report its own loaded context rather than having me infer it from an answer:

| Arm | `.gemini/settings.json` | `/memory show` reports |
| --- | --- | --- |
| **A — as shipped** | `{"context":{"fileName":["AGENTS.md","GEMINI.md"]}}` | `Current memory content from 1 file(s):` … `Context from: …/AGENTS.md` |
| **B — control** | `fileName: ["NOSUCHCONTEXTFILE.md"]` | `Memory is currently empty.` |
| **C — control** | file removed entirely | `Memory is currently empty.` |

Arm A loads `AGENTS.md` and reports **1 file**, raising no error for the absent `GEMINI.md` — the entry is inert. It is also deliberate: it keeps a repo-local `GEMINI.md` working if one is ever added, without a settings change.

Arms B and C exist so the measurement can fail — without them, "it loaded" would be unfalsifiable. They also establish something worth stating: the settings file is load-bearing. Remove it and Gemini loads **nothing at all** in this repo, which is the pre-conversion status quo this PR is fixing.


---

**Cross-repo note.** This PR is one of 12 in the ENG-5577 batch. The `@`-import and `GEMINI.md` findings recurred across several of them and are answered by the same measurements above, run once against a converted repo rather than re-asserted per PR.

CI is being re-verified after the latest push. Merge remains a human decision.

nsportsman:
## Adjudication — round 2

My round-1 comment missed a finding, and I want to be explicit about that rather than quiet. This repo has **two** Codex reviewers: the `codex-review` workflow (posting as `github-actions`, which said "No critical issues — LGTM") and the `chatgpt-codex-connector` App, which posted an **inline** review comment on `AGENTS.md:138`. I adjudicated the workflow reviews and overlooked the App's inline comment. It is the only inline comment on any of the 12 PRs in this batch, and it is correct.

### CONFIRMED and FIXED — the maintenance contract named the pointer stub

`AGENTS.md:138` read:

> After adding or changing features, review `README.md` and `CLAUDE.md` for accuracy and update them if needed.

Post-conversion `CLAUDE.md` is 11 bytes (`@AGENTS.md`). The failure is concrete and has two branches, both bad, exactly as the finding states:

- A contributor who obeys it literally edits an import stub, and the canonical `AGENTS.md` goes stale — the instruction quietly stops maintaining the file it is written in.
- A contributor who instead *expands* `CLAUDE.md` back into real content recreates the two-divergent-instruction-files problem this conversion exists to remove.

Fixed in 209b16d:

```
- After adding or changing features, review `README.md` and `AGENTS.md`
  (this file — `CLAUDE.md` is only a one-line pointer to it) for accuracy
  and update them if needed.
```

The pointer's role is stated inline so the next reader does not have to re-derive why `CLAUDE.md` is not listed.

### Scope check — is this instruction elsewhere in the batch?

I did not want to fix the one file a reviewer happened to look at and leave eleven siblings broken, so I swept every converted `AGENTS.md` for self-references:

| Repo | `CLAUDE.md` references in `AGENTS.md` | Disposition |
| --- | --- | --- |
| vespasian | `:1` title, **`:138` maintenance contract** | **fixed here** |
| julius | `:1` title, `:93` `local/CLAUDE.md` | see below — correct as written |
| augustus, constantine, florian, hadrian, pertinax, pius, trajan, Haruspex | `:1` title only | deferred, see below |
| domitian-engine, probus | none | — |

`:138` in vespasian is the **only** imperative contract naming `CLAUDE.md` anywhere in the batch. Nothing to fix on the other eleven.

**julius `:93` is not a false positive of the same kind** and was deliberately left alone: it cites `local/CLAUDE.md`, a *different* file. `local/` is gitignored (`.gitignore:9`) and `git ls-files local/` returns nothing, so that is an untracked per-developer research directory the conversion never touched. Retargeting it at a non-existent `local/AGENTS.md` would break a reference that is currently correct.

### Deliberately deferred — the `# CLAUDE.md` title line

Nine of the twelve `AGENTS.md` files still open with a `# CLAUDE.md` heading and Claude-specific framing ("guidance to Claude Code"). That is genuinely stale for a file now serving Claude, Codex and Gemini, and it is **not** being fixed in this batch, for a reason I'd rather state than leave implicit:

- It is prose inaccuracy, not a broken contract. Nothing acts on the heading; following it cannot corrupt a file or emit a false CI signal. `:138` could, which is why that one is fixed inline and this is not.
- Every reviewer on these 12 PRs verified the same property — that `AGENTS.md` is **byte-identical** to the old `CLAUDE.md`. That byte-exactness is what makes a 12-repo batch reviewable at a glance. Rewriting headers mid-batch trades a cheaply-verifiable move for twelve prose diffs that each need reading.

It is being filed as a follow-up — the non-guard mirror of ENG-5578, which covers exactly this content generalization on the guard side.

---

CI on this PR was green at `3892e43`; the push above re-triggers it and I will confirm it settles. Merge remains a human decision.

AGENTS.md
```diff
@@ -0,0 +1,145 @@
...
+- Test files match source files (`foo.go` → `foo_test.go`)
+- Formatting enforced by `gofmt -s` (run `make fmt`)
+- Linting via `golangci-lint` with gocritic, misspell, revive (run `make lint`)
+- Package-level documentation lives in `doc.go` files
+
+## Development Workflow
+
+- After implementing a feature or fix, run `make check` to ensure all tests and the linter pass.
+- After modifying a Go source file, update its package's `doc.go` if the change affects the package's public API or purpose.
+- After adding or changing features, review `README.md` and `AGENTS.md` (this file — `CLAUDE.md` is only a one-line pointer to it) for accuracy and update them if needed.
```
nsportsman, line 149: @codex Re-review request for round 2. Your round-1 inline finding on this line was confirmed and fixed in 209b16d: the maintenance contract named `CLAUDE.md`, which is now an 11-byte `@AGENTS.md` pointer, so following it literally would have edited the stub and let this file go stale. It now names `AGENTS.md` and explains the pointer's role inline.

Please confirm the fix reads correctly and flag anything else the conversion broke in this file.

nsportsman:
## Adjudication — round 3

One new finding, from the Codex App inline on `AGENTS.md:3`.

### CONFIRMED and FIXED — the canonical file still titled itself `# CLAUDE.md`

The finding is correct, and it reverses a call I made in round 2. I had deferred the heading to a follow-up (ENG-5588) on the argument that these 12 `AGENTS.md` files should stay **byte-identical** to the `CLAUDE.md` they replaced, because that is the property every reviewer on this batch verified and what makes a 12-repo move reviewable at a glance.

That argument no longer holds *for this file*: round 2's fix to `:138` already broke byte-exactness here. So the reason for deferring was spent, and I was still citing it. Deferring a reviewer-raised finding to a ticket is the weaker move when the fix is two lines and the rationale for waiting has lapsed.

Fixed in `7b6425a`:

```diff
-# CLAUDE.md
-
-This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+# AGENTS.md
+
+This is the canonical agent instruction file for this repository. It provides guidance to any coding
+agent working with this code — Claude Code, Codex, and Gemini all load it. `CLAUDE.md` is a one-line
+pointer to this file, and `.gemini/settings.json` names it for Gemini.
```

The replacement states the file's role and names where each harness picks it up, so the next reader does not have to re-derive the topology from three files.

**Checked before changing it** — nothing depends on the heading text. `grep -rn "# CLAUDE.md\|guidance to Claude Code"` across `*.yml *.yaml *.sh *.go *.json` returns nothing. The only non-markdown `CLAUDE.md` reference left in the repo is `.github/workflows/claude-md-drift.yml:1`, a workflow `name:` string — cosmetic, and not a dependency on this file's contents.

### Scope

This repo only. The other 11 in the batch still carry `# CLAUDE.md` titles and remain byte-exact, which keeps them cheap to review; **ENG-5588** covers generalizing them together with the guard-side mirror (ENG-5578) so the wording lands consistently rather than drifting per repo. vespasian is now noted there as already done.

### Evidence the conversion works, since this finding is about what a non-Claude harness sees

Measured live on `capabilities/julius`, each provider against a control that removes the mechanism under test:

- **Codex** — byte-identical content (md5 `351a55ba…`) named `AGENTS.md` → answers a repo-only fact; named `NOTES_LOCAL.md` → `NONE`. Token counts corroborate: the control and an empty directory are identical at 3,120, the `AGENTS.md` arm is 883 higher.
- **Gemini** — `AGENTS.md` + `.gemini/settings.json` → answers; `AGENTS.md` present but **no** settings file → `NONE`. Gemini does not read `AGENTS.md` natively, which is why the opt-in is there.
- **Claude** — verified on constantine #227 with file tools disabled: `@AGENTS.md` loads, the same file with the `@` stripped does not.

---

Round-3 cap reached, so I am not firing another re-review. CI will re-run on `7b6425a`; anything further goes to a follow-up ticket rather than another round here. Merge remains a human decision.

PR #207, opened 2026-08-04
ci(deps): stop dependabot bumping the leaderboard pin (ENG-5687)

nsportsman:
## Problem

The leaderboard reusables are SHA-pinned, and that pinned SHA is allow-listed by `job_workflow_ref` in the **`LeaderboardMetricsRole` IAM trust policy in guard**. So a pin bump is not a version change — it presents a `job_workflow_ref` that is *not* on the trust list, OIDC `AssumeRole` is denied, and the metrics run fails.

The failure is **silent**. `leaderboard-metrics.yml` contains no failure-notification construct, and there is no reconciliation job anywhere, so contribution metrics simply stop being delivered and the loss is only discoverable by auditing run conclusions after the fact.

This has already landed twice as a routine merged bump:

| PR | merged | effect |
| -- | -- | -- |
| `brutus#229` | 2026-07-11 | metrics dark until repaired 07-16 (`brutus#266`) |
| `augustus#241` | 2026-07-11 | metrics dark until reverted 07-26 (`augustus#252`) |

**`vespasian#198` is the same bump, open against this repo right now** (`c351826e` v2.16.3 → `96b68220` v2.16.13). This repo has no ignore rule, so merging it would take vespasian's metrics dark exactly as above.

## Fix

An `ignore:` entry on the `github-actions` ecosystem. Rolling this pin is a coordinated operation — the new SHA must be added to the IAM trust policy and deployed to prod *before* the pin moves (the ENG-4164 runbook). Dependabot cannot do that half, so it must not do this half either.

## Why three entries and not one

- The **two exact names come first, and they are what do the work.** `leaderboard-backfill.yml` is listed because the ENG-5688 fan-out adds that second pinned reusable — identical trust-policy coupling, its own separately-trusted pin `6c715ccf` (v2.17.0) — so the control precedes the dependency rather than following the first broken bump. **Stated so it is not mistaken for a tested control: that entry is a no-op today**, since no backfill caller exists in this repo yet.
- The **`leaderboard-*` wildcard is explicitly NOT load-bearing**, and the file's comment says so in as many words. A dependabot `ignore` list is **match-ANY**: suppression happens if *either* the wildcard or an exact name matches, which makes the wildcard **unobservable** here — no outcome distinguishes "it matched" from "it is inert." Its semantics against a *reusable-workflow* dependency name are unverified in this org and cannot be verified from this PR.

The harm would not be the untested wildcard itself; it would be the instruction a future engineer infers from it. "A reusable added later needs no edit here" would rebuild the exact failure mode this rule exists to stop — a control that looks present, does nothing, and tells no one. So the comment states the opposite in the imperative: **adding a new leaderboard reusable means adding its exact name to this list.**

The exact-name form is also the only one with a positive control behind it, rather than just an absence of PRs: augustus and brutus both still pin **v2.16.3 while v2.16.13 is available**, and both took a `github-actions` dependabot bump on **2026-08-02** for a *different* public-workflows reusable (`claude-md-drift.yml`) while the leaderboard pin was skipped. Same pin, same available upgrade, same ecosystem run — the only difference is the ignore rule.

## Verification

- `dependabot.yml` parses as YAML; the `github-actions` entry has exactly the 3 ignore entries and the `gomod` entry is untouched (`32` insertions, `0` deletions, 1 file).
- **Post-merge check on the exact-name path:** if the rule takes effect, Dependabot should auto-close `vespasian#198` on its next run with an ignore notice. If #198 is still open and unremarked after the next weekly run, then all three entries failed to match and this needs revisiting — please do not close #198 by hand before then.
- **Correction, on the record.** An earlier revision of this body called that a falsifier for the change *as a whole*, and a reviewer on this PR called it "the right way to de-risk" the wildcard. Both were wrong, and the claim is withdrawn. Because the ignore list is match-any, the exact names suppress the bump regardless, so an auto-close confirms only the exact-name path and carries **zero** information about the wildcard — the check could not have failed for the reason I claimed it would pass. Confirmed and fixed in `96656cbf` by demoting the wildcard rather than by trusting the check.

## Not in scope

This is the per-repo stopgap. It is a manually-maintained control that must be re-adopted in every repo and degrades silently, so it is not the durable fix — that is a scheduled reconciler that detects a delivery gap regardless of cause, tracked separately under ENG-5687/ENG-5690.

Refs ENG-5687.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,3 +13,35 @@ updates:
       interval: weekly
     commit-message:
       prefix: "ci"
+    ignore:
+      # The leaderboard reusables are SHA-pinned, and the pinned SHA is
+      # allow-listed by `job_workflow_ref` in the LeaderboardMetricsRole IAM
+      # trust policy in guard. Bumping the pin therefore does not just change a
+      # version — it presents a `job_workflow_ref` that is NOT on the trust
+      # list, so OIDC AssumeRole is denied and the run fails.
+      #
+      # That failure is SILENT in the only way that matters: nothing alerts on
+      # it, so contribution metrics simply stop being delivered and the loss is
+      # only discoverable by auditing run conclusions after the fact. brutus#229
+      # and augustus#241 were both merged as routine dependabot bumps and each
+      # took its repo's metrics dark until someone went looking (ENG-5687).
+      #
+      # Rolling this pin is a COORDINATED operation, not a dependency update:
+      # add the new SHA to the IAM trust policy and deploy it to prod BEFORE
+      # the pin moves (the ENG-4164 runbook). Dependabot cannot do that half,
+      # so it must not do this half either. Roll it by hand as part of that
+      # runbook.
+      #
+      # The wildcard comes first so a leaderboard reusable added later is
+      # covered without a follow-up edit here — an ignore rule whose absence is
+      # invisible is exactly the failure mode above. The two exact names are
+      # then listed as well, and they are NOT redundancy for its own sake: the
+      # wildcard's match semantics against a reusable-workflow dependency name
+      # are unverified in this org, whereas the exact form is empirically proven
+      # (augustus and brutus have had zero leaderboard bumps since adopting it).
+      # If the wildcard silently fails to match, the exact entries still hold —
+      # and a control that cannot be tested before it is needed should not be
+      # the only control.
+      - dependency-name: "praetorian-inc/public-workflows/.github/workflows/leaderboard-*"
+      - dependency-name: "praetorian-inc/public-workflows/.github/workflows/leaderboard-metrics.yml"
+      - dependency-name: "praetorian-inc/public-workflows/.github/workflows/leaderboard-backfill.yml"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -40,16 +40,27 @@ updates:
       # skipped — same pin, same available upgrade, same ecosystem run,
       # differing only in the ignore rule.
       #
-      # The `leaderboard-*` wildcard is OPPORTUNISTIC FORWARD COVERAGE and is
-      # NOT load-bearing. Its match semantics against a reusable-workflow
-      # dependency name are unverified in this org, and — the part that
-      # matters — they CANNOT be verified by watching this rule work:
-      # suppression happens if EITHER the wildcard or an exact name matches, so
-      # a quiet dependabot run only proves the exact names held and says
-      # nothing about the wildcard. Reading it as "a reusable added later needs
-      # no edit here" would rebuild the very failure mode this rule exists to
-      # stop — a control that looks present, does nothing, and tells no one.
+      # The two exact entries below are the load-bearing control. The
+      # `leaderboard-*` wildcard is forward coverage for the window between "a
+      # new coupled reusable lands" and "someone remembers to add its exact
+      # name" — which is exactly how this was learned the expensive way.
+      #
+      # The wildcard's behaviour is NOT observable from here, in either
+      # direction. An ignore list is match-ANY, so a quiet dependabot run proves
+      # only that SOMETHING matched and never which entry; watching this rule
+      # work can therefore never confirm the wildcard fires. Its match semantics
+      # against a reusable-workflow dependency name are unverified in this org.
+      #
+      # So do not read the wildcard EITHER way:
+      #   - NOT as "a reusable added later needs no edit here". If it does not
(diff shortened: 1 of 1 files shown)


nsportsman:
## Adjudication — round 1

Reviewers that actually commented: Codex, Gemini, Claude Review, CodeRabbit. Verdict on each.

### 1. Claude Review — "a version-less `ignore` suppresses all future bumps of these reusables, including any security patch to the workflow itself" → **CONFIRMED as a real consequence; REFUTED as a defect in this PR**

The mechanism is correct and I am not disputing it: `ignore` with no `versions:` constraint suppresses every future bump of these dependencies, security patches included. Worth stating plainly rather than burying, so:

- The dependency is **first-party** — `praetorian-inc/public-workflows` — not a third-party action. A security fix to it is authored by us and rolled by us.
- Rolling the pin was **already** a manual, coordinated operation before this PR, and still is: the new SHA must be added to the `LeaderboardMetricsRole` trust policy and deployed to prod *before* the pin moves (the ENG-4164 runbook). Dependabot structurally cannot do the IAM half, so an automated bump of this specific dependency was never a viable delivery path for a security patch — it would deny OIDC `AssumeRole` and take metrics dark, which is the bug this PR closes.
- So the honest framing is that this PR does not remove a working security-patch channel; it stops a channel that **breaks the workflow whenever it fires** from firing. The security-patch channel for these two reusables is the runbook, and it is unaffected.

No change made. The consequence is now on the record here and is already stated in the config comment.

### 2. Claude Review — "the `leaderboard-*` wildcard is redundant with those exacts by design … the PR ships a concrete post-merge falsifier (auto-close of #198), which is the right way to de-risk it" → **REFUTED, and fixed anyway**

This is the one place I disagree with the review, and the sibling PR palatine#539 got it right where this one did not.

The falsifier is **not** a valid de-risking of the wildcard. Dependabot suppresses a bump if **either** the wildcard or an exact name matches, so #198 auto-closing proves the *exact names* held and says **nothing at all** about the wildcard. My stated test could not fail for the reason I claimed it would pass — it was a check that cannot fail.

That mattered beyond the test, because the old comment told a future engineer that a leaderboard reusable added later is *"covered without a follow-up edit here."* If the wildcard silently does not match a reusable-workflow dependency name, that instruction is false **and its falseness is invisible** — rebuilding the exact failure mode this PR exists to stop.

Fixed in `96656cb`:

- The two **exact names now come first** and are documented as the load-bearing entries; adding a new leaderboard reusable explicitly **means adding its exact name**.
- The wildcard is demoted to **opportunistic forward coverage, explicitly NOT load-bearing**, with the either-match untestability spelled out so the next reader cannot repeat my error.
- Inline annotation above the wildcard entry: *"Non-load-bearing; see above. Never a substitute for an exact entry."*
- Also corrected a weaker claim of my own in the same block: the old text called the exact form *"empirically proven (augustus and brutus have had zero leaderboard bumps since adopting it)"* — that is absence of evidence, not evidence. Replaced with the actual positive control: both repos still pin `v2.16.3` with `v2.16.13` available, and **both took a `github-actions` bump for `claude-md-drift.yml` on 2026-08-02 while the leaderboard pin was skipped.** Same pin, same available upgrade, same ecosystem run, differing only in the ignore rule.

What is ignored today is unchanged — this corrects false documentation, not behaviour.

**Corrected falsifier, superseding the one in the PR body:** #198 auto-closing confirms the **exact-name** path only. The wildcard's match semantics remain unverified and are not relied upon. I am deliberately not manufacturing a test for it: the only way to test a wildcard-only rule is to leave a production repo's metrics exposed to the failure this PR prevents, which is not a trade worth making. The wildcard stays present, free, and explicitly untrusted.

### 3. Codex — no critical issues, no security concerns, no suggestions → **nothing to adjudicate**

### 4. Gemini — no critical issues → **nothing to adjudicate**

### 5. CodeRabbit — **posted no review at all**

Its only comment is a "Review limit reached" warning: *"Your organization has reached its usage spending cap."* That is an org-shared cap, not a verdict on this diff — it read `.github/dependabot.yml` ("Files selected for processing") and then produced nothing. Recording it explicitly, because a rate-limit notice renders in the timeline much like a clean review and treating it as one would let this gate pass vacuously. I will retry after the cap window; if it stays capped I will substitute my own adversarial pass rather than count silence as approval.

### Scope note

The same fix was applied to all four repos in this set — vespasian#207, palatine#539, praetorian-inc/guard#7362, praetorian-inc/umber#109 — not just where it was caught. The defect was in text I wrote four times, not in one file.

CI re-verified after the new commit.

PR #214, opened 2026-08-18
docs: add CHANGELOG.md and changelog contributor guide (LAB-5326)

eli-wald:
## What

Adds a hand-maintained `CHANGELOG.md` at the repo root and documents the changelog workflow in `CONTRIBUTING.md`. Closes LAB-5326 (Pillar 7 — community governance).

## Format decision

**Keep a Changelog + SemVer, maintained by hand, complementary to goreleaser's generated release notes** (the "both" option from the ticket):

- `CHANGELOG.md` is the curated, human-readable history and the **canonical record of breaking changes**. Contributors add entries under `[Unreleased]` per-PR; a maintainer renames it to a version at release time.
- goreleaser's `changelog:` block (`.goreleaser.yml`) keeps generating per-release GitHub Release notes from conventional commits — the commit log for the release, minus `docs:`/`ci:`/`deps:` commits (per `changelog.filters.exclude`). The two are complementary, not redundant.

## Backfill

Only `v1.0.0` (2026-04-03) is tagged, so the file is:

- **`[1.0.0]`** — initial public release: two-stage crawl/import → generate pipeline; OpenAPI 3.0 / GraphQL SDL / WSDL generators; Burp/HAR/mitmproxy importers; SSRF protection.
- **`[Unreleased]`** — the substantial work landed on `main` since v1.0.0, grouped into Breaking changes / Added / Changed / Fixed / Security / Compatibility notes (gRPC, JS/SPA discovery, proxy support, multi-value query params, form/SOAP body parsing, go-rod crawler, determinism, etc.).

## Breaking change recorded

The `capture.json` `query_params` shape change `map[string]string` → `map[string][]string` (LAB-2110) is called out first under `[Unreleased]` → **Breaking changes**, with the migration note (regenerate captures — v1.0.0 files won't round-trip). This break previously lived only in `CLAUDE.md`, which external users never read. Because it changes a persisted on-disk format, the changelog notes the next tagged release is a new major version under SemVer.

## Contributor guidance (`CONTRIBUTING.md`)

- New **Changelog** section + TOC entry: when to add an entry, breaking-changes-first rule, releases cut by a maintainer, and how `CHANGELOG.md` relates to the generated release notes.
- New PR-checklist item: `CHANGELOG.md` updated under `[Unreleased]` if the change is user-facing.

## Acceptance criteria

- [x] `CHANGELOG.md` exists at repo root
- [x] Covers all released versions to date (`v1.0.0` + `[Unreleased]`)
- [x] `capture.json` `query_params` incompatibility recorded as a breaking change
- [x] Release process (hand-maintained vs. generated) documented and consistent with `.goreleaser.yml`
- [x] Contributor guidance documented in `CONTRIBUTING.md` (LAB-2773)

## Verification

`python3 test/check-docs.py --verbose` → `ok (links, presence, roster)` — all relative links/anchors resolve (including the new `#changelog` and `#unreleased` anchors). Docs-only change; the Go CI job does not run (no Go sources touched), which is expected per `CONTRIBUTING.md`.

Refs: LAB-5326, LAB-2110, LAB-2773


CHANGELOG.md
```diff
@@ -0,0 +1,137 @@
...
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+This file is maintained by hand and is the curated, human-readable history: each
+pull request that changes user-facing behavior adds an entry under
+[Unreleased](#unreleased), and at release time a maintainer renames that section
+to the new version. It complements — rather than replaces — the per-release
+GitHub Release notes that goreleaser generates from conventional commits (see
+[`.goreleaser.yml`](.goreleaser.yml)). Those notes list every shipped commit;
```
santiago-praetorian: **AC #4 is not met — this claim is false as written.**

`.goreleaser.yml`'s `changelog:` block filters three commit types out of the generated notes:

```yaml
filters:
  exclude:
    - "^docs:"
    - "^ci:"
    - "^deps:"
```

Measured on `v1.0.0..origin/main`: **59 of 451 commits (13%)** match `^(docs|ci|deps):` and are dropped — including this PR's own `docs:` commit. So the generated notes are not "every shipped commit."

The ticket's AC #4 is *"documented and consistent with `.goreleaser.yml`"*, and the PR description ticks that box, so this is the criterion rather than a stray wording nit.

The same claim appears in **three** places and all three need the edit — `CONTRIBUTING.md:329` ("the complete, raw commit log") and the PR description's *Format decision* section, which is what the squash-merge record preserves.

Suggested: *"the commit log for the release, minus `docs:`/`ci:`/`deps:` commits per `.goreleaser.yml`'s `filters.exclude`."*

CHANGELOG.md
```diff
@@ -0,0 +1,137 @@
...
+breaking changes. See [CONTRIBUTING.md](CONTRIBUTING.md#changelog) for the
+per-PR update workflow.
+
+## [Unreleased]
+
+Work landed on `main` since `v1.0.0` but not yet tagged. Because this includes a
+change to a persisted on-disk format (see Breaking changes), the next tagged
+release is a new major version under SemVer.
+
+### Breaking changes
```
santiago-praetorian, line 26: **A second breaking change is missing from this section.**

Since v1.0.0, `crawl` enforces SSRF protection by default. Verified at the tag:

- v1.0.0's `CrawlCmd` was `URL` + `CrawlOptions` **only** — no SSRF flag existed for `crawl`
- `--dangerous-allow-private` existed solely on `GenerateCmd`/`ScanCmd`, with help scoped to *"probes"*
- `pkg/ssrf` did not exist at v1.0.0

At HEAD, `pkg/crawl/engine.go:170-183` rejects a private seed unless the flag is passed. So `vespasian crawl http://localhost:8080` **worked at v1.0.0 and now hard-errors.**

Calibrating fairly: the error message names its own remedy (*"if crawling a private host … pass `--dangerous-allow-private`"*), so a human recovers on the first run. The real exposure is **scripted/CI callers** crawling internal targets, which start exiting non-zero with no changelog warning.

Also fair: `CHANGELOG.md:93-96` does carry the LAB-2438 `Changed` bullet ("Seed URLs rejected by the crawl frontier now fail loudly instead of silently"). That documents the *error-reporting* improvement, not the new flag requirement, and it's in the wrong section.

`CONTRIBUTING.md:319-323` — added by this PR — routes "changes a default in a way that alters output" to `Breaking changes` with a major-version bump. Worth an entry here so the PR satisfies its own new rule; keep the LAB-2438 bullet, which is a separate fact.

CHANGELOG.md
```diff
@@ -0,0 +1,137 @@
...
+
+- Honor `<base href>` and skip static asset URLs during SPA crawling.
+  ([LAB-2221](https://linear.app/praetorianlabs/issue/LAB-2221))
+
+### Security
+
+- SSRF validation applied to proxied source-map fetches, so reaching private
+  targets still requires `--dangerous-allow-private`.
+  ([LAB-4993](https://linear.app/praetorianlabs/issue/LAB-4993))
+- Importer input hardened against untrusted traffic: host/port smuggling rejected
```
santiago-praetorian: **Imprecise in a way that matters, since this is the Security section.**

To state it precisely — `pkg/importer/doc.go:30` says the **500 MB file cap applies to all formats**, so payload sizes *are* bounded on every path. The claim isn't false; it's imprecise about granularity.

What is **native-mitmproxy-only**:

- 64 MB per-tnetstring-element cap — `doc.go:40`, *"Per-element size (native mitmproxy only)"*
- 500k flow-count cap — `doc.go:32-36`, *"**only** the native mitmproxy path carries an in-parser flow-count cap … Burp XML, HAR, and mitmproxy JSON paths are bounded **only** by the 500 MB file-size cap"*
- host/port smuggling rejection — appears only in `pkg/importer/mitmproxy.go`

So for 3 of the 4 import formats the sole bound is the blanket 500 MB cap. "Bounded across all parse paths" reads as per-element coverage, which could invite over-confidence when importing an untrusted Burp XML / HAR / mitmproxy-JSON file.

Suggested: attribute the per-element and flow-count caps to the native mitmproxy path, and name the 500 MB file cap as what covers the rest.

santiago-praetorian (changes requested):
# PR #214 — curated review

**Recommendation:** REQUEST CHANGES · **Head SHA:** `a4ef95b6ea5367b18f96099983c17039ecacc4c7`
**Ticket:** LAB-5326 · **Scope:** `CHANGELOG.md` (+137), `CONTRIBUTING.md` (+29), 0 deletions

**6 defects worth acting on**, of which 2 will actually break a user's run. Everything else in
`FINDINGS.json` is non-blocking.

---

## Start here: this review disagrees with a box you ticked

The PR description's acceptance list says:

> - [x] Release process (hand-maintained vs. generated) documented and consistent with `.goreleaser.yml`

That criterion **is not met**, and it is the ticket's own AC #4. `.goreleaser.yml`'s `changelog:` block is:

```yaml
changelog:
  sort: asc
  filters:
    exclude:
      - "^docs:"
      - "^ci:"
      - "^deps:"
```

Measured against `v1.0.0..origin/main`: **59 of 451 commits (13%)** match `^(docs|ci|deps):` and are
dropped from the generated release notes — including this PR's own `docs:` commit. So the generated
notes are not "the complete, raw commit log."

The claim is asserted in **three** places, which is why it leads — it is a held belief, not a typo:

| Where | Text |
|---|---|
| `CHANGELOG.md:13` | "Those notes list every shipped commit" |
| `CONTRIBUTING.md:329` | "— the complete, raw commit log." |
| **PR description, "Format decision"** | "— the complete, raw commit log. The two are complementary, not redundant." |

**Fix all three.** The description matters because it is what the squash-merge record preserves.

Related: the description also calls `CHANGELOG.md` **"the canonical record of breaking changes."**
Defect 3 is a real breaking change missing from that section, so that claim does not hold yet either.

---

## Calibration — I am the outlier on this PR

Weigh these on their citations, not on the count:

- The repo's automation **skipped** Claude, Gemini and Codex auto-review as a *"non-code PR"*.
- **CodeRabbit ran and found nothing**: *"No actionable comments"*, *"Merge Risk: ⚪ Minimal"*.
- **Codex (separate connector) filed 3.** All 3 are **still open** — no author reply, none resolved.
- This review: **6 defects**, of which **3 are the same 3 Codex already filed** (marked ⟳ below).
  The **3 marked ★ are new** and no other reviewer caught them.

---

## Tier 1 — will break a user's run (2)

### 1. ⟳ `VESPASIAN_NO_SANDBOX=1` is documented; the code accepts only `"true"`
`CHANGELOG.md:86` · **high** · QUAL-008 · *open Codex thread on `CHANGELOG.md:88`*

`pkg/crawl/browser.go:93` is `opts.NoSandbox || os.Getenv("VESPASIAN_NO_SANDBOX") == "true"` — strict
equality, no truthy coercion. The repo's own CI sets `"true"`
(`.github/workflows/live-tests.yml:162`).

What makes this the worst of the six: it **fails silently in the wrong direction.** Nothing tells the
operator the variable was ignored — the sandbox simply stays enabled and Chrome dies at launch in the
container. Compare defect 3, where the error message names its own remedy.

**Fix:** `VESPASIAN_NO_SANDBOX=true`.

### 2. ⟳ `scan` cannot regenerate a `capture.json`
`CHANGELOG.md:31` · **medium** · REQ-004 · *open Codex thread on `CHANGELOG.md:31`*

The migration note for the one recorded breaking change says: *"Regenerate the capture against the
target with `crawl`, `scan`, or `import`."* `crawl.WriteCapture` has exactly two non-test call sites —
`cmd/vespasian/main.go:442` (`CrawlCmd.Run`) and `:481` (`ImportCmd.Run`). `ScanCmd.Run` (`:704`) ends
by writing the generated **spec**.

So `scan -o capture.json` produces an OpenAPI/WSDL document under a filename that says capture, and
the follow-up `generate` chokes on it. **Silent wrong result**, and it is the migration advice for the
criterion the ticket names by name (AC #3).

**Fix:** name only `crawl` and `import`. If `scan` is worth mentioning, say what it actually does —
skips the capture and regenerates the spec directly.

---

## Tier 2 — factually wrong or incomplete; cheap fixes (4)

### 3. ★ `crawl`'s new default SSRF protection is missing from `Breaking changes`
Belongs at `CHANGELOG.md:24` (`### Breaking changes`); filed against `:81` · **medium** · QUAL-006

**Enforcement verified, not assumed:** `pkg/crawl/engine.go:170-183` rejects a private seed unless
`--dangerous-allow-private` is passed. At the tag, v1.0.0's `CrawlCmd` was `URL` + `CrawlOptions`
**only**, the flag existed solely on `generate`/`scan` scoped to *"probes"*, and `pkg/ssrf` did not
exist. So `vespasian crawl http://localhost:8080` worked at v1.0.0 and now hard-errors.

**Calibrated honestly:** the error message names the remedy verbatim — *"if crawling a private host
(localhost, 127.0.0.1, RFC1918, link-local), pass --dangerous-allow-private"* — so a human recovers on
first run. An earlier draft of this review called it "the single most disruptive change in the
release"; that was overstated. The real harm is narrower: **scripted/CI callers** crawling internal
targets start exiting non-zero with no changelog warning.

It stays in scope because `CONTRIBUTING.md:319-323` — added by **this PR** — mandates a
`Breaking changes` entry for a default change that alters output. The PR fails its own new rule.

Not silent, to be fair: `CHANGELOG.md:93-96` has a LAB-2438 `Changed` bullet ("Seed URLs rejected by
the crawl frontier now fail loudly instead of silently"). That covers the *error-reporting*
improvement, not the flag requirement, and it is in the wrong section.

**Fix:** move under `### Breaking changes` with a one-line migration note; keep the LAB-2438 bullet.

### 4. ★ goreleaser's notes described as complete; the config filters three commit types out
`CHANGELOG.md:13` · `CONTRIBUTING.md:327-329` · PR description · **medium** · QUAL-002, QUAL-014, REQ-002

See the opening section. Its *harm* is low — nobody's run breaks — but it is **false as written** and
it fails the ticket's named AC #4. One clause, three places.

**Fix:** *"the commit log for the release, minus `docs:`/`ci:`/`deps:` commits per `.goreleaser.yml`'s
`filters.exclude`."*

### 5. ★ Importer hardening claim is imprecise, in a Security section
`CHANGELOG.md:108` · **medium** · QUAL-010

Stated precisely, because the first draft of this finding overreached: `pkg/importer/doc.go:30` says
the **500 MB file cap applies to all formats**, so payload sizes *are* bounded on every path. The
claim is therefore **imprecise rather than false**. What is native-mitmproxy-only is the *fine-grained*
bounding — 64 MB per element (`doc.go:40`) and the 500k flow-count cap (`doc.go:32-36`) — plus
host/port smuggling rejection, which appears only in `pkg/importer/mitmproxy.go`.

Worth fixing anyway because it sits in the **Security** section, where "bounded across all parse
paths" invites over-confidence when importing an untrusted Burp XML / HAR / mitmproxy-JSON file.

**Fix:** attribute the per-element and flow-count caps to the native mitmproxy path, and name the
500 MB file cap as what covers the rest.

### 6. ⟳ The documented release procedure leaves the link-reference definitions dangling
`CONTRIBUTING.md:324-326` · **medium** · REQ-007 · *open Codex thread on `CONTRIBUTING.md:325`*

**Verified by running it.** I applied the documented procedure literally for a v2.0.0 cut — renamed
`[Unreleased]`, opened a fresh one, touched nothing else, because nothing else is documented:

- `## [2.0.0]` has **no** reference definition → renders as literal bracketed text, not a link
- `[Unreleased]:` still compares `v1.0.0...HEAD` → the "unreleased" diff shows already-released work
- `make check-docs` exits **0**

No guard can catch it: `test/check-docs.py:50` — *"Reference-style links carry no inline target and
are deliberately not matched."* Certain to fire exactly once, at the next release.

**Fix:** add the two missing steps — add `[<new>]: .../compare/<prev>...v<new>`, and retarget
`[Unreleased]:` to `compare/v<new>...HEAD`.

---

## Withdrawn from this review

**The equivalence / "byte-identical" claims** — `CHANGELOG.md:47-49` and `:90` (QUAL-003, QUAL-009).
Previously rated `medium` and listed as blockers. **Downgraded to `low`, not blocking.**

The mechanism is real and verified (`supersedeConcatMirrors`, `pkg/crawl/jsreplay.go:2521-2533`, drops
an offline-recovered endpoint on a live 404; replay runs at generate time vs scan time). But the
claims are **true under the conditions a reader assumes** — same target, same session. Divergence
requires the target's SPA surface to change between the `crawl` and the `generate`, or a transient
404: an *environmental* precondition, not a default setting. The reachability test therefore yields
`low`, not `medium`.

The original rating held a changelog summary to CLAUDE.md's exhaustive-internal-doc standard.
CLAUDE.md carries that caveat because it documents mechanisms; a changelog legitimately simplifies. A
"you didn't add a caveat" objection with a real code citation attached is still not a defect.

Also ruled out so nobody chases it: the `--fetch-sourcemaps` default asymmetry between `scan` and
`generate` is **moot** — `pipeline.AnalyzeJS`'s idempotency guard
(`internal/pipeline/augment.go:95-99`, `crawl.AnyStaticSource`) skips the static pass once `crawl` has
populated `static:js*`.

---

## Worth fixing, not blocking (6)

| ID | Where | What |
|---|---|---|
| QUAL-005 | `CHANGELOG.md:81` | `--concurrency` (default 10) is new since v1.0.0 and never named; the bullet says "concurrent go-rod engine" without the knob. |
| QUAL-007 | `CHANGELOG.md:86` | `VESPASIAN_ALLOW_BROWSER_DOWNLOAD=true` is the escape hatch for the browser-pinning change in that same bullet, and is omitted. |
| QUAL-004 | `CHANGELOG.md:65` | `--proxy-insecure` reads as covering the gRPC path listed beside it. It applies to http/https proxies only; socks5 always verifies, and gRPC target certs are governed by `--grpc-insecure-skip-verify`. |
| QUAL-012 / REQ-006 | `CONTRIBUTING.md:315` | The policy enumerates six groups; `CHANGELOG.md:111` uses a seventh, `Compatibility notes`, defined nowhere. Add it or fold the section into `Changed`. |
| REQ-001 | `.github/PULL_REQUEST_TEMPLATE.md:19` | The checklist a contributor actually sees has no changelog item, so the new `CONTRIBUTING.md:357` item is invisible at PR time. |
| REQ-008 | `GOVERNANCE.md:47` | "## Releases" says *"There is no manual publishing step"* and never mentions the changelog rename this PR makes mandatory. |

Plus **TEST-006** (`test/check-docs.py:35`): `CHANGELOG.md` is not in `REQUIRED_ROOT_FILES`. **Proven
by mutation:** delete `CHANGELOG.md` and its one inbound link and `make check-docs` still exits 0 —
the new required document's only guard is a single prose link this same PR adds.

## Before merge, not a defect

**REQ-003** — `[Unreleased]` omits LAB-4678's `--max-requests` and `--interact`. **Not an authoring
error:** `git merge-base --is-ancestor` confirms LAB-4678 landed on `main` on 2026-08-19, a day
*after* this PR's head (`a4ef95b`, 2026-08-18). Main moved under the branch. Rebase and add the
entries, then re-check `git log HEAD..origin/main`.

## File separately — not this PR's to fix

Five pre-existing bugs in `test/check-docs.py`, surfaced only because this PR leans on that harness
for its verification. They would exist if this PR had never been written:

- **`:238` roster check passes vacuously** — `check_roster` returns early when either input is `None`,
  and both helpers return `None` *without appending a failure* when their file is missing (`:199`,
  `:221`), unlike their three sibling `None` paths (`:215`, `:226`, `:230`) which all append first. A
  repo with no `GOVERNANCE.md` prints `ok (roster)` having asserted nothing. Two-line fix, and the
  sharpest of the five.
- **`:141` presence accepts a directory** — `(root / name).exists()` is true for `mkdir SUPPORT.md`.
- **`:194` no floor on links examined** — "zero links checked" is indistinguishable from "all valid".
- **`:119` `anchors_for` swallows read errors** — caches an empty anchor set, so an unreadable target
  is reported as a broken anchor in the file pointing at it.
- **`:258` no self-test**, unlike its siblings `validate_test.sh` (69 KB) and
  `setup-live-targets_test.sh` (23 KB).

---

## Honest limits of this run

Full record: `FINDINGS.json` (34 findings, 8 blocker IDs → **6 distinct defects**) and
`PR-REVIEW-COMMENT.md`.

- **The raw count overstates the problem.** 34 findings resolve to 6 defects worth acting on. The
  goreleaser belief alone accounts for 4 findings — two lanes filed the same `CHANGELOG.md:13` line,
  plus the `CONTRIBUTING.md` copy, plus a separate "it's duplicated" entry. The dedup signature
  includes the reporting lane by design, so one defect found twice survives as two entries.
- **3 findings should not have shipped**: QUAL-001, QUAL-011 and TEST-005 each argue themselves down
  to "no action required" in their own text.
- **Two corrections applied during review**, both recorded in `FINDINGS.json`: `reachability_changed`
  was initially stamped per *file* and wrongly inherited by every `test/check-docs.py` finding (now
  per finding, five set to `may_block=false`); and QUAL-003/QUAL-009 were withdrawn from `medium` as
  described above. Neither changed the verdict.
- **No security lane ran.** Neither `SEC-BE` nor `SEC-FE` triggers on a two-markdown diff — correct
  per the dispatch matrix, but it means defect 5, a claim in a *Security* section, was assessed by the
  quality lane rather than a security reviewer.
- **This audit is not complete.** The quality lane surfaced new medium findings on all four of its
  passes (3, 1, 2, 1) and never went dry, so it hit the pass cap unconverged and no watermark was
  written. A re-review re-audits the whole PR and may find more.

_No cap applied — this is a reading order, not a quota. Nothing was posted to the PR._

santiago-praetorian:
# PR #214 — review

**Recommendation:** REQUEST CHANGES · **Head:** `317b278` · **Ticket:** LAB-5326
**Scope:** `CHANGELOG.md` (+148, new), `CONTRIBUTING.md` (+37), 0 deletions

**All six findings from the previous round are fixed.** I re-verified each against the code rather than taking the commit message for it; none is re-reported. `VESPASIAN_NO_SANDBOX=true` now matches `browser.go:93`, the `scan` migration advice is gone, the crawl SSRF breaking change is recorded, the goreleaser wording is corrected in all three places including the PR description, the importer hardening is attributed correctly, and the release procedure now carries the link-reference steps.

**Acceptance criteria: 4/4 met.** `v1.0.0` is the only tag and the only release, and `## [1.0.0] - 2026-04-03` matches its date. The `query_params` break leads `### Breaking changes`. The goreleaser description now matches `.goreleaser.yml:28-34` exactly. Zero unmet criteria — last round had three.

**So the blockers below are not unmet requirements. They are factual defects in the prose the requirements asked for.**

---

## The pattern

This is a changelog, so every bullet is a claim about the code — and nine of them don't survive being checked against it. Two are in text the round-1 fix commit itself introduced, which is why they read as new rather than as regressions.

Worth noting what didn't catch any of this: `docs-check` validates links and anchors only; `ci.yml`, `security.yml` and `verify-pins.yml` skip on path filters; and Claude, Gemini and Codex all posted *"non-code PR"* and skipped. A file making two dozen verifiable assertions about Go behaviour is precisely the file that most needs a content-aware reviewer, and it's the one the path filter excludes.

---

## Will mislead a user (3)

### `crawl` on a host with no system Chrome now hard-fails, and the entry names the wrong remedy
`CHANGELOG.md:95-97`

The bullet says the crawler *"pins the system Chrome (no auto-download) and disables browser telemetry; set `VESPASIAN_NO_SANDBOX=true` for containerized runs."* The pinning claim is true; the behaviour change it omits is the one that breaks people.

At v1.0.0, `NewBrowserManager` left the launcher binary unset unless `--chrome-path` was given, so go-rod downloaded a managed Chromium and the crawl succeeded. At head, `pkg/crawl/browser.go:172` returns instead:

```
no system Chrome/Chromium found in standard paths: set BrowserOptions.ChromePath,
install a browser, or set VESPASIAN_ALLOW_BROWSER_DOWNLOAD=true to allow go-rod to download one
```

`--headless` defaults to `true` and `AllowBrowserDownload` to `false`, so `crawl` and `scan` now exit non-zero on any host without a system browser — the code comment at `browser.go:154-157` names arm64 Linux specifically. `VESPASIAN_ALLOW_BROWSER_DOWNLOAD` appears **zero times** in `CHANGELOG.md`; it lives only in `README.md:248`. The one actionable string this entry offers, `VESPASIAN_NO_SANDBOX=true`, controls the Chrome sandbox (`browser.go:93`) and has no effect on binary resolution. A CI operator following this entry literally sets it and still fails.

**Fix:** state the failure and the real migration — install a browser, pass `--headless=false`, or set `VESPASIAN_ALLOW_BROWSER_DOWNLOAD=true`. By this PR's own new rule at `CONTRIBUTING.md:319-323` (a changed default that alters output), it belongs under `Breaking changes`.

### The `[1.0.0]` SSRF entry contradicts the `[Unreleased]` breaking change 110 lines above it
`CHANGELOG.md:144-145`

`:144` states unqualified that at v1.0.0 *"private targets require an explicit `--dangerous-allow-private`"*. `:35-37` states that at v1.0.0 a crawl against a private host **succeeded**. Both cannot be true.

`git show v1.0.0:cmd/vespasian/main.go` settles it: `--dangerous-allow-private` is declared only at `:479` (`GenerateCmd`) and `:552` (`ScanCmd`), with help text *"Disable SSRF protection for probes"*. `CrawlCmd` at `:386` has no such field, and `pkg/ssrf` did not exist at the tag — v1.0.0's SSRF protection lived entirely in `pkg/probe`. The breaking-change bullet is right; the `[1.0.0]` bullet is over-broad.

This matters because a reader who trusts `:144` concludes the crawl gate is pre-existing and their private-target scripts are unaffected — which is the exact migration the breaking-change entry exists to surface.

**Fix:** scope the `[1.0.0]` bullet to the probe path.

### The host/port-smuggling fix is scoped to the format that didn't exist at v1.0.0
`CHANGELOG.md:118-120`

The Security bullet attributes the smuggling rejection to the new native tnetstring path, alongside the 64 MB and 500k caps. Those two caps are genuinely tnetstring-only. The smuggling rejection is not: `validateHost` has exactly one call site, `pkg/importer/mitmproxy.go:578`, inside the **shared** `parseFlow` (three call sites), and its own doc comment at `:491-493` says *"Applied at the importer boundary (both JSON and native paths)"*.

The mitmproxy JSON path is the importer that shipped in v1.0.0, and it had no such guard. So the one security fix in this release that actually affects released code is described as belonging to a format that didn't exist then, and a v1.0.0 JSON-importer user reads "not my format" and stays put.

This text was rewritten by the round-1 fix commit (the imprecise-hardening finding). That fix made the file-size attribution precise and introduced this narrower mis-attribution.

**Fix:** split the claims — keep the 64 MB / 500k caps on the native path, and give host-field validation its own sentence naming both mitmproxy paths.

---

## Wrong or materially incomplete (6)

### "disables browser telemetry" is also disabling two Chrome security controls
`CHANGELOG.md:96`

`disableChromeTelemetry` (`pkg/crawl/browser.go:279-284`) sets `disable-component-update` and appends `SafeBrowsingHashPrefixRealTimeLookups` to `disable-features`, on every crawl. The code documents both trade-offs — `browser.go:187-191` notes Chrome falls back to the build-time CRLSet for the crawl's duration, and `:245-256` records the Safe Browsing loss as a reviewed, accepted posture. The changelog says none of it, and frames the change as *"egress hardened"*, which reads as a pure security win. This belongs in the `### Security` section, stated as the two-way trade it is.

### OpenAPI component names changed since v1.0.0, and it's in no section
`CHANGELOG.md:99`

`git diff v1.0.0 origin/main -- test/rest-api/expected-spec.yaml` shows `Email-addressResponse` → `EmailAddressResponse` with the matching `$ref` flip, for a target path that is byte-identical in `test/rest-api/main.go` across both revisions. Downstream OpenAPI codegen emits different type names — silently, no warning. Nothing in the changelog signals it.

### `[Unreleased]` says it covers `main`; the branch is 24 commits behind
`CHANGELOG.md:21`

`git rev-list --count HEAD..origin/main` = 24 (merge-base `c0a41ec`, 2026-08-18). Diffing kong `name:` tags between `origin/main` and this head returns exactly `interact` and `max-requests` as present on main and absent here — two shipped CLI flags in neither the changelog nor the section that claims to cover them. A rebase plus two bullets closes this; alternatively soften the preamble's scope claim.

### "preserved end-to-end" vs. two caps that drop data silently
`CHANGELOG.md:67-69`

`pkg/crawl/capture.go:31` is `MaxQueryParamValues = 256` and `:38` is `MaxQueryParamKeys = 512` — unconditional consts, no flag. The function's own doc at `:44-46` says *"Excess values are also dropped silently."* The changelog asserts unbounded preservation.

### The breaking-change rule omits exported Go API changes
`CONTRIBUTING.md:319-323`

The rule lists four triggers — invalidates a `capture.json`, changes a persisted format, removes a flag, changes an output-altering default — all CLI or on-disk. None covers a source break in an exported symbol, though the module is importable and `pkg/sdk` is advertised for embedding.

The gap is already realized in this PR: `pkg/crawl/types.go:22` went from `map[string]string` to `map[string][]string`, a compile break for every consumer of `pkg/crawl`, and the entry offers only *"Regenerate the capture"* — which does nothing for a downstream Go build.

### The link-reference definitions this PR adds are unverifiable by `make check-docs`
`CHANGELOG.md:147-148`

This PR introduces the repo's first reference-style link definitions — `origin/main` has zero. `test/check-docs.py`'s `LINK_RE` structurally cannot see them; the comment above it says so: *"Reference-style links carry no inline target and are deliberately not matched."*

Run against a clean exit-0 baseline in an isolated clone, each mutation's application confirmed by file hash:

| Mutation | Result |
|---|---|
| delete `CHANGELOG.md:148` (`[1.0.0]` definition) | **exit 0 — undetected** |
| delete `CHANGELOG.md:147` (`[Unreleased]` definition) | **exit 0 — undetected** |
| `## [Unreleased]` → `## Unreleased` | **exit 0 — undetected** |
| control: break the `.goreleaser.yml` link | exit 1 — caught |
| control: rename `## Changelog` | exit 1 — caught |

The controls matter: the checker works, it just can't see this construct. The round-1 fix documented the gap in prose; ~15 lines would close it, and the release procedure now depends on two hand-edits nothing verifies.

---

## Not blocking, but worth a look

`README.md` links every root-level document — `CONTRIBUTING.md`, `CODE_OF_CONDUCT.md`, `GOVERNANCE.md`, `SUPPORT.md`, `SECURITY.md`, `LICENSE` — except this one. `git grep -l CHANGELOG` outside the two files in this PR returns zero. GitHub gives the others dedicated UI slots and gives `CHANGELOG.md` none, so discoverability rests on that missing line. Not gating the merge, since the fix is in a file this PR doesn't touch.

Also non-blocking: `.goreleaser.yml`'s excludes are anchored `^docs:`, so they miss *scoped* commits — 67 of 96 `docs`/`ci`/`deps` commits since v1.0.0 are filtered, the other 29 (`ci(security): …`, `docs(cmd): …`) are not, while `CONTRIBUTING.md:296` encourages scopes. This PR's prose is accurate about what the config says, so it isn't this PR's defect, but `^docs(\(.*\))?:` would make the config mean what everyone reads it to mean. Worth a follow-up.

Fifteen further `low`/`nit` observations aren't listed here. Several are of the form "flag X isn't mentioned" — factually true, but `CHANGELOG.md:15` says the file *"summarizes what matters to users"*, so those are judgement calls you can reasonably decline rather than defects.

---

<sub>Two lanes ran: quality (12/12 rubric items, 4 passes) and tests (9/9, 3 passes). The quality lane hit its pass cap still producing findings, so this audit is not exhaustive — expect another round to surface more. Security lanes didn't run: no server-side or frontend files.</sub>

test/check-docs.py
```diff
@@ -47,11 +52,31 @@
...
 LINK_RE = re.compile(r"!?\[(?P<text>[^\]]*)\]\((?P<target>[^)\s]+)(?:\s+\"[^\"]*\")?\)")
 HEADING_RE = re.compile(r"^(?P<hashes>#{1,6})\s+(?P<text>.+?)\s*$", re.MULTILINE)
 FENCE_RE = re.compile(r"^([ \t]*)(```|~~~)", re.MULTILINE)
 
+# Reference-style link plumbing (Keep a Changelog leans on it: a "## [1.0.0]"
+# heading is a shortcut reference that resolves through a "[1.0.0]: <url>"
+# definition at the foot of the file). LINK_RE sees neither half, so without a
+# dedicated check a deleted definition (heading renders as literal brackets) or
+# an orphaned definition passes silently.
+REF_DEF_RE = re.compile(r"^ {0,3}\[(?P<label>[^\]\n]+)\]:\s+\S", re.MULTILINE)
```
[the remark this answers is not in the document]
eli-wald: **CONFIRMED — fixed in a9fc8b9.**

CommonMark makes the whitespace after a link-reference-definition colon optional, so `[release]:https://example.test` is a valid definition. `REF_DEF_RE` has been relaxed from `\]:\s+\S` to `\]:\s*\S` — colon-adjacent destinations are now collected, while the trailing `\S` still rejects a bare empty destination (`[empty]:` at end of line, which requires `<>` to be valid).

Added a unit test `test_definition_colon_adjacent_target_and_empty_rejected` pinning both boundaries (colon-adjacent collected; empty rejected). Full suite: 25 tests pass; `make check-docs` green.

CHANGELOG.md
```diff
@@ -0,0 +1,137 @@
...
+### Changed
+
+- **Headless crawler replaced.** The Katana-based crawler was replaced by a
+  concurrent go-rod engine, then generalized behind a `Crawler` interface with
+  go-rod (headless) and `net/http` backends; Katana was removed entirely.
+  ([LAB-2785](https://linear.app/praetorianlabs/issue/LAB-2785),
+  [LAB-2786](https://linear.app/praetorianlabs/issue/LAB-2786))
+- **Crawler egress hardened.** The crawler now pins the system Chrome (no
+  auto-download) and disables browser telemetry; set `VESPASIAN_NO_SANDBOX=1` for
+  containerized runs.
```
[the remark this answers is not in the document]
eli-wald: **CONFIRMED as a real concern — already addressed.**

`pkg/crawl/browser.go:93` gates on `os.Getenv("VESPASIAN_NO_SANDBOX") == "true"`, so only the literal `true` takes effect. The current `CHANGELOG.md` no longer references the variable at all (the flagged line was superseded in b2004bd), and where a value is shown — `test/README.md:62` and `:140` — it correctly uses `VESPASIAN_NO_SANDBOX=true`. No further change needed.

CHANGELOG.md
```diff
@@ -0,0 +1,137 @@
...
+release is a new major version under SemVer.
+
+### Breaking changes
+
+- **`capture.json` `query_params` changed shape.** To support repeated query
+  parameters (e.g. `?tag=a&tag=b`), the `query_params` field of each captured
+  request changed from `map[string]string` to `map[string][]string`. Capture
+  files produced by `v1.0.0` use the old single-value shape and are **not
+  compatible** — they will not round-trip through `generate`. Regenerate the
+  capture against the target with `crawl`, `scan`, or `import`.
```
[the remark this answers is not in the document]
eli-wald: **CONFIRMED — already addressed in b2004bd.**

`CHANGELOG.md` now reads: "Regenerate the capture against the target with `crawl` or `import` (`scan` writes a generated spec, not a `capture.json`)." Verified against source: `crawl.WriteCapture` is invoked only from the crawl and import paths (`cmd/vespasian/main.go:488` and `:527`), never from `scan`. No further change needed.

CHANGELOG.md
eli-wald: **CONFIRMED — already addressed in b2004bd.**

Both `CHANGELOG.md` (lines 13–15) and `CONTRIBUTING.md` (lines 356–358) now state the goreleaser notes are "the commit log for the release, minus `docs:`/`ci:`/`deps:` commits (per that file's `changelog.filters.exclude`)", matching `.goreleaser.yml`'s `changelog.filters.exclude`. The overstated "every shipped commit" phrasing is gone. No further change needed.

CHANGELOG.md
eli-wald: **CONFIRMED — already addressed in b2004bd.**

The Security section (`CHANGELOG.md:150–156`) now distinguishes the shared 500 MB file-size cap (bounding every format: Burp XML, HAR, and both mitmproxy paths) from the native-mitmproxy-tnetstring-only 64 MB per-element and 500k flow-count caps — matching `pkg/importer/doc.go`. No further change needed.

santiago-praetorian (approved):

**PR:** https://github.com/praetorian-inc/vespasian/pull/214
**Linear:** https://linear.app/praetorianlabs/issue/LAB-5326/add-changelogmd
**Head SHA:** `a9fc8b9cd9a6b0106ac460cc1dcfdc49974d303b`

**Recommendation:** APPROVE

All four LAB-5326 acceptance criteria are met:

- `CHANGELOG.md` exists at repo root
- Covers all released versions (`v1.0.0` + `[Unreleased]`)
- `capture.json` `query_params` incompatibility recorded as a breaking change
- Release process (hand-maintained vs goreleaser) documented and consistent with `.goreleaser.yml`

Previous-round findings are already addressed. No blockers, no critical/high.

|---|---|---|---|---|---|
| Count | 0 | 0 | 0 | 0 | 1 leftover, non-blocking |

<details>

<details>

<details>
<summary><code>QUAL-001</code> · nit · <code>AGENTS.md:223</code></summary>

**Issue:** `docs-check` now runs `python3 test/check-docs_test.py` before the checker, but AGENTS.md still says the job runs only `python3 test/check-docs.py --verbose`.

**Recommendation:** Name both commands in the `docs-check` paragraph. Optional follow-up; not blocking.

</details>

</details>

</details>

`DOC-001` (CLAUDE.md staleness) was pruned: `CLAUDE.md` is a one-line `@AGENTS.md` pointer and AGENTS.md is already in this PR.

_Review generated by `capability-pr-review`, then pruned before posting._

PR #208, opened 2026-08-07
test(live-tests): close 6 assertion-coverage gaps from PR #187 review (LAB-5611)

eli-wald:
## Summary

Follow-up to the merged **LAB-3890** PR #187 (`test/run-live-tests.sh` + friends). The `/capability-pr-review` of #187 raised **7 blocking** assertion-coverage findings; per the author decision all 7 are tracked in **[LAB-5611](https://linear.app/praetorianlabs/issue/LAB-5611/vespasian-pr-187-follow-up-close-live-test-assertion-gaps-in)**. This PR closes the **6 fixable** ones — each with a committed regression case in the live-test harness — and closes **QUAL-002** as won't-do.

Test harness + JSON fixtures only; **no production Go changed**.

Fixes LAB-5611.

### Fixed, with committed regression cases

**TEST-004** — `test/run-live-tests.sh` · `test_rest_api` printed `endpoint_count`/`expected_count` in the PASS row without comparing them (so `rest-api` could emit `PASS 12 11`), while the twin `test_scan_rest` already compared under a comment citing this exact defect. Added the same hard-fail on mismatch, via the shared `assert_exact_path_count` helper (`test/validate.sh`) that numeric-guards both operands so a `?`/empty count read hard-fails instead of string-comparing. Every exact-count call site — `test_rest_api`, `test_scan_rest`, `test_concat_spa`, `test_forms_target`, `test_generate_js_static` — now routes through that helper. *(Live-only assertion; mirrors the already-green `test_scan_rest` count-compare, exercised by the gated `test` job.)*

**TEST-005** — `test/rest-api/expected-paths.json` + `test/rest-api/scan-expected-paths.json` + `test/validate.sh` + `test/validate_test.sh` · `total_paths` was a hand-maintained duplicate of `len(paths)` with nothing diffing the two fixtures mechanically. `validate_path_coverage` now fails when `total_paths != len(paths)` (hardens every fixture); `validate_test.sh` adds a cross-file fixture-parity check over both rest-api fixtures, placed deliberately in the **un-gated** `validator-regression` job so `skip-live-tests` cannot switch it off (it briefly lived in `test-runner-args.sh`, whose drift-guard runs only in the gated `test` job). Regression in `validate_test.sh`: match accepted, mismatch rejected (`PARITY MISMATCH`), missing-key tolerated.

[...]

PR #215, opened 2026-08-19
docs(test): rebase live-test README onto current main (LAB-3672)

logan-bayless:
Supersedes #174.

## What

Rebases the #174 live-test README work onto current `main`, re-deriving every count against the current tree, and corrects eleven claims in the live-test docs that contradicted the suite.

## Notes / limitations

Docs and test-fixture `description` fields only. No product code, no change to what the suite asserts.

**Rebase.** #174 forked 169 commits back and its headline numbers were stale: `scan-rest`, `forms-target`, `concat-spa-two-stage`, `auth-capture`, `ssrf-rejection` and `no-download` all landed after its fork point. Counts are re-derived from `OFFLINE_TARGETS` (21) + `LIVE_TARGETS` (11) = 32, not copied. This carries over the parts of #174 that `main` still lacks (concat-spa ports/config/help entries, the `query_params` note, the Directory Structure additions) and closes three gaps predating #174: the missing Output directories, the drifted `--sweep` help text, and the undocumented validator install that `test/spec-validators` needs and `setup-live-targets.sh` does not run.

**Corrections from review.** Three, all in `6adf103`:

- The REST-counts blockquote said the deliberate two-stage-vs-scan method divergence on `/api/login` and `/api/upload` was "asserted around rather than enforced". `assert_path_methods` enforces it, unconditionally, in both live REST targets. The premise the sentence was built on is true and stays. The same stale sentence sat in both `rest-api` fixture `description` fields, kept in lockstep by a note telling you to update both, so all three are corrected together.
- The sample `.live-test-config` showed six `TARGETS_SETUP` entries where the setup script writes seven, since it expands `concat-spa` into `concat-spa,concat-spa-two-stage` for the run-list. The sentence describing that sample had to change with it: it called the value "every member of the setup script's `ALL_TARGETS`", and `concat-spa-two-stage` deliberately is not one.
- Step 6's list of the extra `rest-api`/`scan-rest` assertions omitted `assert_path_methods`.

**Corrections from self-review.** Eight more inaccuracies this PR introduced, found reviewing the branch against current `main` and fixed across `4b2376b`, `5530eb0`, `f4a0c00`, `7e548e1` and `533fc3a`. Most share the shape of the first review finding, a true premise with a false conclusion drawn from it:

- `crawl` was said to recover nothing for the concat targets and the concat-derived paths never to enter the capture. `--analyze-js` defaults on, so `crawl` reconstructs them statically and they land in the capture tagged `static:js-concat`; what `crawl` omits is the probe. The file restates this split in ten places and several were individually wrong, including one crediting JS-replay with recovering the paths when it drops the 404 control.
- The GET-only classification of `/api/login` and `/api/upload` was attributed to the two-stage shape plus `--probe=false`. `crawl` defaults to `--headless=true` and does run JS; the cause is the runner's `--headless=false` leg.
- The `protoc` prerequisite quoted a skip message that exists nowhere in the repo and contradicted the "No `protoc` required." note below it.
- The Node.js bullet claimed the spec validators are what every OpenAPI and GraphQL target validates through; the WSDL and `.proto` producers have no Node validator at all.
- "Spec generation always uses the net/http capture" is false for `soap-service`, which generates from a synthetic capture.
- The deviation sentence sent readers to a `scan-rest` section that does not exist, omitted `soap-service`, and called the two-stage shape the one "most targets use".
- The three `npm ci` instructions omitted `--ignore-scripts`, which every executable call site uses.
- The `grpc-server` results tree omitted `proto-validate.err`.

**Verification**

- `test/validate_test.sh`: 95 passed, 0 failed, including all six `assert_path_methods` cases and the `rest-api` fixture-parity cases that read both edited fixtures. Needs `(cd test/spec-validators && npm ci --ignore-scripts)` first.
- `test/check-docs.py` passes.
- Both fixtures still parse; only `description` changed, with key order and formatting preserved.
- The sample `TARGETS_SETUP` byte-matches the run-list the setup script computes from `ALL_TARGETS`.

Original work by @santiago-praetorian in #174 — leaving LAB-3672 assigned to him.


santiago-praetorian (changes requested):
Requesting changes on one item. The other two are cheap fixes in the same neighbourhood, so they're worth folding into the same push rather than a follow-up.

First, the part that isn't a finding: this rebase holds up well under checking. All 31 result directories in the Output tree match `run-live-tests.sh`'s actual writes in both directions; 86 of 86 Directory Structure entries exist except the two marked gitignored; both embedded `--help` blocks are byte-identical to the scripts' real output; all 31 rows of the sample table match either their fixture or the hardcoded expected value in their `set_test_result` call; and the `query_params` note plus its two cited values are verbatim-correct against `expected-har-duplicates-capture.json`. The three verification claims in the PR body all check out.

---

### 1. `test/README.md:394` — the closing clause is the reverse of what the suite does

The REST-counts blockquote ends:

> `_assert_fixture_parity` in `test/validate_test.sh` pins ... but not on the method lists, **so this divergence is asserted around rather than enforced.**

The first half is correct — `_assert_fixture_parity` really does compare only the path set, `total_paths`, `post_form_paths` and `post_form_body_fields_by_path`. The conclusion drawn from it is not. `assert_path_methods` (`test/form-spec-asserts.sh:400`) enforces exactly that method divergence, and it's wired into both live targets: `run-live-tests.sh:447` for `rest-api` and `:571` for `scan-rest`.

Its own header comment names the gap it closed in #208 — *"So the divergence was inert: a regression that flipped either path's classification would not fail a test."*

I ran `test/validate_test.sh` offline to confirm it isn't decorative:

```
PASS (rejected invalid): assert_path_methods rejects a spec missing a declared POST
PASS (rejected invalid): assert_path_methods rejects an unexpected POST on a GET-only decl
```

The cost of leaving it is a debugging cycle: someone changing generator classification reads this sentence, concludes the fixture method lists are decorative, and gets failed by an assertion the document told them doesn't exist.

One thing to catch while fixing it — `test/rest-api/expected-paths.json`'s own `description` field carries the same stale sentence ("...asserted-around rather than enforced"). It predates this PR, but fixing only the README leaves the contradiction sitting in the repo.

### 2. `test/README.md:223` — the sample `.live-test-config` omits `concat-spa-two-stage`

`setup-live-targets.sh:912` computes `run_targets="${targets/concat-spa/concat-spa,concat-spa-two-stage}"` and passes `run_targets` — not `targets` — to `write_config`, so the generated file has seven entries where the sample shows six:

```diff
-TARGETS_SETUP=rest-api,soap-service,graphql-server,grpc-server,concat-spa,forms-target
+TARGETS_SETUP=rest-api,soap-service,graphql-server,grpc-server,concat-spa,concat-spa-two-stage,forms-target
```

No coverage is lost either way, since `concat-spa-two-stage` is already in `LIVE_TARGETS` and a bare run exercises it regardless. This is CodeRabbit's open comment on the PR, and it's right.

### 3. `test/README.md:53` — worth bundling with #1

Step 6 enumerates what `rest-api` and `scan-rest` additionally assert — `assert_post_get_operations`, `assert_form_body_fields` — and omits `assert_path_methods`. That's the same defect as #1 in its other half. Not worth a round on its own, but fixing 394 without 53 leaves the document self-consistent in the wrong direction.

---

Everything else I found was either optional wording or predates this PR, and none of it needs to hold up the merge. Happy to post the rest separately if it's useful.

logan-bayless:
All three addressed in 6adf103.

**1. The "asserted around rather than enforced" clause.** Corrected. The first half stands and stayed: `_assert_fixture_parity` does compare only the path set, `total_paths`, `post_form_paths` and `post_form_body_fields_by_path`. Only the conclusion is replaced, naming `assert_path_methods` as the enforcing assertion. Cited by file and function rather than line, since the `:447`/`:571` in your review had already moved under the rebase.

The sentence was in a third place as well: `test/rest-api/scan-expected-paths.json`'s `description`, the lockstep sibling of the fixture you flagged. Its own note instructs keeping the claim matching "in both files", so all three are corrected rather than two.

**2. Sample `.live-test-config`.** `concat-spa-two-stage` added. The one-line diff needed a second edit to be correct: the sentence immediately after the sample describes the value as "every member of the setup script's `ALL_TARGETS`", and `ALL_TARGETS` (`setup-live-targets.sh:55`) is exactly the six — `concat-spa-two-stage` is deliberately not a member, per the comment at `:1129`. That sentence now describes run-list semantics instead. Without it the fix would have traded one false statement for another.

**3. Step 6.** `assert_path_methods` added, placed after the `forms-target` cross-reference rather than inside the list, because that clause covers only the post-operation and body-field checks the `forms-target` section documents.

---

Reviewing the branch against current `main` then turned up eight further inaccuracies that this PR introduces, fixed across 4b2376b, 5530eb0, f4a0c00, 7e548e1 and 533fc3a. Most are the same shape as your finding 1 — a true premise with a false conclusion drawn from it.

- **`crawl` and the concat targets.** The file said a plain `crawl` recovers nothing here and that the concat-derived paths never enter the capture. `crawl` does not run JS-replay, but `CrawlCmd.AnalyzeJS` defaults true (`cmd/vespasian/main.go:269`) and `crawl` calls `pipeline.AnalyzeJS` at `:485`, reaching `crawl.ExtractStaticConcatPaths` via `jsstatic` (`extractor.go:687`), so the paths are reconstructed statically and land in the capture tagged `static:js-concat`. The idempotency guard at `main.go:480` exists because `crawl` already analysed those bundles. This file restates that split in ten places — prose, two results-tree annotations, a target-table row, a troubleshooting heading and body — and several were individually wrong, including one annotation crediting JS-replay with *recovering* the paths when `jsreplay.go:2462` calls what it drops the "offline js-concat mirror", which exists only because the static pass already wrote it. All ten now agree.
- **Cause of the GET-only classification.** The REST-counts note attributed `rest-api`'s GET-only reading of `/api/login` and `/api/upload` to "two-stage `crawl` + `generate --probe=false`, so no JavaScript executes". `crawl` defaults to `--headless=true` and does run JS; the real cause is the runner's `--headless=false` leg (`run-live-tests.sh:568`, copied at `:596`, generated from at `:600`). Following the documented commands produced GET+POST, the `scan-rest` result, which then fails `assert_path_methods`. Both fixture descriptions already had this right, so the README was the lone dissenter.
- **`protoc` prerequisite.** It quoted `protoc not installed — skipping .proto compile check (AC4)`, which exists nowhere in the repo, and contradicted the "No `protoc` required." note 170 lines below. The AC4 check is unconditional (`run-live-tests.sh:1621`). Reframed to what `protoc` is still for: regenerating `grpc-server/labpb/` via `make proto`.
- **Validator coverage.** The Node.js bullet claimed the spec validators are what "every OpenAPI and GraphQL target validates through". `test/validate.sh` defines only `validate_openapi_structure` (`:169`) and `validate_graphql_structure` (`:521`), so the WSDL and `.proto` producers have no Node validator at all. The claim is removed rather than re-scoped.
- **Rod captures.** "Spec generation always uses the net/http capture" is false for `soap-service`, whose WSDL comes from a synthetic `soap-capture.json` (`run-live-tests.sh:1350` says so outright). The claim is now scoped to the three targets that crawl with both backends, where it holds.
- **The deviation sentence.** It sent readers to a `scan-rest` section that does not exist, omitted `soap-service`, and called the two-stage shape the one "most targets use" — false under every denominator, four of thirty-two dispatch targets. The quantifier is gone and the list no longer reads as closed.
- **`npm ci`.** The three operator instructions omitted `--ignore-scripts`, which every executable call site uses and `setup-live-targets_test.sh:844` asserts.
- **`grpc-server` results tree.** Omitted `proto-validate.err`, which the runner writes on every run (`run-live-tests.sh:1595`, via the `2>` redirect at `:1622`).

`test/validate_test.sh` 95 passed / 0 failed and `check-docs.py` ok at every step; CI is green.

santiago-praetorian (approved):

**PR:** https://github.com/praetorian-inc/vespasian/pull/215
**Linear:** https://linear.app/praetorianlabs/issue/LAB-3672/update-testreadmemd-with-new-expected-live-test-results
**Head SHA:** 533fc3a01bb4e06edf5947f426daa19b5b48b867
**Reviewed at:** 2026-09-03T22:06:15Z

**Recommendation:** APPROVE

|---|---|---|---|---|---|
| Count | 0 | 0 | 0 | 0 | 0 |

---

PR #216, opened 2026-08-19
ci: enforce an 85% statement-coverage gate (LAB-5331)

eli-wald:
## What

Adds a **local test-coverage gate** to CI (Option 2 from LAB-5331): a `make coverage-gate` target that parses the total from `go tool cover -func=coverage.out` and exits non-zero when total statement coverage is below `COVERAGE_THRESHOLD` (default **85**), run from a new sibling job in `ci.yml`.

Closes LAB-5331.

## Why

`AGENTS.md` and `CONTRIBUTING.md` claimed CI enforced an 80% coverage threshold. It did not — the reusable `go-ci.yml` measures coverage but has no `fail_under`, and Vespasian never set `upload-coverage`, so the profile was produced and discarded. A PR could drop coverage arbitrarily and still go green, undercutting the "a green PR is safe to merge" goal.

Option 2 (local gate) is the cheapest honest gate: no third-party service, no `CODECOV_TOKEN`, and **no new egress**, so it does not couple to the harden-runner audit→block work in LAB-4732.

## Threshold: 85% (baseline measured first)

Per the acceptance criteria, baseline was measured before choosing the number:

- `make coverage` on `main`: **86.4%** total statement coverage, stable across two runs (`test/` already excluded from the profile).
- **85%** leaves ~1.4% headroom for minor rounding/refactor noise while still failing a build on a real regression, and sits well above the legacy (never-enforced) 80% claim. Won't block anything on day one.

## Changes

- **`Makefile`** — `COVERAGE_THRESHOLD ?= 85` (single source of truth) and a `coverage-gate` target. It rebuilds the profile via `coverage`, parses the `total:` line, exits 1 below threshold, and **fails closed** (exit 2) if that line is ever absent. The threshold reaches awk through the environment (`export COVERAGE_THRESHOLD`, read via `ENVIRON[]`) rather than interpolated into the awk command, so its value is never parsed by a shell — closing an injection the earlier single-quoting did **not** stop (`100' 'BEGIN{exit 0}` smuggled a second program argument into `awk`). The enforcement now lives in a profile-only `coverage-gate-check` target, with `COVERAGE_PROFILE`/`COVERAGE_FUNC` seams, that `coverage-gate` runs after `coverage`.
- **`.github/workflows/ci.yml`** — new `coverage-gate` job (a reusable-workflow job can't carry extra steps, so a separate job is required). Runs on `ubuntu-24.04` and SHA-pins harden-runner (audit) + `checkout`/`setup-go`, matching the `comment-claims` / live-tests convention so the whole workflow stays on one explicitly chosen image; `CGO_ENABLED=1` for `-race`. Inherits the workflow's Go-only `paths` filter, so it runs exactly when coverage can change. The job **id** now matches its concept (`coverage-gate`, not `coverage`) so id, Make target, and docs agree. A **"Self-test the coverage gate"** step runs the regression test below before the gate itself, mirroring the existing "Self-test the claims checker" step.
- **`scripts/coverage-gate_test.sh`** (new) — regression self-test for the gate's awk program (see Verification).
- **`AGENTS.md`** — corrected the CI line that claimed an 80% threshold. (The claim the issue attributes to `CLAUDE.md` now lives here: `CLAUDE.md` was made an `@AGENTS.md` pointer in #204.)
- **`CONTRIBUTING.md`** — coverage section rewritten to describe the enforced 85% gate; added `make coverage-gate` to the command list.
- **`README.md`** — added `make coverage-gate` to the Build-and-Test target list (it previously listed only `make coverage`).
- **`pkg/sdk/capability_test.go`** — a stray test comment referencing "the 80% threshold" made generic.

## Verification

- `make coverage-gate` at HEAD → `PASS: coverage meets the threshold` (86.4% ≥ 85%), exit 0.
- `COVERAGE_THRESHOLD=99 make coverage-gate` → `FAIL: coverage 86.4% is below the 99% threshold`, non-zero exit — the gate bites.
- **Failing path proven in-repo (AC3):** `scripts/coverage-gate_test.sh` runs in two layers. It extracts the gate's awk program live from `make -n coverage-gate-check` (so it can't drift from the recipe) and asserts exact exit codes for every documented mode — coverage below threshold → exit 1, non-numeric/empty/negative threshold → exit 2, threshold > 100 → exit 2, exactly-100 accepted → exit 0, missing `total:` line → exit 2 (fail-closed), plus at/above-threshold, decimal, and single-digit (guarding the `+ 0` numeric coercion) cases; the threshold-error cases also assert the "no total:" message is absent, pinning the END `bad`-guard branch. It then drives the **real** `coverage-gate-check` target against fixtures through the `COVERAGE_FUNC` seam (no Go build) and asserts make's own exit code, catching three regressions the fragment cases cannot see: a `-` recipe prefix (make swallows the FAIL), a dropped `+ 0` (numeric compare turns lexicographic), and a zeroed default threshold. A case-count guard fails the suite if a case is silently deleted. All 15 cases pass, and the test runs in CI as a "Self-test the coverage gate" step before the gate — the layer that matters, since the `+ 0` regression is caught by the runner's gawk even where the local onetrueawk masks it. This **demonstrates rather than assumes** that the gate fails a below-threshold build.

## Note for repo admins (follow-up, not in this PR)

Branch protection: decide whether to mark the `coverage-gate` check **required**. Because `ci.yml` is Go-path-filtered, a docs-only PR won't trigger it — if marked required without a "skipped = success" arrangement, such PRs could hang. This matches the existing `ci` job's behavior; no protection change is made here.


santiago-praetorian:
## Capability PR review — LAB-5331

**Verdict: approve.** 0 blockers, 0 critical, 0 high, 33 low, 13 nit across 46 findings. CI is green including this PR's own new gate (`total coverage 87.0% (threshold 85.0%)`).

Reviewed at `05a77d8`. Exhaustive sweep, three lanes, all converged (`capability-reviewer` 12/12 rubric items over 4 passes · `backend-security` 18/18 over 2 · `test-lead` 9/9 over 3; `frontend-security` not dispatched — no frontend files).

### Acceptance criteria: all five met

AC3 was the one worth checking properly, since the demonstration ran on an **earlier revision** of the gate and worked by raising the threshold to 99 rather than by dropping coverage. It still holds: the decisive branch

```awk
if (pct + 0 < threshold + 0) { printf "FAIL: ..."; exit 1 }
```

is byte-identical between #219's `ed1ca0a` and this head, and the inequality is symmetric — the gate cannot distinguish "threshold raised above coverage" from "coverage dropped below threshold". #219's CI shows `coverage-gate` failing with `FAIL: coverage 86.2% is below the 99% threshold` / `make: *** [Makefile:82: coverage-gate] Error 1`, so the failing path is demonstrated on a real runner rather than assumed. (#217/#218 show no `coverage-gate` check because no `pull_request` run was ever created for their head SHAs — shared branch under a `cancel-in-progress` group — not because the job was missing.)

AC4 names `CLAUDE.md`, which is a one-line `@AGENTS.md` pointer since #204, so editing `AGENTS.md:202` *is* editing what `CLAUDE.md` says. A coverage-scoped grep for the old 80% claim returns zero hits repo-wide.

### Four things worth fixing, none blocking

Each was verified by running it against a fixture, with a green baseline before and after. All four are edits the new self-test does **not** catch, because it extracts the awk *fragment* from `make -n` and never executes the *target*:

| # | Where | Mutation | Verified result |
|---|---|---|---|
| 1 | `Makefile:89` | prefix the recipe with make's `-` | **`make` exits 0** while printing `FAIL: coverage 80.0% is below the 85.0% threshold` and `Error 1 (ignored)`. Control exits 2. |
| 2 | `Makefile:89` | drop both `+ 0` operands | Comparison becomes lexicographic: a **9.5% fixture PASSES an 85% gate** at exit 0. Control exits 1. |
| 3 | `Makefile:12` | `COVERAGE_THRESHOLD ?= 0` | All 11 cases stay green; gate passes any coverage. |
| 4 | `scripts/coverage-gate_test.sh` | delete the `non-numeric threshold fails closed` case | Suite prints `all 10 cases passed`, exit 0. |

All four close with the same small change: a `COVERAGE_PROFILE ?= coverage.out` seam plus a profile-only target, so a case can invoke real `make` against a fixture profile and assert make's exit code. Still ~1s, still no Go build. For #4 the repo already has the convention — `EXPECTED_ASSERTIONS=87` in `test/setup-live-targets_test.sh`.

For balance, 6 of the 20 mutations I ran **were** caught (the `exit 1` flip, deleting the BEGIN guard, `> 100`→`>= 100`, a sign-admitting regex, and both END guards), and 5 more survived only because the mutant isn't actually broken (`$NF`→`$3` is a no-op on the real `total:` line; deleting `sub(/%/…)` still converts correctly; the printf swap is log-only). The BEGIN validation added in this PR does its job.

### One correction to the PR description

> The threshold is single-quoted into the awk invocation (`threshold='$(COVERAGE_THRESHOLD)'`) so a future lower-trust source for the value can't inject at the shell level.

This doesn't hold. `COVERAGE_THRESHOLD="100' 'BEGIN { exit 0 } #"` expands to

```
awk -v threshold='100' 'BEGIN { exit 0 } #' 'BEGIN { seen = 0; ... }'
```

— the real program becomes a *file* argument and awk exits 0 without reading input. Against a 0% fixture the gate exits 0 where the control exits 1. It's `low` severity because the only path needs an actor who can already edit the `Makefile` or the workflow (and could just delete the job), so it grants nothing — but the sentence claims a property the code doesn't have, and this is the still-open review thread behind CodeRabbit's *Merge Risk: Moderate*. `export COVERAGE_THRESHOLD` + `threshold = ENVIRON["COVERAGE_THRESHOLD"]` removes the quoting layer entirely and keeps the BEGIN validation.

### Two facts from outside the diff

- **No ruleset on `main` declares any required status check** (six active rulesets, none with `required_status_checks`; the legacy endpoint 404s). So `CONTRIBUTING.md`'s "CI enforces a minimum of 85%" is true in the job-goes-red sense but not the merge-blocking one. Your admin note already flags this — worth a LAB ticket so the doc and the ruleset move together.
- **The job's documented expected-egress list is incomplete.** Its own run reaches `api.github.com`, `results-receiver.actions.githubusercontent.com`, `release-assets.githubusercontent.com`, `productionresultssa6.blob.core.windows.net`, and `example.com` (from `cmd/vespasian/main_test.go:1239`, which probes with `Probe: true`); the comment lists `sum.golang.org` and `storage.googleapis.com`, which don't appear. That comment is the artifact LAB-4732's audit→block work will read, so it's worth correcting from the `endpoint called` log lines.

Also: `AGENTS.md:202`'s new sentence says the `paths` filter is Go-only and applies to pushes. Neither is right — `paths:` is nested under `pull_request` only (so a push to `main` always runs), and it lists `go.mod`, `go.sum`, `.golangci*`, `Makefile` and `scripts/**` besides `**/*.go`. The previous wording was wrong in the opposite direction, so this traded one inaccuracy for two.

### All 46 findings

Nothing here withholds the merge. 39 are on lines this PR changed; 7 are pre-existing and belong on `main` (marked *pre*).

<details>
<summary>Full list</summary>

| ID | Sev | File:Line | Finding |
|---|---|---|---|
| `QUAL-007` | low | `AGENTS.md:202` | AGENTS.md CI-trigger claim 'paths filter is Go-only' is false |
| `QUAL-001` | low | `.github/PULL_REQUEST_TEMPLATE.md:15` | PR template checklist omits the new coverage-gate CI check *(pre)* |
| `QUAL-002` | low | `.github/workflows/ci.yml:86` | coverage-gate job has no timeout-minutes, unlike its sibling comment-claims job |
| `QUAL-003` | nit | `.github/workflows/ci.yml:95` | coverage-gate job duplicates comment-claims job's harden-runner/checkout boilerplate verbatim |
| `QUAL-004` | low | `.github/workflows/ci.yml:116` | Stale 'ubuntu-latest' comment contradicts the pinned ubuntu-24.04 runner |
| `QUAL-005` | low | `.github/workflows/ci.yml:117` | New coverage-gate job has no structural drift-protection test, unlike live-tests.yml jobs |
| `QUAL-006` | low | `AGENTS.md:38` | The 85% threshold is restated in prose five times across three docs, with no test tying them to COVERAGE_THRESHOLD |
| `QUAL-008` | nit | `AGENTS.md:209` | AGENTS.md CI section for docs-check repeats the same stale 'paths filter is Go-only' claim (context, informational) *(pre)* |
| `QUAL-009` | low | `CONTRIBUTING.md:73` | CONTRIBUTING.md quick-reference `make check` line omits lint-comments/check-docs *(pre)* |
| `QUAL-010` | low | `CONTRIBUTING.md:236` | Docs never mention that coverage-gate reruns the full -race test suite |
| `QUAL-011` | low | `Makefile:80` | Makefile 'fails closed' guarantee comment does not name the test that pins it |
| `QUAL-012` | low | `Makefile:88` | coverage-gate's self-test is not wired into the Makefile, unlike its lint-comments sibling |
| `QUAL-013` | low | `Makefile:89` | Single-quoting COVERAGE_THRESHOLD does not prevent shell injection |
| `QUAL-014` | low | `Makefile:89` | coverage-gate awk one-liner mixes validation, parsing, comparison, and presentation in a single 747-char recipe line |
| `QUAL-015` | low | `Makefile:89` | go tool cover -func=coverage.out is invoked twice per coverage-gate run |
| `QUAL-016` | low | `README.md:598` | README.md `make check` comment omits lint-comments/check-docs *(pre)* |
| `QUAL-017` | nit | `README.md:603` | README.md coverage-gate line breaks the file's comment-column alignment convention |
| `SEC-BE-001` | low | `.github/workflows/ci.yml:6` | paths filter lets a coverage-affecting PR skip the new gate entirely *(pre)* |
| `SEC-BE-002` | nit | `.github/workflows/ci.yml:17` | cancel-in-progress can cancel the gate for a commit that lands on main *(pre)* |
| `SEC-BE-003` | low | `.github/workflows/ci.yml:86` | Add timeout-minutes to the new coverage-gate job |
| `SEC-BE-004` | nit | `.github/workflows/ci.yml:86` | Confirm coverage-gate is a required check before documenting it as enforced |
| `SEC-BE-005` | nit | `.github/workflows/ci.yml:91` | Complete the documented expected-egress list for the coverage-gate job |
| `SEC-BE-006` | low | `.github/workflows/ci.yml:117` | Coverage percentage now depends on live third-party network reachability |
| `SEC-BE-007` | nit | `.github/workflows/ci.yml:117` | coverage-gate re-runs the full -race suite the ci job already ran |
| `SEC-BE-008` | nit | `AGENTS.md:202` | Correct AGENTS.md: the paths filter does not apply to pushes |
| `SEC-BE-009` | low | `Makefile:75` | Substring package exclusion silently shrinks the gated coverage denominator *(pre)* |
| `SEC-BE-010` | low | `Makefile:88` | Wire the gate self-test into the Makefile target it guards |
| `SEC-BE-011` | nit | `Makefile:89` | Reject a zero/absurdly-low COVERAGE_THRESHOLD as a mis-set knob |
| `SEC-BE-012` | low | `Makefile:89` | Pass COVERAGE_THRESHOLD via the environment, not shell single quotes |
| `SEC-BE-013` | low | `scripts/coverage-gate_test.sh:55` | Make the awk-program isolation guard able to fail: it is tautological |
| `SEC-BE-014` | low | `scripts/coverage-gate_test.sh:161` | Assert the expected case count so the gate self-test cannot be hollowed out |
| `TEST-002` | low | `Makefile:12` | Assert the recipe's threshold binding, not only the extracted awk program |
| `TEST-004` | low | `Makefile:89` | Execute the coverage-gate Make target once; today only the awk fragment runs |
| `TEST-009` | low | `scripts/coverage-gate_test.sh:108` | Add a single-digit coverage fixture; all five fixtures share one magnitude |
| `TEST-001` | low | `.github/workflows/ci.yml:113` | Pin the coverage-gate job's existence and self-test-before-gate step order in ci.yml |
| `TEST-003` | low | `Makefile:88` | Wire the gate self-test as a Make prerequisite, as lint-comments already does |
| `TEST-005` | low | `scripts/coverage-gate_test.sh:19` | Discriminate the awk implementation the suite claims to guard, or drop the claim |
| `TEST-006` | low | `scripts/coverage-gate_test.sh:45` | Stop discarding make's stderr in the extraction step |
| `TEST-007` | low | `scripts/coverage-gate_test.sh:57` | Fix the extraction-failure guard: it can never fire |
| `TEST-008` | nit | `scripts/coverage-gate_test.sh:74` | Reuse the sibling's leading-! inversion instead of a sixth positional parameter |
| `TEST-010` | nit | `scripts/coverage-gate_test.sh:117` | Assert the reported percentages, not just the words PASS and FAIL |
| `TEST-011` | nit | `scripts/coverage-gate_test.sh:149` | Number the last two cases: reviewers already cite case numbers the file lacks |
| `TEST-012` | low | `scripts/coverage-gate_test.sh:162` | Pin the total case count so a deleted assert_gate call cannot pass silently |
| `DOC-001` | low | `AGENTS.md:202` | Correct the ci.yml trigger description: the paths filter is neither push-wide nor Go-only |
| `DOC-002` | low | `CONTRIBUTING.md:236` | Soften "CI enforces" -- no ruleset on main requires the coverage-gate check |
| `DOC-003` | nit | `Makefile:10` | Refresh or date-stamp the 86.4% baseline figure cited in two places |

</details>

Four findings were rated `medium` by their lanes and downgraded to `low` on review: all four premises were verified to **hold**, but their consequence requires a future code edit rather than being reachable under default settings, and regression exposure is `low` by our own severity rules. Two `test-lead` findings on `ci.yml` were dropped by the lane filter; both are filed independently as `SEC-BE-003` and `SEC-BE-005`, so no substance was lost.

---

santiago-praetorian (approved):
Approving. Exhaustive multi-agent review at `05a77d8`: 0 blockers, 0 critical, 0 high, 46 findings all low/nit, all five LAB-5331 acceptance criteria met with the failing path demonstrated on a real CI run (#219). Detail in the review comment above.

The four items worth picking up are non-blocking but real, and all four are edits the new self-test cannot catch because it extracts the awk fragment rather than executing the target — most notably make's `-` prefix (job exits 0 while printing `FAIL`) and dropping `+ 0` (9.5% passes an 85% gate). A `COVERAGE_PROFILE` seam plus a profile-only target closes all four.

One correction regardless of merge: the PR description's claim that single-quoting stops shell-level injection into `awk -v` does not hold — that is the still-open thread behind CodeRabbit's Merge Risk banner. It grants no privilege (the actor must already be able to edit the Makefile), so it does not block, but the sentence should go or change.

eli-wald:
Thanks for the exhaustive review (#pullrequestreview-5035440328). Pushed `7cca8b5` addressing the substantive items:

**The four mutation-gaps — closed with the `COVERAGE_PROFILE` seam you proposed.** Split a profile-only `coverage-gate-check` target out of `coverage-gate`, added `COVERAGE_PROFILE`/`COVERAGE_FUNC` seams, and extended `scripts/coverage-gate_test.sh` with a second layer that drives the **real** target against fixtures (via `COVERAGE_FUNC='cat …'`, still no Go build) and asserts make's own exit code:
- **`-` recipe prefix** → real-make cases 13/14 catch it (make exits 0 while printing FAIL → caught).
- **dropped `+ 0`** → a single-digit fixture (9.5% vs 85) catches it. Worth flagging: on this machine's onetrueawk the mutation *isn't* caught (strnum semantics), but under the CI runner's **gawk** it is — exactly the "a local green says nothing about the runner" case the header warns about, so the CI wiring is doing real work here.
- **`COVERAGE_THRESHOLD ?= 0`** → the default-threshold real-make case catches it.
- **deleted test case** → a `EXPECTED_CASES` count guard fails the suite (the `EXPECTED_ASSERTIONS` convention). 15/15 pass.

**Injection correction (code + description).** Switched to `export COVERAGE_THRESHOLD` + `threshold = ENVIRON["COVERAGE_THRESHOLD"]`, so the value never transits a shell. Verified `COVERAGE_THRESHOLD="100' 'BEGIN{exit 0}"` now fails "not a number" (exit 2) instead of short-circuiting. The PR description's single-quoting sentence is replaced.

**AGENTS.md:202.** Corrected — the `paths` filter is under `pull_request` only (pushes unfiltered) and covers `go.mod`/`go.sum`/`.golangci*`/`Makefile`/`scripts/**`, not Go files alone. Aligned the docs-check line (209) for consistency.

**Egress comment (ci.yml).** Rewritten to the endpoints the audit run actually reaches (api.github.com, results-receiver/release-assets/blob Actions infra, and example.com from `main_test.go`'s `Probe:true` path); dropped `sum.golang.org`/`storage.googleapis.com`.

**Not in this PR:** the "no ruleset marks any check required" gap — leaving that as a separate LAB follow-up so the doc and ruleset move together, per your note.

santiago-praetorian:
## Capability PR review — LAB-5331 (round 2)

**Verdict: approve.** 0 blockers, 0 critical, 0 high, 34 low, 13 nit across 47 findings. Reviewed at `7cca8b5`, incremental scope against the round-1 watermark `05a77d8`. CI green, including this PR's own gate: `total coverage 87.0% (threshold 85.0%)`.

### The four mutations from round 1 are closed — verified, not assumed

I re-ran each one against a fixture in a throwaway worktree, judging by exit code only, with a green baseline (`all 15 cases passed`) before and after, and a per-mutation match count so a `sed` that silently matched nothing couldn't read as a pass:

| Mutation | Round 1 | Now |
|---|---|---|
| `-` recipe prefix (`Makefile:126`) | survived | **caught** — 2/15 fail |
| drop both `+ 0` operands (`Makefile:126`) | survived | **caught** — 2/15 fail |
| `COVERAGE_THRESHOLD ?= 0` (`Makefile:15`) | survived | **caught** — 1/15 fail |
| delete a whole case | survived | **caught** — `EXPECTED_CASES` guard fires |

16 of round 1's 46 findings are resolved, 0 dropped unexplained, and **0 previously-missed** — the convergence signal that makes another exhaustive round unnecessary.

### The one thing I'd still fix

Both layers of `scripts/coverage-gate_test.sh` target **`coverage-gate-check`** (extraction at `:69`, execution at `:143`/`:145`). CI runs **`make coverage-gate`** (`ci.yml:126`). Nothing in the repo greps or executes that target, so three one-line edits leave the suite green while enforcement disappears:

| Edit | Suite | Effect |
|---|---|---|
| `Makefile:109` — `coverage-gate: coverage coverage-gate-check` → `coverage-gate: coverage` | **15/15 green** | `make -n coverage-gate` emits **0** awk lines (1 unmutated). CI then runs only `go test` + `go tool cover`, both exit 0 at any coverage. |
| delete `ci.yml`'s *Enforce coverage threshold* step | **15/15 green** | gate never invoked |
| delete `ci.yml`'s *Self-test the coverage gate* step | **15/15 green** | self-test never invoked |

This is a genuine coverage regression rather than a pre-existing gap: before this commit the extraction anchored on `make -n coverage-gate`, which transitively asserted *the gate is part of what CI runs*. The split moved the gate one level below the anchor and it followed. It also makes the comment at `:29` — "the test exercises exactly what CI runs and cannot drift from it" — untrue as written.

**Cheapest fix:** one `assert_make` case driving `coverage-gate` itself, or restore the extraction anchor to `make -n coverage-gate`.

I rated this `low` rather than `medium` on the reachability floor — the gate works correctly today and the failure needs a future edit — so it does not block. That call is arguable and I'd have no objection to it being treated as merge-blocking.

### Two smaller survivors from the same run

- **`COVERAGE_THRESHOLD ?= 81` passes green.** Case 13 pins the default only to `>80`, so a 4-point silent loosening is invisible. (`?= 0` *is* caught, so the guard exists — it's just loose.)
- **A case can lose its trailing `want_absent` argument silently.** The paired control settles that this matters: removing `if (bad) exit bad` from the awk `END` block fails 4 cases, **all four on the `want_absent` check and none on the exit code** — so that argument is the sole discriminator for that defect.

### Two comment claims that don't match the artifact

- **`ci.yml:93-100`** — the rewritten egress list omits `github.com`, which is the *first* `endpoint called` line in this job's own audit telemetry (`git-remote-http`, from checkout). The comment names itself the input to LAB-4732's allowlist, so an allowlist built from it as written would omit the host checkout needs. `proxy.golang.org` is also written as "may appear" but did fire twice.
- **`ci.yml:121`** — comment says `ubuntu-latest` ships gcc; the job pins `runs-on: ubuntu-24.04`. Still open from CodeRabbit's unresolved thread.

### Acceptance criteria: 4 met, 1 partially met

**AC3 is partially met.** The demonstration on #219 is real (job 96463673591 failed with `FAIL: coverage 86.2% is below the 99% threshold`), and the "wrong direction" objection is **not** a gap — `if (pct + 0 < threshold + 0)` is symmetric, so raising the threshold exercises the identical branch. The modest gap is staleness: the demo ran on the pre-split recipe. Combined with the wiring point above, the end-to-end chain at this head is verified only piecewise.

**AC1 and AC2** are answered in Linear *comments*, not the ticket description — both met, but anyone reading only the description will score them open. **AC4** names `CLAUDE.md`, a one-line `@AGENTS.md` pointer since #204; the fix correctly landed in `AGENTS.md` and a repo-wide grep finds no surviving stale 80% claim, so it's met in substance and the criterion's wording is now stale on the ticket.

### One residual I could not close

The `+ 0` guarantee is awk-implementation-dependent. I confirmed **mawk** catches the mutation; `gawk` and `original-awk` aren't installed in my environment, so the four-interpreter matrix is incomplete and the arm you flagged remains unmeasured.

---

*Review generated by `capability-pr-review` (incremental mode; `capability-reviewer` 12/12 · `backend-security` 18/18 · `test-lead` 9/9; `frontend-security` not dispatched). Full record: 47 findings.*

santiago-praetorian (approved):
Approving — full review posted as a separate comment above.

**0 blockers, 0 critical, 0 high** (34 low, 13 nit across 47 findings) at `7cca8b5`. CI green including this PR's own gate at `total coverage 87.0% (threshold 85.0%)`. All four mutations round 1 proved uncatchable are now caught — I re-ran each against a fixture rather than taking it on trust.

The one item I'd still like fixed is non-blocking by my rating but arguable: both layers of `scripts/coverage-gate_test.sh` test `coverage-gate-check`, while `ci.yml:126` runs `make coverage-gate`, and nothing pins the link between them. Details and the cheapest fix are in the comment.

PR #180, opened 2026-07-17
fix(crawl): harden crawler egress — pin system Chrome (no auto-download) + disable telemetry (LAB-4999)

santiago-praetorian:
## Summary

Hardens the vespasian headless crawler's egress ([LAB-4999](https://linear.app/praetorianlabs/issue/LAB-4999/vespasian-harden-crawler-egress-pin-system-chrome-no-auto-download)), a follow-up to the harden-runner audit added in #178 (LAB-4732). Two sources of non-essential outbound traffic, both rooted in how `pkg/crawl/browser.go` `configureLauncher` launches Chrome, are eliminated so a tight block-mode egress allowlist becomes practical.

**Finding 1 — supply chain (no auto-download):** `configureLauncher` never set `.Bin`, so go-rod auto-downloaded a managed Chromium from third-party mirrors (`storage.googleapis.com`, `registry.npmmirror.com`, `playwright.*`) at launch. It now pins the system browser via `launcher.LookPath()` + `.Bin()`; when none is found it **errors rather than downloading**, unless explicitly opted in via `BrowserOptions.AllowBrowserDownload` or `VESPASIAN_ALLOW_BROWSER_DOWNLOAD=true` (keeps local dev working on platforms without a system Chrome, e.g. arm64 Linux).

**Finding 2 — telemetry/phone-home:** adds `--disable-component-update`, `--disable-domain-reliability`, `--no-pings`, and appends `OptimizationHints,AutofillServerCommunication` to `disable-features` (via `Append`, preserving go-rod's `site-per-process`/`TranslateUI` defaults).

Scope note: this affects **every** vespasian crawl, not just CI — a legitimate hardening for a security crawler (blast radius flagged for reviewer awareness).

## Changes
- `pkg/crawl/browser.go` — `pinBrowserBinary` + `disableChromeTelemetry` helpers; new `BrowserOptions.AllowBrowserDownload`; `pkg/crawl/doc.go` note.
- `pkg/crawl/browser_test.go` — unit tests (all new branches; new funcs 100% covered) via a minimal `browserLookPath` test seam.
- `pkg/crawl/browser_integration_test.go` — dynamic `//go:build integration` assertion that a real launch pins a system binary (no download).
- `test/run-live-tests.sh` — `no-download` live guard (snapshots go-rod's browser cache around a headless crawl; fails if a new `chromium-<rev>` appears), registered in `LIVE_TARGETS`.
- `test/test-runner-args.sh` — drift-guard count literals (6→7 live, 25→26 total).
[...]

PR #226, opened 2026-08-23
ci(security): flip live-tests harden-runner to block with per-job egress allowlists (LAB-6015)

santiago-praetorian:
Closes LAB-6015.

## What

Five of the six non-container jobs in `live-tests.yml` ran `egress-policy: audit`. Audit observes egress; it does not restrict it — so the workflow that drives a browser and an HTTP crawler had unrestricted outbound network access and nothing failed when a step reached an unexpected host. This flips all five to `block` with a per-job `allowed-endpoints` list.

LAB-4732 landed audit deliberately, because crawler egress was non-deterministic at the time. LAB-4999 (#180) removed that obstacle by pinning the system browser, so the flip has been actionable since 2026-07-27.

## Where the lists come from

Not guessed. The union of the `endpoint called` lines in the harden-runner post-step output of six green `live-tests` push runs on `main`:

| run | head |
| -- | -- |
| 32500986374 | `393bf02` |
| 32494976039 | `b575e67` |
| 32398237813 | `c0bb167` |
| 32395398188 | `d2a4e90` |
| 32382787790 | `f6c0624` |
| 32297452612 | `ff299ba` |

read back with `gh run view <id> --log`. `endpoint called` (a connect that happened) rather than `domain resolved` (a DNS lookup that may never be dialled) — the two differ here.

Three endpoints that had been assumed necessary turned out not to be:

- `objects.githubusercontent.com` and `nodejs.org` are never contacted — `setup-go`/`setup-node` fetch from `release-assets.githubusercontent.com`.
- `storage.googleapis.com` appears on **no** list. It was only ever reached by `preflight-selftest`'s `GOTOOLCHAIN` fetch (that job has no `setup-go`, so the `go version` call in `check_prerequisites` pulled a toolchain); setting `GOTOOLCHAIN: local` on the job removes the fetch, and with it that endpoint plus `proxy.golang.org` and `sum.golang.org` from its list.

## Per job, not shared

| job | endpoints |
| -- | -- |
| `preflight-selftest` | 2 |
| `validator-regression` | 6 |
| `docs-check` | 2 |
| `integration-tests` | 7 |
| `test` | 8 |

`docs-check` and `preflight-selftest` need two endpoints each; a single shared list would over-grant them by six.

## What the flip actually removes

Run [32651870901](https://github.com/praetorian-inc/vespasian/actions/runs/32651870901) (head `9266744`) was the first execution under `block` — all five jobs green, and these were refused with no effect on any assertion. Note that run predates the later commits that tightened `preflight-selftest`'s list from five entries to two and added the AC3 enforcement step, so see the acceptance-criteria section below for a run against the current head:

| blocked | job(s) | why it is safe |
| -- | -- | -- |
| `clients2.google.com`, `www.google.com`, `google.com`, `update.googleapis.com` | `integration-tests`, `test` | Chrome's component updater and connectivity probes. LAB-4999's telemetry-suppressing flags reduced this but did not eliminate it — 149 and 134 calls respectively in one audit-mode run *under those flags*. No assertion reads them. |
| `example.com` | `integration-tests` | `TestDoCrawl_ProxyPortlessWarning` crawls it through a deliberately broken proxy and discards the result (`_, _ = doCrawl(...)`), asserting only on a stderr warning. |
| `hosted-compute-{watchdog,request-orchestrator}-prod-*.githubapp.com` | three jobs | GitHub's own host agents. Blocked in three jobs, all green. The region suffix rotates, so listing them means a wildcard over `githubapp.com` for traffic no step of ours makes. |

## Two costs, recorded rather than hidden

- **`*.blob.core.windows.net:443` is wider than any job needs.** harden-runner resolves the job's *own* Actions results storage account at start-up and injects it (`productionresultssa13` shows up in the effective allow-set of all five jobs, including the two whose lists name no blob endpoint) — that is why `preflight-selftest` and `docs-check` need no entry. It does **not** cover the Actions cache or `upload-artifact`: in the same run the `test` job used `productionresultssa2` *and* `productionresultssa14` alongside the injected `sa13`. Three accounts in one job, so no tighter form exists.
- **`sum.golang.org:443` in the two `setup-go` jobs is a deliberate superset**, not an observation. `setup-go` restored the module cache in all six runs, so nothing verified against the checksum DB. A cache miss takes the path that does, and a blocked sumdb fails `go mod download` in a way that reads as a network flake rather than a policy gap.

## Regression guard

`test/test-runner-args.sh` gains a per-job pin. It compares each job's WHOLE policy against a pinned expectation — the action SHA, `block`, `disable-sudo`, no `if:`, harden-runner as the job's first step, the `allowed-endpoints` node type, and the exact endpoint set — rather than checking its shape, because a shape check ("is it `block`, is the list non-empty") let eight mutations through. The job list is HARDCODED, not derived: a derived list cannot see a job that should carry the step but does not, and deleting a step removed an assertion instead of failing one. Four further pins cover the directions a per-job comparison cannot: the full seven-job set, the set of jobs that carry harden-runner (the only check that catches the step being ADDED to a job meant to be exempt), the AC3 enforcement step (its position, and that it still invokes `test/assert-egress-enforced.sh`), and the two exempt jobs' exemption rationale. 135 → 144 assertions, measured. The AC3 probe itself lives in `test/assert-egress-enforced.sh` rather than inline in the workflow: pinning inline shell proved unable to be exhaustive across five review rounds, so the pin is one exact invocation string. The script's CONTENTS are deliberately not pinned — the same trust boundary this suite already applies to the four guard suites, whose invocation it pins rather than their bodies.

Mutation-proven — each of these left every other assertion in the suite green and was caught only by the new pin:

- `preflight-selftest` reverted to `egress-policy: audit` → 143 passed, 1 failed
- the `test` job's `allowed-endpoints` emptied → 143 passed, 1 failed
- a second harden-runner step on `audit`; the allowlist collapsed to `*:443`; an extra endpoint; the SHA zeroed; a look-alike action name; `if: false`; the step moved below checkout; `disable-sudo` deleted → each caught
- a new job added *without* a harden-runner step; `allowed-endpoints` rewritten as a YAML sequence; the AC3 step deleted, moved off the end of the job, or repointed away from `test/assert-egress-enforced.sh`; ANY edit to `check-label`'s step body, including `gh api`, `python3 -c urllib` and `exec 3<>/dev/tcp` → each caught

## Acceptance criteria

- [x] Every `harden-runner` step sets `block` with a job-scoped `allowed-endpoints`.
- [x] Derived from audit telemetry; the six source runs are named here, on the ticket, and in the workflow header.
- [x] A deliberate request to an unlisted host fails the job — scratch PR #227, run [32651913938](https://github.com/praetorian-inc/vespasian/actions/runs/32651913938): `curl https://proxy.golang.org/` failed in 1 ms with `domain not allowed: proxy.golang.org.` while a control request to `github.com` succeeded in the same job. Closed unmerged.
- [x] A normal run is green under `block` across all five jobs — run [33080599136](https://github.com/praetorian-inc/vespasian/actions/runs/33080599136), head `bd362ec`, all five policy jobs `success` (and the AC3 enforcement step green in its new last position) (`install-chrome-e2e` skipped, as it is opt-in). Each push produces a new run, so the latest run on this branch is the authoritative one; this citation names the head it covers so a reader can tell when it has gone stale. **The literal criterion is not satisfiable before merge** — it asks for a run on `main`, and `main` still carries the four `audit`-mode TODOs. Treating this as met on the PR-run evidence, with the `main` run following on merge, is the wording decision this box records.
- [x] The stale `"'block' ... once telemetry confirms the set"` comments are removed.
- [x] `install-chrome-e2e` keeps no harden-runner step; the action does not support container jobs and its own comment records that.

`ci.yml` is still `audit` and stays out of scope per the ticket — its needs are close to but not identical to `integration-tests`', so it wants its own telemetry pass rather than this list copied across.

## Follow-ups filed from the security review (out of scope here)

Round 8 of `/capability-pr-review` dispatched a security lane for the first time — the skill's
trigger table keys on file extension and lists neither `.sh` nor `.yml`, so a change whose whole
subject is a CI egress control had gone seven rounds without one. It found six items. Four were
in scope and are fixed in this PR. Two are not this ticket's, and are tracked rather than dropped:

- **LAB-6137** — `disable-telemetry` is now an available choice (it requires `block`, so this PR is
  what unlocked it), and `disable-sudo` is the deprecated spelling upstream:
  `disable-sudo-and-containers` supersedes it and is strictly stronger, since it also blocks the
  `container:` route this PR closed statically. Deliberately not changed here: the telemetry flip
  needs a check that harden-runner still logs `endpoint called` with it off, because the
  allowlist-provenance process behind AC2 reads exactly those lines.
- **LAB-6138** — `verify-pins.yml` scopes itself to first-party `praetorian-inc/public-workflows`
  pins, and this file contains none, so the version comments on `actions/checkout`, `setup-go`,
  `setup-node` and `upload-artifact` are verified by no check. Pre-existing and repo-wide;
  `harden-runner`'s own SHA is the one that IS pinned, by `test/test-runner-args.sh`.

`AC4` remains open by construction rather than by omission — it asks for a green run on `main`
under `block`, and the block-mode workflow exists only on this branch. Measurements and two
wording options are recorded on LAB-6015.


eli-wald (changes requested):

**PR:** https://github.com/praetorian-inc/vespasian/pull/226  
**Linear:** https://linear.app/praetorianlabs/issue/LAB-6015/vespasian-flip-live-tests-harden-runner-from-egress-policy-audit-to  
**Head SHA:** d24f2cbf268e7ba87622def0115b0564c5388b07  
**Reviewed at:** 2026-08-26T15:15:05Z

|---|---|---|---|---|---|
| Count | 0 | 0 | 1 | 1 | 1 |

|---|---|---|
| capability-reviewer | 12/12 | 6 |
| test-lead | 9/9 | 1 |

**Vs previous run** (head `d24f2cbf268e`, identical to current head): carried 0 · new 0 · **previously-missed 2** · previously-missed? 1 · dropped 2

> The prior run was against the **same head SHA**, so the determinism contract expects an identical finding set. It did not reproduce: 0 carried, 2 previously-missed, 1 previously-missed?, 2 dropped. This is agent-judgment variance (the contract's documented "honest residual"), not a code change. TEST-001 below is the same runtime-enforcement gap the prior run's blocker raised, re-expressed as a file-level finding.
- `capability-reviewer item 3  n_a -> clean`
- `capability-reviewer item 4  clean -> findings`
- `capability-reviewer item 6  clean -> n_a`
- `capability-reviewer item 8  clean -> n_a`
- `capability-reviewer item 11  n_a -> clean`
- `test-lead item 9  clean -> n_a`

---

<details>

<details>
<summary><code>TEST-001</code> · medium · TEST · <code>.github/workflows/live-tests.yml:None</code></summary>

**Issue:** The regression guard added in test/test-runner-args.sh verifies the STATIC YAML of every harden-runner step (SHA, egress-policy: block, disable-sudo, no if:, first step, exact allowlist set). It does not, and cannot, verify that block mode enforces at runtime. A normal green live-tests run under `block` is indistinguishable from `audit` unless something is actually refused and that refusal is asserted — precisely the 'assertion that passes by not executing' failure mode the parent epic (13T-146) and this ticket's AC3 call out. AC3 ('a deliberate request to an unlisted host fails the job') was demonstrated only by scratch PR #227 (run 32651913938: curl to proxy.golang.org refused with 'domain not allowed'), which was CLOSED UNMERGED. grep of test/ confirms no standing assertion of egress refusal exists; run-live-tests.sh:3704-3723 only catches the go-rod browser-download side-effect, not general unlisted-host blocking. So the enforcement half of the behavior this PR ships has no permanent regression coverage: a future change that silently reverts block to a non-enforcing state in a way the static pin still accepts (e.g. an upstream action behavior change, or block failing open) would not be caught. Mitigation to weigh: a standing test that makes a real egress call to an unlisted host adds nondeterminism and outbound traffic to CI, and harden-runner's enforcement is third-party behavior — reasonable grounds to accept the risk, but the decision should be recorded rather than left implicit.

**Recommendation:** Add a standing (opt-in or push-to-main) job step that makes one deliberate request to a host guaranteed absent from the allowlist and asserts it is refused (the #227 pattern: curl an unlisted host, expect failure; a control request to an allowlisted host must succeed in the same job), so block-mode enforcement is proven on an ongoing basis rather than once in a discarded PR. If a standing runtime check is judged not worth the CI egress/nondeterminism, record that as an explicit 'rely on existing controls' decision on LAB-6015 (the AC's documented second branch) so the gap is a choice, not an oversight.

```
# LAB-6015 flipped the five policy-carrying jobs in live-tests.yml from
# `egress-policy: audit` to `block` with a per-job allowlist. This block is what
# stops that from being silently undone.
#   It compares each job's policy against a PINNED EXPECTATION below rather than
#   checking its shape
```

</details>

</details>

<details>

<details>

<details>
<summary><code>QUAL-001</code> · low · <code>.github/workflows/live-tests.yml:147</code></summary>

**Issue:** The harden-runner step header is byte-identical across all five policy-carrying jobs (lines 147-152, 219-224, 286-291, 519-524, 593-598): the `name`, the SHA-pinned `uses`, `egress-policy: block`, `disable-sudo: true`, and the `allowed-endpoints: >` wrapper. Only the endpoint list values below differ per job. This is copy-paste of 6 contiguous non-comment lines x5, so a SHA bump or a policy-key change must be applied in five places. The author explicitly acknowledges this ('bump ALL FIVE copies in lockstep') and guards it with a per-job pin in test/test-runner-args.sh, which materially reduces the drift risk.

**Recommendation:** Optional and likely NOT worth the churn: a local composite action (`uses: ./.github/actions/harden-runner` with an `allowed-endpoints` input) is the idiomatic way to DRY a shared step across GitHub Actions jobs while keeping it the job's first step. Note the deliberate tradeoff before acting: the existing regression guard asserts on the LITERAL workflow structure (SHA, block, disable-sudo, no if:, harden-runner first), so extracting to a composite action would require reworking that guard to inspect inside the composite. Given the lockstep-bump comment plus the test pin, accepting the duplication is a defensible choice; documented here only for rubric completeness.

```
      - name: Harden runner
        with:
          egress-policy: block
          disable-sudo: true
          allowed-endpoints: >
```

</details>

<details>
<summary><code>QUAL-002</code> · nit · <code>.github/workflows/live-tests.yml:507</code></summary>

**Issue:** The integration-tests harden-runner comment reads 'storage.googleapis.com is NOT needed here (unlike preflight-selftest)'. Read literally, the parenthetical asserts preflight-selftest DOES need storage.googleapis.com. That contradicts the provenance block at line 32 ('storage.googleapis.com was only ever needed by preflight-selftest's GOTOOLCHAIN fetch, which that job's GOTOOLCHAIN: local now removes — so it appears on no list at all') and the preflight-selftest env comment at lines 125-137, both of which establish that preflight-selftest no longer reaches that endpoint. The intended contrast is mechanistic (integration-tests avoids the fetch via setup-go pinning the toolchain; preflight-selftest avoids it via GOTOOLCHAIN: local), but as written the parenthetical is misleading in a PR that is otherwise meticulous about not over-claiming. No security impact — storage.googleapis.com is correctly absent from every list.

**Recommendation:** Reword to make the mechanism explicit and drop the misleading contrast, e.g. 'storage.googleapis.com is NOT needed here: setup-go pins the go.mod toolchain so there is no GOTOOLCHAIN fetch (preflight-selftest reaches the same end via GOTOOLCHAIN: local).'

```
      # GOTOOLCHAIN fetch and storage.googleapis.com is NOT needed here (unlike
      # preflight-selftest). Chrome's egress is deliberately absent: this job drives
```

</details>

</details>

</details>

---

.github/workflows/live-tests.yml
```diff
@@ -371,17 +453,31 @@ jobs:
       # would catch that (LAB-4678 review).
       VESPASIAN_REQUIRE_CHROME: "1"
     steps:
-      # Defense-in-depth (LAB-4732): monitor and record all
-      # outbound network traffic. Audit mode is non-blocking.
+      # Egress allowlist derived from telemetry — see the ALLOWLIST PROVENANCE
+      # block above jobs:. setup-go pins the toolchain from go.mod, so there is no
+      # GOTOOLCHAIN fetch and storage.googleapis.com is NOT needed here (unlike
```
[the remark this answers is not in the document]
santiago-praetorian: Checked this against the workflow at head and it does not hold today — the premise is that `proxy.golang.org` redirects module ZIP downloads to `storage.googleapis.com`, and it does not.

**Measured two ways.**

Directly, against six of this repo's own dependencies — `go-rod/rod`, `google.golang.org/grpc`, `google.golang.org/protobuf`, `alecthomas/kong`, `golang.org/x/net`, `golang.org/x/sync`. Every `@v/<version>.zip` returns `HTTP/2 200` from `proxy.golang.org` itself, with no `Location` header. The sumdb (`sum.golang.org/lookup/...`) and the toolchain list are the same: 200, no redirect.

And from the real thing rather than from a model of it — run `33178235067`, both `integration-tests` and `test`:

```
endpoint called ip address:port 142.250.189.113:443, domain: proxy.golang.org., pid: 3012, process: go
```

That is the Go tool actually dialling out for modules under `block`, and neither job logged a `domain not allowed:` line for `storage.googleapis.com`. If the cache-miss path needed that host, harden-runner would have said so there.

**Why it is not being added anyway.** AC2 of LAB-6015 requires every entry to be derived from harden-runner telemetry of real runs — the allowlists here come from six named green runs on `main`, recorded in the ALLOWLIST PROVENANCE block at the top of the workflow. Adding an endpoint that telemetry has never shown would be exactly the folklore entry that criterion exists to prevent, and it would widen egress on the two jobs with the broadest allowlists already.

Worth saying plainly that this is a point-in-time fact, not a guarantee: if Google moves ZIP delivery back onto a CDN this becomes real, and the failure mode would be a hard CI failure with `domain not allowed: storage.googleapis.com` in the log — loud, not silent, and one line to fix. That is a better trade than pre-emptively widening the list.

Leaving this thread open rather than resolving it, since I am disputing the premise rather than acting on it — happy to be shown a case where the redirect does occur.

test/test-runner-args.sh
```diff
@@ -2046,6 +2046,237 @@ TIMEOUT_STUB
...
+        *)
+            ac3_last=${ac3_raw%% || *}
+            ac3_body=${ac3_raw#* || }
+            # Comments stripped BEFORE matching: a hostname surviving only in a
+            # comment must not satisfy the pin. Every count is `|| true`-guarded
+            # because grep -c exits 1 on zero matches and this file runs set -e,
+            # where a bare assignment would abort the suite mid-section.
+            ac3_code=$(printf '%s\n' "$ac3_body" | grep -vE '^[[:space:]]*#' || true)
+            ac3_unlisted=$(printf '%s\n' "$ac3_code" | grep -cE 'curl .*https://proxy\.golang\.org/' || true)
+            ac3_control=$(printf '%s\n' "$ac3_code" | grep -cE 'curl .*https://github\.com/' || true)
```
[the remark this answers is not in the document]
santiago-praetorian: You were right, and this was the most useful finding on the PR. Thank you.

The grep counted the probe calls but could not read their **sense**. Measured at the time: flipping `-eq 0` to `-ne 0` on the unlisted probe, and dropping the `!` from the control probe, each left all three counts at exactly `1/1/2` and the suite green at `191/0` — while the assertion meant the opposite of what AC3 requires. Under the first inversion, the unlisted host becoming *reachable* would have passed as proof that the egress policy was enforcing. That is the one runtime check that `block` mode actually works, inverted, silently.

**Fixed structurally rather than by widening the grep.** No pattern over free-form shell can read a condition's sense, so `test/test-runner-args.sh` now *executes* `test/assert-egress-enforced.sh` against a stub `curl` and asserts its exit code in three scenarios:

| scenario | stubbed | required |
| -- | -- | -- |
| enforcing | unlisted exits 7, control exits 0 | script exits **0** |
| policy not enforcing | unlisted exits 0 | script exits **1** |
| no egress at all | both exit 7 | script exits **1** |

Both inversions are now CAUGHT, as is no-opping the script body and retyping either pinned host. Those three outcomes run in CI on every PR — you can see them in the `preflight-selftest` job:

```
PASS: AC3 behaviour: a correctly enforcing runner (unlisted refused, control reachable) exits 0
PASS: AC3 behaviour: an unlisted host that is REACHABLE fails the job (the policy is not enforcing)
PASS: AC3 behaviour: a runner with NO egress at all fails the job rather than passing vacuously
```

One residual, stated rather than claimed closed: the stub keys on exit codes, so anything the three scenarios cannot distinguish stays unpinned — log wording, the `--max-time` value, and real-network behaviour. Adding `--insecure` to both probes is a measured MISS. That is recorded in the script's own header and in AGENTS.md.

Resolving, since it is fixed in code and verified in CI.

eli-wald (approved):

Only one issue, strictly cosmetic: there's a repeated sentence in AGENTS.md. Leaving it up to you if you want to delete that. Otherwise LGTM, marking as approved. 

**PR:** https://github.com/praetorian-inc/vespasian/pull/226  
**Linear:** https://linear.app/praetorianlabs/issue/LAB-6015  
**Head SHA:** 9fe97596536ae18313ad9193597c62d466570094  
**Reviewed at:** 2026-08-31T07:48:14.920991-07:00

**Recommendation:** APPROVE

|---|---|---|---|---|---|
| Count | 0 | 0 | 0 | 1 | 0 |

|---|---|---|
| capability-reviewer | 12/12 | 5 |
| test-lead | 9/9 | 1 |

**Vs previous run** (head `d24f2cbf268e`): carried 0 · new 1 · **previously-missed 0** · previously-missed? 0 · dropped 3
- dropped `QUAL-001`: harden-runner step block duplicated verbatim across all five jobs — resolved or no longer flagged this run
- dropped `QUAL-002`: integration-tests comment implies preflight-selftest needs storage.googleapis.com, contradicting provenance block — resolved or no longer flagged this run
- dropped `TEST-001`: No standing test proves block policy actually refuses an unlisted host — resolved or no longer flagged this run

---

<details>

<details>

<details>
<summary><code>QUAL-001</code> · low · <code>AGENTS.md:148</code></summary>

**Issue:** The single-line paragraph at AGENTS.md:148 ends with the same two sentences repeated back-to-back verbatim: '`test/preflight-selftest.sh`, `test/test-runner-args.sh`, `test/install-chrome-selftest.sh`, and `test/setup-live-targets_test.sh` are the CI-run regression guards for the setup preflight, ... . All four run in the un-gated `preflight-selftest` job, which the `test` job depends on so a guard failure stops the full suite before it builds anything.' This is PR-introduced: `grep -o "are the CI-run regression guards for the setup preflight"` returns 1 occurrence on origin/main and 2 on HEAD. The consecutive duplicate is confusing redundancy (no emphasis purpose) and is a doc-rot signal in a file the workflow itself repeatedly points readers to.

**Recommendation:** Delete the second copy of the duplicated sentence pair on AGENTS.md:148, leaving a single occurrence.

```
`test/preflight-selftest.sh`, `test/test-runner-args.sh`, `test/install-chrome-selftest.sh`, and `test/setup-live-targets_test.sh` are the CI-run regression guards for the setup preflight, the target-group/dispatch wiring (plus the un-gated job's own step list and the browser-target classification), the installer's non-privileged surface, and the setup script's teardown / orphan-process hardening (LAB-2893). All four run in the un-gated `preflight-selftest` job, which the `test` job depends on so a guard failure stops the full suite before it builds anything. `test/preflight-selftest.sh`, `test/test-runner-args.sh`, `test/install-chrome-selftest.sh`, and `test/setup-live-targets_test.sh` are the CI-run regression guards for the setup preflight, the target-group/dispatch wiring (plus the un-gated job's own step list and the browser-target classification), the installer's non-privileged surface, and the setup script's teardown / orphan-process hardening (LAB-2893). All four run in the un-gated `preflight-selftest` job, which the `test` job depends on so a guard failure stops the full suite before it builds anything.
```

</details>

</details>

</details>

---

santiago-praetorian:
## Follow-up work is now tracked: LAB-6222, and the AC4/AC6 bookkeeping is closed

Recording the ticket state behind this PR so the deferred scope is visible from the PR rather than only from Linear.

### New ticket — LAB-6222

**[LAB-6222 — Vespasian: flip the three devcontainer live-tests jobs from `egress-policy: audit` to `block`](https://linear.app/praetorianlabs/issue/LAB-6222/vespasian-flip-the-three-devcontainer-live-tests-jobs-from-egress)** (Backlog, under LAB-41).

After merging `main`'s LAB-5766 (#228, merged 2026-08-29), `live-tests.yml` has **10 jobs**, of which **8 carry harden-runner**:

| job | policy | allowlist |
| -- | -- | -- |
| `preflight-selftest` | `block` | 2 endpoints |
| `validator-regression` | `block` | 6 endpoints |
| `docs-check` | `block` | 2 endpoints |
| `integration-tests` | `block` | 7 endpoints |
| `test` | `block` | 8 endpoints |
| `devcontainer-changes` | **`audit`** | — |
| `devcontainer-image` | **`audit`** | — |
| `devcontainer-image-arm64` | **`audit`** | — |

The other two carry no step at all: `check-label` deliberately, and `install-chrome-e2e` because harden-runner does not support container jobs.

**Why this PR does not flip the three devcontainer jobs.** Not an oversight — a sequencing constraint. LAB-6015's AC2 requires each allowlist to be *derived from harden-runner audit telemetry of real runs, with the source runs named so the list is auditable rather than folklore*. This branch's six source runs were captured 2026-08-19 → 2026-08-21; those three jobs landed on `main` on 2026-08-29, **after** that telemetry was read. Allowlisting them now would mean guessing, which AC2 forbids.

That caution is earned rather than theoretical. Three of the endpoint guesses written into LAB-6015's own description did not survive contact with the telemetry: `objects.githubusercontent.com` and `nodejs.org` were never contacted at all (`setup-go`/`setup-node` fetch from `release-assets.githubusercontent.com`), and `storage.googleapis.com` turned out to be needed by no job once `GOTOOLCHAIN: local` was set. It is also the same observe-then-flip ordering LAB-4732 chose deliberately when it landed `audit` first.

These jobs additionally run `docker buildx`, QEMU and the devcontainer CLI, so their egress sets are materially wider than the five flipped here — another reason not to fold them in late.

**The deferral is guarded, not just documented.** All three are pinned as `audit` in `test/test-runner-args.sh` via `EXPECTED_AUDIT_JOBS` plus a per-job policy assertion covering the step's `if:`, `continue-on-error:`, `disable-sudo` and `with:` key set. A silent flip **in either direction** fails the build and names the job. Their workflow comments were rewritten to say why they are on `audit`, replacing the inherited and now-misleading *"flip once telemetry confirms the set"*.

### LAB-6015 acceptance criteria — AC4 and AC6 resolved

**AC4** was unsatisfiable as written (*"a normal run on `main` … across all five jobs"*): `main` today has **0** occurrences of `egress-policy: block` and **8** of `audit`, so no run on `main` can exercise `block` until this merges — a gate that only opens after the thing it gates. It was also stale on the job count. Split into:

- the substance, checked off against run 33280726582 (head `c0f907c`, `event=pull_request`, success, all five block-mode jobs green, no allowlist gaps);
- an explicit unchecked *"re-confirmed on `main` after merge"* line, satisfied by a single green `main` run once this lands.

**AC6** (*"if any job cannot be flipped, the reason is recorded here and that job keeps `audit` explicitly rather than silently"*) is **met**: the reason is recorded, the jobs are explicitly `audit`, and the pin makes a silent change fail. The follow-up work it defers is what LAB-6222 now owns.

### Effect on 13T-146

13T-146's **AC6** — *"Is the live-tests CI workflow hardened against unexpected network egress … or is a documented decision to rely on the existing controls recorded?"* — previously named LAB-6015 as its only gate. It now has two: **LAB-6015** (this PR, five of eight jobs) and **LAB-6222** (the remaining three). Both are listed under that AC in 13T-146's Notes.

### Other follow-ups from this work

| Item | Status |
| -- | -- |
| `ci.yml` still `audit`; its `ci` job delegates to a shared reusable workflow accepting only `build-*` inputs, so no harden-runner step can be injected from this repo | **LAB-6115** (Backlog) |
| `disable-telemetry` decision, and `disable-sudo` → `disable-sudo-and-containers` migration in this same file | **LAB-6137** (Backlog) — worth sequencing with LAB-6222 so the harden-runner blocks are edited once |
| Chrome bypasses the runner's resolver — under `block` it connects direct to `8.8.8.8:53` / `8.8.4.4:53`; harden-runner enforces on resolved domains (StepSecurity's GHSA-46g3-37rh-v698 covers the DoH variant) | **not filed** — a limit on what this control can promise rather than a defect here; nothing is exploited and no crawl assertion depends on it. Filing decision pending |

### Current state of this PR

Head `bfa7d61`. All four workflows green: Live Tests, Verify Pins, CLAUDE.md Drift Detection, Secrets Scan. The `test` job ran the live suite rather than skipping it — **11 targets passed, 0 failed, 0 skipped**.

The last commit fixed a latent false-red inherited from `main`'s own guard step: `printf '%s\n' "$out" | grep -qE …` under `set -o pipefail`, where `grep -q` exits on its first match, `printf` takes SIGPIPE, and `pipefail` promotes that to the pipeline's status — so the leading `!` inverted the verdict. Run 33405732303 failed that way while 40 of the diagnostics it claimed were absent sat in the log it had just printed. Six sites converted to herestrings, plus a mechanical guard so the class cannot return to this workflow; the file already carried a measured comment prescribing the herestring, and prose had not held it.

eli-wald:
# Capability PR Review — PR #226 vs LAB-6015

Ran three rounds of `/capability-pr-review`, commenting with the union of their findings:

**Recommendation: 🔴 REQUEST CHANGES** — 2 blockers (1 medium, 1 low·missing-tests).

PR #226 flips the five flippable `harden-runner` jobs in `live-tests.yml` from
`egress-policy: audit` to `block` with per-job, telemetry-derived allowlists,
bakes an AC3 runtime-enforcement proof into CI, removes the stale audit-era
comments, and pins the three deliberately-audit devcontainer jobs. The security
substance of the change is sound and well-guarded. **AC3, AC6, and AC7 are fully
met; AC1/AC2/AC4 are partial and AC5 is post-merge by construction** (all
non-blocking). The two blockers below are a documentation contradiction the flip
introduced into `AGENTS.md` and a named runtime-coverage gap for the wildcard
jobs — both fixable in-PR.

| | Count |
|---|---|
| Critical | 0 |
| High | 0 |
| Medium | 1 |
| Low | 6 |
| Nit | 2 |
| **Blockers** | **2** |

Dispatched this run: **capability-reviewer (QUAL)** and **test-lead (TEST)** only
— PR #226 touches only `.yml`/`.md`/`.sh` files, so the security agents are not
in scope per the trigger table. See _Reconciliation_ below: a prior run dispatched
`backend-security` off-contract and surfaced additional (dropped) findings worth a
look.

<details>
<summary><b>🔴 Blockers (2)</b></summary>

### QUAL-004 · medium · `AGENTS.md:243` — audit-mode paragraph not reconciled with the block flip (partial AC6)

The added CI paragraph (line 243) correctly states that five of the eight
`harden-runner` jobs now run `egress-policy: block`, but the immediately-preceding
retained paragraph (line 241) still describes **all eight** jobs in the present
tense as `audit` mode that "logs (does not block) outbound traffic," with
`registry.npmjs.org` "part of the anticipated allowlist." Two of the five block
jobs (`validator-regression`, `test`) now **enforce** `block` with
`registry.npmjs.org` as a telemetry-confirmed allowlist entry — so line 241's
"logs (does not block)" and "anticipated allowlist" claims are now false for a
security control. LAB-6015 AC6 explicitly requires the stale audit/flip language
to be removed or updated; the workflow comments were cleaned, but this doc
paragraph is the remaining doc-side instance of exactly that stale class.

```
L241: ...open with a `step-security/harden-runner` step in `egress-policy: audit`
      mode (LAB-4732) — ... logs (does not block) outbound traffic;
      `registry.npmjs.org` is part of the anticipated allowlist.
L243: Of those eight, FIVE run `egress-policy: block` with a per-job
      `allowed-endpoints` list ... LAB-6015 flipped these five.
```

**Fix:** reframe line 241's LAB-4732 sentence as historical origin (e.g. "were
introduced in audit mode under LAB-4732") and drop "logs (does not block)" /
"the anticipated allowlist" for the jobs that now enforce `block`, so the two
adjacent paragraphs no longer contradict each other. This closes the `AGENTS.md`
half of AC6.

_Blocker rationale:_ medium severity, and the only Linear reference is LAB-6015
itself (the ticket under review), not a distinct follow-up tracking issue — so the
follow-up-ticket exemption does not apply.

### TEST-001 · low · `.github/workflows/live-tests.yml:358` — runtime egress proof covers only the literal-allowlist job, not the wildcard jobs (missing tests)

The AC3 runtime-enforcement proof (`test/assert-egress-enforced.sh`, invoked only
at `live-tests.yml:318` inside `preflight-selftest`) demonstrates that `block`
actually **refuses** an unlisted host for exactly one job whose allowlist is the
two-entry, wildcard-free literal set. The three jobs carrying the
`*.blob.core.windows.net:443` wildcard — `validator-regression` (:358),
`integration-tests` (:1157), `test` (:1237) — get only static text/shape pins;
nothing exercises harden-runner's **wildcard matching** at runtime. By the
workflow's own comment the wildcard is "the only entry whose matching is
non-trivial," and `assert-egress-enforced.sh`'s header states that a wildcard
failing open is "exactly what this step exists to catch and exactly what it cannot
see, while every static pin stayed green." The single class of allowlist entry
most likely to fail open therefore has zero runtime coverage.

This is a defense-in-depth gap, **not** an unmet acceptance criterion: AC3 asks
only that a request to an unlisted host be demonstrated to fail, which the
`preflight-selftest` proof satisfies, and both scripts document the limitation
deliberately.

**Fix:** `assert-egress-enforced.sh` notes that dropping the identical step into
`validator-regression` is a verbatim one-liner (that job does not allowlist
`proxy.golang.org`, so the existing `UNLISTED_URL` still refuses there),
extending the runtime proof into a wildcard-carrying job at near-zero cost — add
that step plus its AC3 step/body pins in `test-runner-args.sh`, **or** file a
distinct follow-up ticket for the wildcard runtime gap so it is owned rather than
only commented.

_Blocker rationale:_ TEST finding with `missing_tests: true` and no distinct
follow-up ticket (the finding references LAB-6015's own AC6, not a separate
tracking issue). Filing that follow-up ticket would clear the blocker.

</details>

<details>
<summary>🟡 Non-blockers (7)</summary>

#### Ticket alignment (4)

**REQ-003 · low · AC1 partial — three devcontainer harden-runner steps remain `egress-policy: audit`.**
Five of eight harden-runner steps are flipped to `block`; the three devcontainer
jobs stay `audit`. This is the documented AC7 exemption (they landed after the six
telemetry source runs, so no telemetry-derived allowlist exists yet), is pinned as
`EXPECTED_AUDIT_JOBS` in `test/test-runner-args.sh`, and is tracked by follow-up
**LAB-6222**. Non-blocking: the partiality is the sanctioned scope boundary, not
missing work. _No change required in this PR; confirm LAB-6222 carries the flips._

**REQ-001 · low · AC2 partial — the six telemetry source runs are named in the workflow header, not on LAB-6015.**
The `ALLOWLIST PROVENANCE` header in `live-tests.yml` names all six green main runs
(with head SHAs) and the allowlists are telemetry-derived with no folklore entries;
the PR author concedes the run IDs are not copied onto the ticket. Auditability is
achieved in a durable, pin-referenced location — the gap is only *where the record
lives*. _Copy the six run IDs into a LAB-6015 comment/AC2 description, or record on
the ticket that the workflow header is the authoritative provenance._

**REQ-004 · low · AC4 — green demonstration is pinned to `c0f907c`, four commits behind HEAD (`bfa7d61`).**
The cited run 33280726582 is at `c0f907c`, an ancestor of the head but four commits
back; the intervening commits touched guard logic (a `printf|grep -q` pipefail
verdict-inversion fix), not the five block-job allowlists (unchanged since
`c0f907c`). Coverage is likely intact, but the cited green demonstration no longer
sits at the tip. _Trigger/link a fresh green block-mode run at `bfa7d61` and update
AC4's cited run/SHA; that run also seeds the AC5 post-merge re-confirmation._

**REQ-002 · low · AC5 — unmet by construction (post-merge main re-confirmation cannot exist pre-merge).**
Structurally unsatisfiable while the PR is open; the branch already carries the AC5
guard and stale-comment guard. Expected-open, non-blocking. _After merge, confirm a
single green live-tests run on main under `block` and check AC5 off._

#### Quality (3)

**QUAL-001 · nit · `.github/workflows/live-tests.yml:1042` — rewritten arm64 audit rationale omits the `disable-sudo` reason its sibling records.**
This PR rewrites the `harden-runner` comment block on all three devcontainer jobs
into an explicit AC6 rationale. Two of the three record every deviation from the
standard shape: `devcontainer-image` states at `:794-797` why `disable-sudo` is
deliberately not set, and `devcontainer-changes` keeps `disable-sudo: true` at `:640`.
`devcontainer-image-arm64` also omits `disable-sudo`, but its freshly-rewritten block
(`:1014-1041`) explains only the `audit` policy and is silent on `disable-sudo` — an
unexplained gap in the one job whose own comment calls it "the arch we understand
LEAST." Since the block was rewritten here anyway, this was the moment to close it.
_Add the one-line cross-reference its sibling carries (e.g. "`disable-sudo` is NOT set
here for the same reason as `devcontainer-image` (:794)…") so the deviation is recorded
rather than inferred._ (Carried from the prior off-contract run; verified against source
at `bfa7d615`.)

**QUAL-002 · low · `AGENTS.md:220` — paragraph begins mid-sentence with lowercase "or if".**
The added paragraph at line 220 starts with a lowercase coordinating conjunction
("or if any harden-runner step drifts …") separated by a blank line from line 218,
whose sentence it grammatically continues — so in rendered Markdown it reads as a
dangling fragment. _Merge line 220 back into the preceding paragraph, or restructure
it into its own complete sentence._

**QUAL-003 · nit · `AGENTS.md:243` — file no longer ends with a trailing newline.**
The final added line is emitted without a terminating newline (`\ No newline at end
of file`). _Add a single trailing newline at the end of `AGENTS.md`._

</details>

---

### Rubric coverage

| Agent | Coverage |
|---|---|
| capability-reviewer (QUAL) | 12/12 items (3 findings this run, 6 clean, 3 n/a; QUAL-001 is a 4th carried by hand from the prior run) |
| backend-security (SEC-BE) | not dispatched — no server-side file in the diff |
| frontend-security (SEC-FE) | not dispatched — no frontend file in the diff |
| test-lead (TEST) | 9/9 items (1 finding, 7 clean, 1 n/a) |

### Vs. previous run

Reconciled against a prior run at the **same head SHA** (`bfa7d615`). Because the
prior and current head SHAs are identical, the author-update range is empty, so no
finding can be `NEW`; everything not carried is a coverage difference between runs.

- **Carried:** 1 — QUAL-001 (the arm64 `disable-sudo` rationale gap, nit), reinstated
  by hand from the prior run's snapshot after source re-verification.
- **Previously-missed (line-level):** 3 — QUAL-002, QUAL-003, QUAL-004
- **Previously-missed? (file-level, TEST/REQ):** 5 — TEST-001, REQ-001…004
- **Dropped:** 28 prior findings (QUAL 6, **SEC-BE 9**, TEST 8, REQ 5)

⚠️ **This is a non-convergent pair of runs at an identical SHA — the signature of
agent non-determinism, not code drift.** The prior run was off-contract (it
dispatched `backend-security` on a YAML/shell PR); its 9 **SEC-BE** findings —
including two it rated blockers (a `*.blob.core.windows.net` wildcard egress
concern and a second medium) — are dropped here *by design*, because
`backend-security` is correctly not in scope for this PR. They are not refuted, only
out of lane. The prior run's `test-lead` also surfaced additional TEST blockers
(a duplicated `SUITE_COMPLETED=1` disarming the truncation sentinel; an AC3
`continue-on-error` grep that a `${{ … }}` value defeats). **Treat the union of the
two runs as the finding surface** — before merge, read this run's blockers together
with `.worktrees/capability-pr-review/pr-226/.review/prior/FINDINGS.json`, or
dispatch `backend-security` by hand for a security lens on the workflow/shell
changes.

santiago-praetorian:
Adjudication of the 2026-09-02 capability-pr-review comment (blockers + non-blockers).

**QUAL-004** CONFIRMED + fixed (PROSE) — `AGENTS.md` L241 still described all eight jobs as present-tense `audit` / “logs (does not block)” / “anticipated allowlist” next to L243’s block flip. Reframed L241 as LAB-4732 origin; `--ignore-scripts` wording no longer claims audit-only.

**TEST-001** CONFIRMED + fixed (RUNTIME) — AC3 script is now last on `validator-regression` as well (`run: ./test/assert-egress-enforced.sh`). That job allowlists `*.blob.core.windows.net:443` and does **not** allowlist `proxy.golang.org`, so the existing unlisted-host probe still refuses. `test/test-runner-args.sh` pins both jobs in the existing AC3 counted outcome.

**REQ-003** REFUTED (no violated requirement) — the three remaining `audit` jobs are the documented AC7/AC6 exemption, pinned as `EXPECTED_AUDIT_JOBS`, tracked by LAB-6222.

**REQ-001** CONFIRMED — six telemetry run IDs belong on LAB-6015 as well as in the workflow header. Will copy them onto the ticket (gateway Linear fetch is proxy-blocked from this agent; doing it via API separately).

**REQ-004** CONFIRMED — AC4 citation at `c0f907c` is behind HEAD. A fresh green block-mode run after the TEST-001 push is the update.

**REQ-002** REFUTED (documented decision) — AC5 is unsatisfiable pre-merge by construction.

**QUAL-001** CONFIRMED + fixed (PROSE) — `devcontainer-image-arm64` harden-runner block now records why `disable-sudo` is omitted, citing `devcontainer-image` `:803`.

**QUAL-002** CONFIRMED + fixed (PROSE) — “or if any harden-runner…” joined back onto the preceding paragraph.

**QUAL-003** CONFIRMED + fixed (PROSE) — trailing newline on `AGENTS.md`.

Fixes are in the PR-branch worktree; pushing a commit next.

santiago-praetorian:
Pushed `b475cbc` — TEST-001 (RUNTIME) and QUAL-001/002/003/004 (PROSE) on that commit.

santiago-praetorian:
Adjudication of the capability-pr-review run on `b475cbc` (5 lows, 0 blockers).

**Frame:** skip — this round is comment wording only; no new machinery.

| Finding | Verdict | Mat. | Action |
|---|---|---|---|
| QUAL-001 stale `go 1.25.8` | CONFIRMED + fixed | PROSE | `09d8f17` — comment names `go.mod`'s `go` directive |
| TEST-001 “pinned seven” | CONFIRMED + fixed | PROSE | `09d8f17` — pass uses `${#EXPECTED_ALL_JOBS[@]}`; fail names `EXPECTED_AUDIT_JOBS` |
| REQ AC1 three audit jobs | REFUTED | — | no violated requirement; AC7 + LAB-6222 |
| REQ AC4 cite behind HEAD | CONFIRMED | — | Linear comment: run 33660068719 at `b475cbc` (all five block jobs green). Will refresh again after `09d8f17` CI. |
| REQ AC5 green on main | REFUTED | — | documented post-merge decision; cannot exist pre-merge |

CI on `b475cbc` was already green including `validator-regression`. Did not merge.

eli-wald (approved):

**PR:** https://github.com/praetorian-inc/vespasian/pull/226
**Linear:** https://linear.app/praetorianlabs/issue/LAB-6015
**Head SHA:** 57b292f4f949125397a1be530e7b1f96f736bca8
**Reviewed at:** 2026-09-03

**Recommendation:** APPROVE

|---|---|---|---|---|---|
| Count | 0 | 0 | 0 | 0 | 0 |

|---|---|---|
| capability-reviewer | 12/12 | 3 |
| test-lead | 9/9 | 0 |

**Vs previous run** (head `bfa7d61`): carried 0 · new 0 · **previously-missed 0** · previously-missed? 0 · dropped 9
_rubric delta unavailable (prior run predates per-item accounting)_

All 9 prior findings were low/nit quality-and-doc items or documented-decision `REQ` items, and each has since been resolved by the author (fixes in `live-tests.yml`/`AGENTS.md` within the `bfa7d61..57b292f` range, the six telemetry runs posted to LAB-6015, and the AC4 citation refreshed to the current head). **previously-missed 0** is the positive convergence signal: this run surfaced nothing that was already present in the original commits.

---

_No findings._ The five in-scope `live-tests.yml` jobs are on `egress-policy: block` with telemetry-derived per-job allowlists (2/6/2/7/8 endpoints); the three devcontainer jobs are explicitly `audit` and pinned as the AC7 exemption (tracked by LAB-6222). The `test/test-runner-args.sh` suite pins each job's whole policy by exact value (not shape), and `test/assert-egress-enforced.sh` proves enforcement by executing the AC3 probe with unambiguous exit-code sense. All 7 Linear acceptance criteria are met or documented post-merge deferrals (AC5 / the "on `main`" half of AC4). `AGENTS.md` accurately describes the workflow as changed.

---

.github/workflows/live-tests.yml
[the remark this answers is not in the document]
santiago-praetorian: Refuted — the reviewer's technical premise is factually false. Evidence: `proxy.golang.org` serves module ZIPs itself (no `Location` redirect to `storage.googleapis.com`), recorded in the ALLOWLIST PROVENANCE block at `.github/workflows/live-tests.yml:37-40` and the `integration-tests` job comment at `:1144-1146`. Neither Go job allowlists `storage.googleapis.com:443`; both were green under `block` at `cca24e6` (run 33778021532: `integration-tests` pass 3m58s, `test` pass 4m5s). Adding an endpoint telemetry never showed would violate LAB-6015 AC2.

Agreeing with the 2026-08-30 reply on this thread; independently re-checked against head after the `origin/main` merge (`528d932`).

santiago-praetorian:
## Frame check (reviewing-the-premise) — PROCEED

Trigger: this PR builds non-trivial machinery (per-job harden-runner `block` allowlists + an 800-line mutation pin in `test/test-runner-args.sh` + `test/assert-egress-enforced.sh`). No prior PROCEED/HOLD/WRONG-LAYER was on the record covering the current frame.

**Claims**

| claim | verdict | evidence |
| --- | --- | --- |
| Symptom: five live-tests jobs ran `egress-policy: audit`, which observes but does not restrict egress | CONFIRMED | pre-fix `live-tests.yml` on merge-base `7868c2a`; harden-runner docs: audit ≠ block |
| Root cause: LAB-4732 left audit because crawler egress was non-deterministic; LAB-4999 unblocked the flip | CONFIRMED | PR body + LAB-4999 pin of the system browser; telemetry block names six green main runs |
| Rejected folklore endpoints (`objects.githubusercontent.com`, `nodejs.org`, `storage.googleapis.com`) | CONFIRMED | ALLOWLIST PROVENANCE at `live-tests.yml:34-40`; Go jobs green without `storage.googleapis.com` |
| Chosen layer: CI workflow / harden-runner policy | CONFIRMED | same layer as the symptom (unrestricted CI egress) |
| Mechanism-mandating AC (telemetry-derived per-job lists, AC3 runtime proof, pin against silent broaden) | CONFIRMED | AC2/AC3 in the workflow header and `assert-egress-enforced.sh`; simpler “just flip audit→block” would not catch `use-policy-store` / `container:` / inverted probes (measured survivors cited in `docs/agents/ci.md`) |

**Layer:** symptom and fix are both CI security-control. No layer gap.

**Simpler-layer (Dimension 3):** deleting the pin suite would leave the policy as unenforced prose. The Simplification Test fails — the pin is what makes `block` a control rather than a comment. No extendable existing mechanism covers silent policy broadening.

**PROCEED.** Implementation review continues.

Merge conflicts with `origin/main` (ENG-6805 AGENTS.md relocation) resolved in `528d932` by porting the LAB-6015 prose onto `docs/agents/{ci,architecture,testing}.md`.

santiago-praetorian:
## Adjudication of remaining bot findings (round after `528d932`)

**Codex P1 (`storage.googleapis.com` on Go jobs)** — REFUTED (technical premise false). Replied on the inline thread. `integration-tests` / `test` ran green under `block` without that host.

**Gemini suggestion (`ci.yml` `comment-claims` still `audit`)** — CONFIRMED as a real unrestricted-egress gap, pre-existing (provenance: `.github/workflows/ci.yml` untouched vs merge-base `7868c2a`, empty `diff --numstat`). Same class as LAB-6222 (flip remaining audit jobs once telemetry exists), not authored by this PR. **Not fixed here** — LAB-6015 AC2 forbids guessed allowlists, and `ci.yml` was not in the six source runs. Linear secret backend is unavailable in this session, so the follow-up ticket is **not filed**. Needs a placed sibling of LAB-6222 before this finding can close as CONFIRMED + deferred.

**CodeRabbit** — paused; no new actionable comments on `cca24e6`/`528d932`. Prior inline threads already addressed or resolved.

**capability-pr-review (eli-wald, `57b292f`)** — APPROVE, zero findings. Carried.

Prior 2026-09-02 findings (QUAL-004, TEST-001, QUAL-001/002/003, REQ-*) stay as previously adjudicated; QUAL-004's prose now lives in `docs/agents/ci.md` after the ENG-6805 relocation (same content, new path).

.github/workflows/live-tests.yml
```diff
@@ -26,21 +136,47 @@ jobs:
...
+      # which jobs carry the step at all and the full job set, so none of those can move
+      # without the build failing and naming the job. The three devcontainer jobs
+      # (LAB-5766) still run `audit` and are pinned as audit, deliberately: see the
+      # AC6 rationale recorded at each of those jobs (devcontainer-changes / devcontainer-image /
+      # devcontainer-image-arm64), and pinned via EXPECTED_AUDIT_JOBS.
       - name: Harden runner
         uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920  # v2.20.0
         with:
-          egress-policy: audit
+          egress-policy: block
```
santiago-praetorian, line 175: @codex @gemini @claude review the current head (`528d932`) after the origin/main merge. Vespasian review workflows do not fire on synchronize.

santiago-praetorian:
@coderabbitai review

santiago-praetorian:
## Adjudication of synchronize-round reviews at `528d932`

Vespasian review workflows do not fire on `synchronize` (`on.pull_request.types: [opened, ready_for_review]` only). Re-requested via inline `@codex @gemini @claude` (`3958794670`) and `@coderabbitai review`. All four ran against `528d932`.

| Finding | Class | Verdict | Mat. | Action |
|---|---|---|---|---|
| Codex: add `storage.googleapis.com:443` to Go jobs | implementation | REFUTED | — | no change: technical premise false (ZIP 200 from `proxy.golang.org`; jobs green at this head without the host; AC2 forbids folklore). Same finding as 2026-08-23. |
| Codex: AC3 any-nonzero curl ≠ policy refusal | implementation | REFUTED | — | no change: documented decision. `test/assert-egress-enforced.sh:58-86` records the measured revert of an exit-6 requirement (run 33107912104: block surfaces as curl 7, same as a down host). Contradicts a documented decision. |
| Codex connector on `3958794670` | — | — | — | "Didn't find any major issues" — no finding. |
| Claude: none critical; cache-miss / github.com flake watches | — | REFUTED | — | no change: observations, no violated requirement. |
| Gemini: `--retry` on CONTROL_URL | implementation | REFUTED | — | no change: enhancement/observation, no violated requirement. |
| CodeRabbit: SC2016 at `live-tests.yml:868` | implementation | REFUTED | — | no change: alleged path unreachable. Line is outside this PR's diff vs `origin/main` (empty `rg` on the merge-base patch for `DEVCONTAINER_CLI`/`npx --yes`); no actionlint job in this repo's checks; `devcontainer-image` green at this head. |
| Gemini (2026-08-23): `ci.yml` still `audit` | implementation | CONFIRMED | — | deferred-eligible (path untouched vs merge-base). Follow-up ticket **not filed** — Linear secret backend unavailable this session. Sibling of LAB-6222. **Blocks round-complete.** |

No RUNTIME or PROSE fix this round. Round budget spent: 1 bot re-request (the `@mention` above). No internal reviewer pass.

santiago-praetorian:
Reviewer items collected through: 2026-09-08T14:23:14Z

Newest non-human item is CodeRabbit's reply `5586663389` (updated 2026-09-08T14:23:14Z). Ten-minute quiet window observed after re-collection of inline comments, review bodies, and issue comments.

test/test-runner-args.sh
```diff
@@ -1920,6 +1920,38 @@ else
...
+    # EPIPE -- so it is matched too. Whole-line comments are stripped first, because
+    # the explanatory comment in live-tests.yml necessarily names the bad idiom. The
+    # pattern uses [|] rather than \| because `awk -v` processes escapes in the
+    # value: `\|` would arrive as a bare `|`, turning the regex into an alternation
+    # that matches nearly every line and making this a permanent false FAIL.
+    sigpipe_re='printf.*[|].*(grep([[:space:]]+-[a-zA-Z]+)*[[:space:]]+-[a-zA-Z]*q|head[[:space:]]+-)'
+    sigpipe_hits=$(awk -v r="$sigpipe_re" '
+        { line = $0; sub(/^[[:space:]]+/, "", line)
+          if (line ~ /^#/) next
+          if ($0 ~ r) printf "%d ", NR }' "$WORKFLOW")
```
logan-bayless: **blocker · medium — SIGPIPE guard misses the line-continued form it exists to police**

The `awk` tests each *physical* record, so a pipeline split on a trailing backslash puts writer and reader on separate lines and matches neither half. The regex is also anchored on a literal `printf`, so `echo … | grep -q` and a herestring both evade it — and `live-tests.yml:263` is an in-tree instance that already does.

Control: with a backslash-continued `printf … | grep -q` injected, the **current** guard reports `PASS`. After this change it reports `FAIL`. The one-line and `echo` forms are caught too.

Pairs with the suggestion on `live-tests.yml:263` — land both.

```suggestion
    # Writers are printf, echo AND a herestring: all three feed a reader that can
    # exit early, and all three take the same EPIPE. Anchoring on printf alone let
    # `sed ... <<< "$out" | head -1` stand at live-tests.yml:263, measured.
    sigpipe_re='(printf|echo|<<<).*[|].*(grep([[:space:]]+-[a-zA-Z]+)*[[:space:]]+-[a-zA-Z]*q|head[[:space:]]+-)'
    # Match the LOGICAL line, not the physical one. A pipeline split on a trailing
    # backslash puts the writer and the reader on separate records, so a per-record
    # regex sees neither half and the idiom passes -- measured MISSED.
    sigpipe_hits=$(awk -v r="$sigpipe_re" '
        { if (start == 0) start = NR
          chunk = $0; sub(/[[:space:]]*\\$/, "", chunk); logical = logical chunk
          if ($0 ~ /\\$/) next
          probe = logical; sub(/^[[:space:]]+/, "", probe)
          if (probe !~ /^#/ && logical ~ r) printf "%d ", start
          logical = ""; start = 0 }
        END { if (logical != "") { probe = logical; sub(/^[[:space:]]+/, "", probe)
                if (probe !~ /^#/ && logical ~ r) printf "%d ", start } }' "$WORKFLOW")
```

.github/workflows/live-tests.yml
```diff
@@ -124,8 +260,8 @@ jobs:
           # run showed 21 failures instead of the expected yq ones and this step
           # still exited 0. MEASURED. Compare the failure count against the number
           # of yq-attributed diagnostics.
-          fails=$(printf '%s\n' "$out" | sed -n 's/^  \([0-9]*\) passed, \([0-9]*\) failed.*/\2/p' | head -1)
-          yqfails=$(printf '%s\n' "$out" | grep -cE "^  FAIL:.*($yqpat)" || true)
+          fails=$(sed -n 's/^  \([0-9]*\) passed, \([0-9]*\) failed.*/\2/p' <<< "$out" | head -1)
```
logan-bayless, line 263: **low — keeps the `| head -1` early-exit pipe the same commit removed at `:897`**

This is the instance the widened guard above catches, so it has to land in the same change. Removing the pipe entirely is the fix; the address block makes `q` fire on the first *matching* line, preserving `head -1`'s semantics (a bare `s///p;q` would quit on the first line instead, which is wrong).

Verified functionally identical to the original on BSD `sed` across three inputs: one summary line, two (takes the first), and none (empty). The construct is POSIX; GNU behaviour unverified on my machine.

```suggestion
          fails=$(sed -n '/^  [0-9]* passed, [0-9]* failed/{s/^  \([0-9]*\) passed, \([0-9]*\) failed.*/\2/p;q;}' <<< "$out")
```

test/assert-egress-enforced.sh
```diff
@@ -0,0 +1,102 @@
...
+#     its target by allowlisting it; and both hostname assignments plus both curl calls
+#     in THIS file are pinned by the AC3 body check in that same suite, so retyping
+#     UNLISTED_URL or deleting a probe fails the build. What remains unguarded is the
+#     rest of this script's body, which no pin covers.
+set -euo pipefail
+
+# Pinned in test/test-runner-args.sh. proxy.golang.org must stay OFF
+# preflight-selftest's allowlist (the endpoints pin enforces that) and must stay a
+# real, resolvable host; github.com must stay the allowlisted control.
+UNLISTED_URL="https://proxy.golang.org/"
```
logan-bayless, line 55: **blocker · medium — runtime enforcement proof absent from the two widest-allowlist jobs (1 of 3)**

The proof runs only in `preflight-selftest` (2 endpoints) and `validator-regression` (6). `integration-tests` (7) and `test` (8) — the widest allowlists and the only jobs that execute third-party code — get static YAML pins only. Both allowlist `proxy.golang.org`, so the default probe host would be *reachable* there and the assertion would pass vacuously.

Taking the host as a positional argument means the existing `run=` pin covers it for free: no new assertion, no `EXPECTED_ASSERTIONS` change.

```suggestion
# $1 overrides the unlisted host, for jobs that DO allowlist the default. The
# host must be absent from the allowlist of the job running this, or the probe
# passes vacuously; test/test-runner-args.sh pins each job's exact invocation,
# so the override is covered by the same pin as the rest of the step.
UNLISTED_URL="${1:-https://proxy.golang.org/}"
```

test/test-runner-args.sh
```diff
@@ -2632,6 +2671,533 @@ TIMEOUT_STUB
...
+    # AGENTS.md records the same reasoning for test/assert-chrome-install.sh.
+    ac3_got=$(yq_query '"preflight=" + ("last=" + (((.jobs."preflight-selftest".steps[-1].name) // "") == "Assert egress policy enforces (AC3)" | tostring)
+      + " shell=" + ([.jobs."preflight-selftest".steps[] | select(.name == "Assert egress policy enforces (AC3)") | (.shell // "<default>")] | join(""))
+      + " run=" + (([.jobs."preflight-selftest".steps[] | select(.name == "Assert egress policy enforces (AC3)") | .run] | join("")) | sub("\s+$"; "")))
+      + " validator=" + ("last=" + (((.jobs."validator-regression".steps[-1].name) // "") == "Assert egress policy enforces (AC3)" | tostring)
+      + " shell=" + ([.jobs."validator-regression".steps[] | select(.name == "Assert egress policy enforces (AC3)") | (.shell // "<default>")] | join(""))
+      + " run=" + (([.jobs."validator-regression".steps[] | select(.name == "Assert egress policy enforces (AC3)") | .run] | join("")) | sub("\s+$"; "")))' -r)
+    ac3_want='preflight=last=true shell=<default> run=./test/assert-egress-enforced.sh validator=last=true shell=<default> run=./test/assert-egress-enforced.sh'
```
logan-bayless, line 2975: **blocker · medium — extend the AC3 pin to all four jobs (2 of 3)**

Without this the two new steps are unpinned coverage. Note each prefix string after the first needs its **leading space** — omitting it concatenates the fields and the pin fails (I hit exactly that).

Controls, all caught: repointing a new probe at an allowlisted host; deleting the step from `test`; moving it off the end of `integration-tests`.

```suggestion
    ac3_got=$(yq_query '"preflight=" + ("last=" + (((.jobs."preflight-selftest".steps[-1].name) // "") == "Assert egress policy enforces (AC3)" | tostring)
      + " shell=" + ([.jobs."preflight-selftest".steps[] | select(.name == "Assert egress policy enforces (AC3)") | (.shell // "<default>")] | join(""))
      + " run=" + (([.jobs."preflight-selftest".steps[] | select(.name == "Assert egress policy enforces (AC3)") | .run] | join("")) | sub("\s+$"; "")))
      + " validator=" + ("last=" + (((.jobs."validator-regression".steps[-1].name) // "") == "Assert egress policy enforces (AC3)" | tostring)
      + " shell=" + ([.jobs."validator-regression".steps[] | select(.name == "Assert egress policy enforces (AC3)") | (.shell // "<default>")] | join(""))
      + " run=" + (([.jobs."validator-regression".steps[] | select(.name == "Assert egress policy enforces (AC3)") | .run] | join("")) | sub("\s+$"; "")))
      + " integration=" + ("last=" + (((.jobs."integration-tests".steps[-1].name) // "") == "Assert egress policy enforces (AC3)" | tostring)
      + " shell=" + ([.jobs."integration-tests".steps[] | select(.name == "Assert egress policy enforces (AC3)") | (.shell // "<default>")] | join(""))
      + " run=" + (([.jobs."integration-tests".steps[] | select(.name == "Assert egress policy enforces (AC3)") | .run] | join("")) | sub("\s+$"; "")))
      + " test=" + ("last=" + (((.jobs."test".steps[-1].name) // "") == "Assert egress policy enforces (AC3)" | tostring)
      + " shell=" + ([.jobs."test".steps[] | select(.name == "Assert egress policy enforces (AC3)") | (.shell // "<default>")] | join(""))
      + " run=" + (([.jobs."test".steps[] | select(.name == "Assert egress policy enforces (AC3)") | .run] | join("")) | sub("\s+$"; "")))' -r)
    ac3_want='preflight=last=true shell=<default> run=./test/assert-egress-enforced.sh validator=last=true shell=<default> run=./test/assert-egress-enforced.sh integration=last=true shell=<default> run=./test/assert-egress-enforced.sh https://example.com/ test=last=true shell=<default> run=./test/assert-egress-enforced.sh https://example.com/'
```

logan-bayless (commented):
## capability-pr-review — 2 blockers

Both blockers are in this PR's own new safety machinery, not the egress flip. The flip itself came back clean: policy correct on all five jobs, endpoint counts 2/6/2/7/8 verified, all five action SHAs resolve to exactly their claimed tags, AC6 met and guarded.

**The two suggestions on `test-runner-args.sh:1945` and `live-tests.yml:263` must land together.** Widening the SIGPIPE guard makes it catch the surviving `sed … <<< "$out" | head -1` at `:263`; applying the guard alone turns CI red. Proven, not assumed — reverting `:263` with the new guard in place produces a FAIL.

Everything below was applied to a scratch tree and the suite run: baseline `246 passed, 1 failed` → `247 passed, 1 failed` (+1 = the one new assertion), with seven negative controls all caught. Caveat: the suite cannot complete on macOS because `test-runner-args.sh:3683` uses a GNU-only `sed` construct (pre-existing, present in `origin/main`, out of scope here) — I neutralised it identically in both trees, so the differential holds but a faithful number needs Linux CI. Not verified: that harden-runner actually refuses the two new probes in CI. That needs a push.

test/test-runner-args.sh
```diff
@@ -2632,6 +2671,533 @@ TIMEOUT_STUB
...
+        ac3_c=$(printf '%s\n' "$ac3_src" | grep -cE 'curl .*"\$CONTROL_URL"' || true)
+        ac3_h=$(printf '%s\n' "$ac3_src" | grep -cE '^(UNLISTED_URL="https://proxy\.golang\.org/"|CONTROL_URL="https://github\.com/")$' || true)
+        [[ "$ac3_p" == 1 && "$ac3_c" == 1 && "$ac3_h" == 2 ]] && ac3_body=yes
+    fi
+    case "$ac3_got" in
+        __NO_YQ__)    fail_no_yq "the AC3 egress-enforcement step" ;;
+        __YQ_ERROR__) fail_yq_error "the AC3 egress-enforcement step" ;;
+        *)
+            if [[ "$ac3_got" == "$ac3_want" && "$ac3_body" == "yes" ]]; then
+                pass "AC3 enforcement step is last in preflight-selftest and validator-regression, invokes test/assert-egress-enforced.sh, and that script still makes both probe calls against both pinned hosts"
```
logan-bayless, line 2997: **blocker · medium — the pin's own message still names only two jobs (3 of 3)**

```suggestion
                pass "AC3 enforcement step is last in all four jobs that carry it (preflight-selftest, validator-regression, integration-tests, test), invokes test/assert-egress-enforced.sh, and that script still makes both probe calls against both pinned hosts"
```

.github/workflows/live-tests.yml
```diff
@@ -992,11 +1274,12 @@ jobs:
...
-      # audit-only egress policy.
+      # Deps are hermetic once installed; the only egress is registry.npmjs.org,
+      # which is on this job's harden-runner allowlist above — telemetry-confirmed,
+      # not anticipated.
+      # --ignore-scripts prevents install/preinstall/postinstall lifecycle
+      # scripts from running at all — see the same note in validator-regression
+      # for why `block` does not make it redundant.
       - name: Install spec validators
         working-directory: test/spec-validators
         run: npm ci --ignore-scripts
```
logan-bayless, line 1285: **blocker · medium — the two new AC3 steps (not applyable as a suggestion)**

GitHub can only anchor a suggestion to a line inside a diff hunk, and both insertion points here are unchanged context ~25 lines outside one, so this is a plain comment rather than an applyable block. Add this step in **two** places:

1. `integration-tests`, after `run: make test-integration` (`:1202`)
2. `test`, after the `Teardown` step (`:1316`) — last, so teardown has already run

`example.com` is not a guess: it is absent from both jobs' allowlists, and this PR's own blocked-host table already records harden-runner refusing it in `integration-tests` under audit.

```yaml
      # RUNS LAST for the same reason it does in preflight-selftest. This job
      # allowlists proxy.golang.org, so the default probe host would be REACHABLE
      # here and the assertion would pass vacuously; example.com is absent from
      # this job's allowlist and is a host harden-runner was already observed
      # refusing here in audit mode. Its exact invocation is pinned in
      # test/test-runner-args.sh, so repointing it at an allowlisted host fails CI.
      - name: Assert egress policy enforces (AC3)
        run: ./test/assert-egress-enforced.sh https://example.com/
```

test/test-runner-args.sh
santiago-praetorian: Confirmed (RUNTIME) — awk matched each physical record, so a `printf … | grep -q` split on a trailing `\` matched neither half and the suite stayed green. Reproduced with an injected continuation: pre-fix hits empty, post-fix FAIL. Fixed in `3d1bd80` by joining logical lines before the existing printf/grep-q/head regex. Did not widen writers to echo/herestring — that is a different contract than the measured printf class.

.github/workflows/live-tests.yml
santiago-praetorian: Refuted — enhancement with no violated requirement. `sed … <<< "$out" | head -1` is not the printf-into-early-exit inversion class: no `!`/`||` to invert a match into a green check, and a non-zero here fail-closes the step. Small writes fit the pipe buffer (local: one/two summary lines rc=0). The SIGPIPE fix in `3d1bd80` deliberately does not catch this line; landing the widened guard without this change would only create a false FAIL, which is why the two suggestions were not applied together.

test/assert-egress-enforced.sh
santiago-praetorian: Refuted — contradicts a documented decision. AC3 is a sampling proof, not coverage of every block job: `test/assert-egress-enforced.sh:73-81` states two jobs (literal allowlist + one wildcard), and `UNLISTED_URL="${1:-…}"` would break the body pin that requires `UNLISTED_URL="https://proxy.golang.org/"` (`test-runner-args.sh:2969-2970`). Dropping the current script unchanged onto `integration-tests`/`test` would FAIL (curl to an allowlisted host exits 0 → "policy not enforcing"), not pass vacuously. Expanding to those jobs is a valid enhancement, not a violated requirement.

test/test-runner-args.sh
santiago-praetorian: Refuted — enhancement with no violated requirement. The pin at `test-runner-args.sh:2949-2955` matches the two AC3 steps that exist. Extending `ac3_want` without those steps fails the suite; it is the companion of the scope expansion, not a standalone defect.

test/test-runner-args.sh
santiago-praetorian: Refuted — the code already handles the case. The pass message names preflight-selftest and validator-regression, which is exactly `ac3_want` and the two workflow steps.

.github/workflows/live-tests.yml
santiago-praetorian: Refuted — enhancement with no violated requirement. `example.com` is a sound unlisted host for those jobs (absent from both allowlists; listed under deliberately-not-allowed) but that makes it usable for an expansion, not a missing AC3. Runtime fail-open of harden-runner is already discriminated on a literal job and a `*.blob` wildcard job (`assert-egress-enforced.sh:73-81`). YAML already pins `block` and the exact endpoint set on `integration-tests`/`test`.

test/test-runner-args.sh
```diff
@@ -1920,6 +1920,56 @@ else
...
+    # pattern uses [|] rather than \| because `awk -v` processes escapes in the
+    # value: `\|` would arrive as a bare `|`, turning the regex into an alternation
+    # that matches nearly every line and making this a permanent false FAIL.
+    # Backslash-continued physical lines are joined into one logical line before
+    # the match: a `printf … | grep -q` (or `| head -N`) split on a trailing `\`
+    # puts writer and reader on separate records, so neither half matches, hits
+    # stay empty, and the inversion class is back.
+    sigpipe_re='printf.*[|].*(grep([[:space:]]+-[a-zA-Z]+)*[[:space:]]+-[a-zA-Z]*q|head[[:space:]]+-)'
+    sigpipe_hits=$(awk -v r="$sigpipe_re" '
+        {
```
santiago-praetorian, line 1952: @codex @gemini @claude review the current head after the SIGPIPE logical-line fix (`3d1bd80`).

test/test-runner-args.sh
```diff
@@ -1920,6 +1920,56 @@ else
...
+          if (cont) { logical = logical $0 }
+          else      { logical = $0; start = NR }
+          if ($0 ~ /\\$/) { sub(/\\$/, "", logical); cont = 1; next }
```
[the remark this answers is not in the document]
santiago-praetorian: Confirmed (RUNTIME) — the joiner only set `cont` on trailing `\`; Bash also continues a pipeline when the physical line ends in `|`. Reproduced: `printf … |` / `grep -q` → hits empty before, nonempty after. Fixed in `9561b27` (keep the `|`; still strip `\`; writers still printf-only).

test/test-runner-args.sh
```diff
@@ -1920,6 +1920,58 @@ else
...
+    # logical line before the match. The `|` is kept; a trailing `\` is stripped.
+    # A `printf … | grep -q` (or `| head -N`) split across records otherwise
+    # matches neither half, hits stay empty, and the inversion class is back.
+    sigpipe_re='printf.*[|].*(grep([[:space:]]+-[a-zA-Z]+)*[[:space:]]+-[a-zA-Z]*q|head[[:space:]]+-)'
+    sigpipe_hits=$(awk -v r="$sigpipe_re" '
+        {
+          if (cont) { logical = logical $0 }
+          else      { logical = $0; start = NR }
+          if ($0 ~ /\\$/) { sub(/\\$/, "", logical); cont = 1; next }
+          if ($0 ~ /[|][[:space:]]*$/) { cont = 1; next }
```
santiago-praetorian, line 1956: @codex @gemini @claude review the current head after the pipe-at-EOL SIGPIPE join (`9561b27`).

santiago-praetorian:
Adjudication of [Codex review 5147206955](https://github.com/praetorian-inc/vespasian/pull/226#pullrequestreview-5147206955) at `9561b27`.

**Critical — add `nodejs.org:443` to `validator-regression` (`:352`) and `test` (`:1247`)** — REFUTED (technical premise false / contradicts documented decision). Evidence: ALLOWLIST PROVENANCE at `live-tests.yml:34-36` records that `nodejs.org` is never contacted; `setup-node` fetches from `release-assets.githubusercontent.com` (already on both lists). The same six telemetry runs that built these allowlists did not show `endpoint called` for `nodejs.org`. Adding it would be the folklore entry AC2 exists to forbid. A future cache-miss fallback that really needed it would fail loud (`domain not allowed: nodejs.org`), not silently — same trade as `storage.googleapis.com`. Both jobs are green under `block` at this head.

**Security — `*.blob.core.windows.net:443` on `integration-tests` (`:1175`) and `test` (`:1255`)** — REFUTED (contradicts a documented decision). Evidence: `live-tests.yml:77-89` already names this as a known cost, not an oversight. harden-runner injects only the job's own results account; Actions cache / upload-artifact used *other* accounts in the same run (productionresultssa2 and sa14 besides the injected sa13). No tighter named-account form held. The exact-set pin in `test/test-runner-args.sh` would fail if the wildcard were replaced with a named account. Removing cache/artifact from those jobs is new scope, not a violated requirement of LAB-6015.

**Suggestion — AC3 any-nonzero curl as proof of block** — REFUTED (contradicts a documented decision). Same finding as the 2026-09-08 Codex review. `test/assert-egress-enforced.sh:58-86` records the measured revert of an exit-6 requirement (run 33107912104: a correct block surfaces as curl 7, indistinguishable from a down host). No change.

logan-bayless (approved):

PR #227, opened 2026-08-23
SCRATCH — do not merge: demonstrate LAB-6015 block policy actually blocks (AC3)

santiago-praetorian:
**Scratch PR for LAB-6015 AC3. Not for merge — closed once the run is recorded.**

Stacked on #226. Adds two steps to the `docs-check` job:

1. **Control** — `curl https://github.com/`, which IS on `docs-check`'s allowlist. Must succeed.
2. **Demonstration** — `curl https://proxy.golang.org/`, which is on `preflight-selftest`, `integration-tests` and `test`'s allowlists but deliberately NOT on `docs-check`'s. Must fail the job.

The control is what makes the demonstration mean something: if both failed, this would prove a broken runner rather than a working policy. The choice of `proxy.golang.org` also demonstrates the lists are scoped **per job**, not workflow-wide — the same host is reachable from three sibling jobs in the same run.


santiago-praetorian:
## Demonstration recorded — AC3 satisfied

Run [32651913938](https://github.com/praetorian-inc/vespasian/actions/runs/32651913938), job `docs-check` ([97224746586](https://github.com/praetorian-inc/vespasian/actions/runs/32651913938/job/97224746586)) — **failed, as designed**.

| step | result |
| -- | -- |
| Control — an ALLOWED host is reachable (`curl https://github.com/`) | ✅ success |
| Demonstration — an UNLISTED host must fail the job (`curl https://proxy.golang.org/`) | ❌ **failure** |

Step output:

```
curl: (7) Failed to connect to proxy.golang.org port 443 after 1 ms: Couldn't connect to server
```

harden-runner post-step:

```
Sun, 23 Aug 2026 16:31:29 GMT:domain not allowed: proxy.golang.org.
```

Three things this pins that a bare "the job failed" would not:

1. **The gate is live, not decorative.** 1 ms to failure is the sinkhole, not a timeout.
2. **The failure is the policy, not the network.** The control request to `github.com` succeeded in the same job, seconds earlier.
3. **The lists are per job, not workflow-wide.** `proxy.golang.org` is on `preflight-selftest`, `integration-tests` and `test`'s allowlists — all three reached it successfully in this same run — and is deliberately absent from `docs-check`'s.

The other four jobs in this run were green, which is the same result #226's own run produces without these two scratch steps.

Closing unmerged. The evidence lives here and on LAB-6015.

PR #228, opened 2026-08-23
test: wire install-chrome.sh into a devcontainer image build (LAB-5766)

santiago-praetorian:
Closes AC1 of [LAB-5064](https://linear.app/praetorianlabs/issue/LAB-5064) — the half PR #197 could not deliver. Fixes [LAB-5766](https://linear.app/praetorianlabs/issue/LAB-5766).

## The repo-boundary problem dissolves

LAB-5766 records the blocker as "the devcontainer definition lives **outside** this repo, so the wiring is not a change PR #197 could make," and step 1 as "identify the image/repo that owns it."

**No repo owns one.** Code-searching `praetorian-inc` returns 62 `devcontainer.json` files and none for vespasian; `praetorian-dev-platform` and `praetorian-development-platform` both point at `chariot-devpod` and neither lists vespasian in `.gitmodules`. So step 1 resolves to "none exists", and the boundary disappears once this repo owns its own definition — which is what this PR does.

## What lands

| File | |
| --- | --- |
| `.devcontainer/Dockerfile` | digest-pinned `devcontainers/go:1.25-bookworm`, adds `python3`, runs the installer as an image layer |
| `.devcontainer/devcontainer.json` | repo-root build context, `VESPASIAN_NO_SANDBOX`, `--shm-size=1g`, `remoteUser` |
| `devcontainer-image` CI job | builds the image and asserts the browser it ships |
| `test/README.md`, `AGENTS.md`, `CONTRIBUTING.md` | the "manual step" caveats go |

### Three details are load-bearing

1. **`"context": ".."`.** The layer copies out of `test/`; the default context (the `.devcontainer` directory) cannot see it and the build fails.
2. **Both `install-chrome.sh` *and* `common.sh`** into one directory. The installer sources `common.sh` from its own `dirname`, so copying it alone dies at `source` — the trap `test/README.md:50-53` already documented.
3. **The `RUN` names `container=docker`.** `in_container()` gates the removal **and** the verify-absent pass over the package's phone-home artifacts, and none of the markers it reads is a dependable *build-time* signal: `/.dockerenv` is placed by the daemon for `docker run`, `REMOTE_CONTAINERS` is a VS Code runtime variable. Without it the build suppresses those artifacts without ever confirming their absence, and "no new background egress" becomes an assumption. With it, `install-chrome.sh` itself iterates `PHONE_HOME_PATHS` and exits non-zero on a survivor — so a clean build **is** that assertion, which is why the new job carries no third copy of the path list.

## Why a CI job

This ticket exists because the last half of LAB-5064 was merged and forgotten. Nothing in this repo builds a Dockerfile, so the COPY paths, the build context and the `container=docker` gate would rot invisibly. `devcontainer-image` reuses the committed `test/assert-chrome-install.sh` rather than restating it, so it cannot disagree with `install-chrome-e2e` about what "a runnable browser" means. Gated on the same two arms as that job, for the same reasons.

The `LookPath` step greps for `--- PASS` instead of trusting the exit code: `TestConfigureLauncher_PinsSystemBrowser` **skips** when `LookPath` finds nothing and `go test` exits 0 on a skip, so the one assertion AC1 is actually worded around could vanish and leave the job green. The other step cannot cover it — it resolves the browser through `common.sh`'s `CHROME_CANDIDATES`, a different list from go-rod's.

## Verified, not assumed

- **arm64 is not a dead end.** Google's apt repo advertises `Architectures: amd64 arm64` and ships `google-chrome-stable 151.0.7922.173-1` for arm64; the base image index carries both platforms. An Apple Silicon build works rather than failing the whole image.
- **Base image fits.** `GOLANG_VERSION=1.25.12` ≥ go.mod's `1.25.8` — which matters because `GOTOOLCHAIN=local` means Go will not fetch a newer toolchain. Builds as root, so the installer resolves `SUDO=""` exactly as `install-chrome-e2e` does. `ca-certificates`/`curl`/`gnupg` present; `coreutils`/`util-linux` are Priority: required. Node baked in.
- **`python3` was the one real gap** — the baked-in features are `common-utils, git, go, node` only, and `setup-live-targets.sh`'s preflight fails outright without it.
- Guards all green: `test-runner-args.sh` 135/0, `preflight-selftest.sh` 91/0, `install-chrome-selftest.sh` 275/0, `setup-live-targets_test.sh` 87/0, `check-docs` ok, `go vet` + `golangci-lint` 0 issues, `make lint-comments` clean, full `go test ./...` and `go test -tags integration ./pkg/crawl` pass.

## Please look at this one thing

`VESPASIAN_NO_SANDBOX=true` in `containerEnv` disables Chrome's OS sandbox for every crawl in this container, and the crawler's job is visiting third-party targets. Setting it is **not** what creates the exposure — go-rod already passes `--no-sandbox` whenever it detects a container (LAB-4994), so the sandbox was already off in any containerised run, and the `pkg/crawl` integration tests pass with it unset. What the line changes is that the decision becomes vespasian's own and greppable through `vespasianEnablesNoSandbox` instead of an implicit library side effect. The trade-off is now stated in the file rather than only the mechanism, but it is worth a conscious sign-off rather than my judgement alone.

## What I could not verify

**The image was never built or booted.** No container runtime exists in the authoring environment — `docker`, `podman`, `devcontainer`, `nerdctl` and `buildah` are all absent and there is no docker socket. Everything above is registry inspection, guard suites, and the assertion script run against a real Chrome; **AC1 itself rests on the new CI job (it runs on push to `main`, or on `workflow_dispatch`) or on a reviewer rebuilding the container.** Saying so explicitly rather than implying the criterion is closed.

Worth noting the AC2/AC3 stakes, since "browserless" understates them: without a runnable Chrome, **16 integration tests hard-fail** (12 in `engine_integration_test.go` via `launchTestBrowser`, 4 in `browser_integration_test.go`) and four live targets fail with no `chrome_available` guard at all — `scan-rest`, `concat-spa`, `concat-spa-two-stage`, `crawl-depth`. Only `no-download` and the dual-backend targets skip cleanly.

## Beyond the ticket's file list

- `CONTRIBUTING.md` carried the same manual-step claim in a third, contributor-facing place, and its project-layout tree had no `.devcontainer/` entry. Leaving them would make the repo contradict itself.
- The `harden-runner` lockstep comment (5 copies) and `AGENTS.md`'s job counts move five→six and seven→eight. A new non-container job without `harden-runner` would be an inconsistent security control, and a comment saying FIVE beside six copies is exactly the stale-comment defect `AGENTS.md` exists to prevent.


santiago-praetorian:
Thanks — going through the three bot reviews.

## Codex's critical issue is a false positive, and it found a stale comment doing it

> `.devcontainer/Dockerfile:18` claims Linux/arm64 support … but the repository itself documents that Google Chrome has no arm64 Linux build (`pkg/crawl/browser.go:64`). Apple Silicon builds will therefore fail during the installer layer.

Google publishes `google-chrome-stable` for arm64. Checked against the live repo, not from memory:

```console
$ curl -s .../deb/dists/stable/Release | grep ^Architectures:
Architectures: amd64 arm64

$ curl -s .../deb/dists/stable/main/binary-arm64/Packages | awk '/^Package: google-chrome-stable$/,/^$/'
Package: google-chrome-stable
Version: 151.0.7922.173-1
Architecture: arm64
Filename: pool/main/g/google-chrome-stable/google-chrome-stable_151.0.7922.173-1_arm64.deb
Size: 132245836

$ curl -sI .../deb/pool/main/g/google-chrome-stable/google-chrome-stable_151.0.7922.173-1_arm64.deb
HTTP 200 · content-type: application/x-debian-package · content-length: 132245836
```

The `.deb` exists and is downloadable, so an Apple Silicon build resolves and installs rather than failing.

The repo is self-contradictory here, and Codex picked the wrong side of it: `test/install-chrome.sh:249` accepts `amd64|arm64`, `:258` says "amd64 and arm64 only", and `test/README.md:41` says "(`.deb`, amd64 or arm64)". The outlier is **`pkg/crawl/browser.go:64`**, which still says "arm64 Linux, where Google Chrome has no build" — true when it was written, not true now. That's a genuinely useful find even though the conclusion drawn from it was wrong: a stale comment that reads as evidence, which is the exact defect class `AGENTS.md`'s "comments that claim a state is impossible must cite a test" section exists for. I have deliberately **not** fixed it here — it is a Go file unrelated to this change and editing it pulls `ci.yml` into the PR — so flagging it for a one-line follow-up rather than widening scope unasked.

## The two suggestions, both fair, neither taken here

**Gemini — run the assertions as `vscode` rather than root.** Agreed on the principle: `docker run` uses the image default (root) and so bypasses `remoteUser`, meaning CI does not exercise the permissions a developer actually has. Not applied yet because doing it correctly needs `-u vscode` **and** `-e HOME=/home/vscode` — `docker run -u` does not update `HOME`, so Chrome's `~/.config` write and Go's `GOCACHE` would both land somewhere unwritable and the failure would look like an image defect rather than a harness one. I could not test that here (see the caveat in the PR description: no container runtime in the authoring environment), and I would rather not ship an unverifiable change to the one job that has to be trustworthy. Worth doing as a follow-up by someone who can run it once.

**Codex — run the build on PRs touching `.devcontainer/**` or the installer.** Also fair, and it is the honest weakness of the current gating: nothing checks the image before merge. GitHub Actions has no per-job `paths` filter, so this needs either a `dorny/paths-filter` step or a separate workflow with its own `on.pull_request.paths` — more machinery than the gating decision that was actually made, which deliberately mirrors `install-chrome-e2e`. Leaving as a maintainer call rather than expanding scope.

## Still open for a human

The PR description says this and it has not changed: **the image was never built or booted** — no container runtime exists in the authoring environment — and `devcontainer-image` is gated off PRs, so it shows `skipping` above. AC1 has no automated proof on this PR. Either dispatch the workflow on this branch or rebuild the container once before merging.

eli-wald (approved):

Approving this PR. Maybe consider reviewing REQ-001 (making sure the devcontainer-image job passed) before merging, if that wasn't done during development already. Remaining notes are lows/nits that don't need to be dealt with here. 

**PR:** https://github.com/praetorian-inc/vespasian/pull/228  
**Linear:** https://linear.app/praetorianlabs/issue/LAB-5766  
**Head SHA:** 6e181b5a25fc05acadc82f8658996f55ef15ce3d  
**Reviewed at:** 2026-08-28T21:46:54.316663+00:00

**Recommendation:** APPROVE

|---|---|---|---|---|---|
| Count | 0 | 0 | 0 | 2 | 4 |

|---|---|---|
| capability-reviewer | 12/12 | 2 |
| test-lead | 9/9 | 0 |

**Vs previous run:** first run for this PR

---

<details>

<details>
<summary>Ticket alignment (4)</summary>

<details>
<summary><code>REQ-001</code> · low · <code>.devcontainer/Dockerfile:213</code></summary>

**Acceptance criterion:** AC1  
**Status:** PARTIAL

**Evidence:** .devcontainer/Dockerfile:213-214 runs test/install-chrome.sh as an image layer (COPY at :188 brings install-chrome.sh + common.sh into one dir so the `source` succeeds); .devcontainer/devcontainer.json:14-17 sets build context ".." so the COPY out of test/ resolves, and :70 wires on-create.sh. The installer provisions a real non-snap Chrome from Google's apt repo (install-chrome.sh header :28-33). Verification exists and runs on THIS PR: the devcontainer-image job (live-tests.yml:544-786) builds through devcontainer.json and asserts a runnable browser, and its pull_request arm fires because devcontainer-changes (live-tests.yml:411-543) matches the changed .devcontainer/** paths. So the wiring is complete and no manual step remains.

**Gap:** The image was never built or booted in the authoring environment (PR body: 'The image was never built or booted' — no docker/podman/devcontainer runtime). AC1's proof therefore rests entirely on the devcontainer-image CI job succeeding; its result is not observable in this review. Residual risk is low because that job is gated to run on this PR (not only on push-to-main), but 'a fresh devcontainer comes up' has not been demonstrated by an actual build.

**Recommendation:** Confirm the devcontainer-image job (and devcontainer-image-arm64 build leg) passed on this PR before merge; treat those check runs as the AC1 acceptance evidence. Optionally have a reviewer 'Reopen in Container' once to boot it manually.

</details>

<details>
<summary><code>REQ-002</code> · nit · <code>.devcontainer/Dockerfile:213</code></summary>

**Acceptance criterion:** AC4  
**Status:** PARTIAL

**Evidence:** .devcontainer/Dockerfile:213 runs the installer with `container=docker`, which forces install-chrome.sh's in_container() branch (install-chrome.sh:733, :852) to both REMOVE and VERIFY-ABSENT the phone-home artifacts (apt source, keyring, /etc/cron.daily update pinger — PHONE_HOME_PATHS at :231, removal at :718). The installer emits 'No Google apt source, keyring, or update pinger left behind' (install-chrome.sh:1378) only on the verify-absent pass, and exits non-zero on a survivor. live-tests.yml:618-620 greps the build log for that exact marker and fails the build if the in_container branch never ran. So a clean build is itself the AC4 assertion.

**Gap:** The removal+verify-absent pass and its log-marker check only execute during the devcontainer-image CI build, which was never run locally. Sound assertion, deferred verification.

**Recommendation:** Treat a passing devcontainer-image build on this PR (which greps the in_container marker) as AC4 evidence; ensure the job is required for merge.

</details>

<details>
<summary><code>REQ-003</code> · nit · <code>.github/workflows/live-tests.yml:685</code></summary>

**Acceptance criterion:** AC2  
**Status:** PARTIAL

**Evidence:** test/assert-devcontainer-lookpath.sh:26-47 runs the exact launch/kill/close tests the criterion names — TestBrowserManager_LaunchAndKill / _KillIdempotent / _Close / _CloseIdempotent (plus TestConfigureLauncher_PinsSystemBrowser) via `go test -tags integration -run ... ./pkg/crawl` with VESPASIAN_REQUIRE_CHROME=1 so an unavailable Chrome fails rather than skips, and :64-75 asserts each reported `--- PASS` (not skip). Those tests exist (pkg/crawl/browser_integration_test.go:44,99,146). The script is executed inside the built image by live-tests.yml:685-696 (devcontainer exec).

**Gap:** Only verified through the devcontainer-image CI job, which was never executed locally (no container runtime). Directly asserted-in-CI but not yet observed to pass; deferred verification.

**Recommendation:** Rely on the devcontainer-image job's LookPath step result on this PR as AC2 evidence; ensure it is required for merge.

</details>

<details>
<summary><code>REQ-004</code> · nit · <code>.github/workflows/live-tests.yml:725</code></summary>

**Acceptance criterion:** AC3  
**Status:** PARTIAL

**Evidence:** live-tests.yml:725-774 runs `./test/setup-live-targets.sh && ./test/run-live-tests.sh --no-build --group live` INSIDE the image via devcontainer exec, then requires the rod-backed targets no-download, concat-spa, forms-target to report PASS (not SKIP) by name-matching captured output — deliberately not trusting the exit code, since run-live-tests.sh only fails an all-skipped run. concat-spa and forms-target are in BROWSER_TARGETS (setup-live-targets.sh:134). This is the exact command the criterion names, run in the container.

**Gap:** Only verified through the devcontainer-image CI job, never run locally. Deferred verification; correctness of the PASS-not-SKIP gate is sound but unobserved.

**Recommendation:** Use the devcontainer-image job's AC3 step result on this PR as the acceptance evidence; keep it a required check.

</details>

</details>

<details>

<details>
<summary><code>QUAL-001</code> · nit · <code>.devcontainer/devcontainer.json:67</code></summary>

**Issue:** The onCreateCommand comment claims '--ignore-scripts matches every other npm call site in this repo (both invocations in .github/workflows/live-tests.yml)'. The universal claim ('every other npm call site in this repo') is then enumerated as exactly two, but grep shows a third runtime npm call site: test/setup-live-targets.sh:334 (`npm ci --ignore-scripts --silent`), which runs inside the `test` job and the new devcontainer AC3 step. The substantive point (all sites use --ignore-scripts) stays true, but '(both invocations in live-tests.yml)' wrongly implies those are the only other sites. The same phrasing was copied from setup-live-targets.sh:325-327 without accounting for the additional site, so it drifts as the set grows.

**Recommendation:** Drop the parenthetical count or make it inclusive, e.g. 'matches every other npm call site in this repo (the two run: steps in live-tests.yml and the npm ci in test/setup-live-targets.sh)'.

```diff
  // --ignore-scripts matches every other npm call site in this repo (both
  // invocations in .github/workflows/live-tests.yml), keeping package lifecycle
```

</details>

<details>
<summary><code>QUAL-002</code> · low · <code>.devcontainer/on-create.sh:14</code></summary>

**Issue:** The header comment titled 'Why the nvm dance' describes resolving npm explicitly (sourcing nvm because /bin/sh loads no profile) and concludes 'resolving npm explicitly is the fix, not a precaution.' But the script body performs no nvm resolution at all: it only runs `command -v npm` and, if absent, fails with a message pointing at the Dockerfile's build-time /usr/local/bin pin (lines 27-34). npm resolution moved into the Dockerfile (node symlinked into /usr/local/bin) in a later commit, and this comment was left describing the superseded in-script approach. A maintainer reading it would expect the script to source nvm and be misled about where the fix lives — the exact stale-comment class AGENTS.md's own conventions target.

**Recommendation:** Rewrite the header to state the current mechanism: the Dockerfile pins node/npm/npx into /usr/local/bin at build time (on every PATH), so this onCreateCommand needs no profile or nvm sourcing; the `command -v npm` check exists only to fail loudly with a Dockerfile-pointing diagnostic if that layer is missing.

```diff
-# Why the nvm dance. The base image ships Node through nvm
-# so resolving npm explicitly is the fix, not a precaution.
```

</details>

</details>

</details>

---

PR #246, opened 2026-09-09
ci(security): block devcontainer-changes; keep image jobs on audit (LAB-6222)

santiago-praetorian:
Fixes LAB-6222.

## What

LAB-6015 flipped five live-tests jobs to `egress-policy: block` and left the three LAB-5766 devcontainer jobs on `audit`. This ticket:

- flips **`devcontainer-changes` to `block`** with a source-verified allowlist
- **keeps `devcontainer-image` and `devcontainer-image-arm64` on `audit` (AC6)** — hostname allowlists (exact Hub names, `*.docker.io:443`, `production.cloudflare.docker.com:443`) still left docker pulling image config at `dial tcp 54.185.253.63:443` (runs 34379783876, 34388259962, 34389087177, 34394518703, 34395398451). The daemon hop is outside harden-runner's hostname policy.

## Allowlists (`devcontainer-changes` only)

| Job | Endpoints | `disable-sudo` |
| --- | --- | --- |
| `devcontainer-changes` | `github.com:443`, `results-receiver.actions.githubusercontent.com:443` | yes |

Image jobs have no `allowed-endpoints` while they stay on `audit`.

## Pins

`test/test-runner-args.sh`: `EXPECTED_HR_JOBS` is six jobs (LAB-6015's five plus `devcontainer-changes`). `EXPECTED_AUDIT_JOBS` documents the two image jobs. `EXPECTED_ASSERTIONS=254` (254 passed, 0 failed).


.github/workflows/live-tests.yml
```diff
@@ -810,7 +813,16 @@ jobs:
       - name: Harden runner
         uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920  # v2.20.0
         with:
-          egress-policy: audit
+          egress-policy: block
+          allowed-endpoints: >
```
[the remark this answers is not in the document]
santiago-praetorian: Confirmed (RUNTIME) — nvm install 20 fetches https://nodejs.org/dist; omitted host would fail the image build under block. Fixed in `f7c39bccc9839c88c4af09b6d50bf6f629e1dfca`.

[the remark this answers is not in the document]
santiago-praetorian: Confirmed (RUNTIME) — go test inside the amd64 image hits default GOPROXY on an empty module cache. Fixed in `f7c39bccc9839c88c4af09b6d50bf6f629e1dfca` (proxy.golang.org:443 and sum.golang.org:443 on devcontainer-image only).

.github/workflows/live-tests.yml
```diff
@@ -1060,7 +1069,16 @@ jobs:
...
```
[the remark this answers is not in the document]
santiago-praetorian: Confirmed (RUNTIME) — docker/setup-qemu-action@29109295 defaults to docker.io/tonistiigi/binfmt:latest. Fixed in `f7c39bccc9839c88c4af09b6d50bf6f629e1dfca` (docker.io:443 on devcontainer-image-arm64 only; no invented auth hosts).

santiago-praetorian:
Adjudication of the Gemini / Codex workflow review bodies (non-threaded):

**nodejs.org:443** — Confirmed (RUNTIME) — `nvm install 20` is not setup-node. Fixed in `f7c39bccc9839c88c4af09b6d50bf6f629e1dfca`.

**objects.githubusercontent.com:443** — Refuted — technical premise false. Measured path for `curl -fsSL https://github.com/mikefarah/yq/releases/download/...` is 302 to `release-assets.githubusercontent.com` (same CDN LAB-6015 recorded for setup-go/setup-node), zero hops to objects.githubusercontent.com. Added `release-assets.githubusercontent.com:443` instead in `f7c39bccc9839c88c4af09b6d50bf6f629e1dfca`.

**Codex workflow (release-assets + nodejs + docker.io)** — Confirmed (RUNTIME). Same commit.

Frame check skipped — no new layer or abstraction; three existing CI jobs, allowlist edits only.

santiago-praetorian:
@coderabbitai review — re-review 8f0500cd5046f20663ca2430550508b80fa60935 (AC6: image jobs back on audit; walkthrough above still describes all three jobs as block).

.github/workflows/live-tests.yml
```diff
@@ -778,35 +783,45 @@ jobs:
...
+      # no step if:, harden-runner FIRST, the job shape, the exact `with:` key set
+      # and the exact allowlist — plus which jobs carry the step at all and the
+      # full job set, so none of those can move without the build failing and
+      # naming the job.
       #
-      # disable-sudo is NOT set here, unlike preflight-selftest: `docker build`
-      # on a GitHub runner reaches the daemon through the docker group, but the
-      # base image's own layers and the browser install run as root inside the
-      # build sandbox, and pinning that down is what the build is for.
+      # THIS JOB STAYS ON `audit` (LAB-6222 AC6). Hostname allowlists (exact
```
santiago-praetorian, line 793: @gemini review — re-review 8f0500cd5046f20663ca2430550508b80fa60935 (AC6: image jobs back on audit).

santiago-praetorian, line 793: @claude review — re-review 8f0500cd5046f20663ca2430550508b80fa60935 (AC6: image jobs back on audit).

santiago-praetorian, line 793: @codex review — re-review 8f0500cd5046f20663ca2430550508b80fa60935 (AC6: image jobs back on audit).

.github/workflows/live-tests.yml
```diff
@@ -778,35 +783,45 @@ jobs:
...
-      # bump ALL EIGHT copies in lockstep — a missed copy silently leaves a job
-      # on a stale pin.
+      # DUPLICATED across preflight-selftest / validator-regression / docs-check /
+      # devcontainer-changes / devcontainer-image / devcontainer-image-arm64 /
+      # integration-tests / test. harden-runner is a SHA-pinned security control:
+      # bump ALL EIGHT copies in lockstep — a missed copy silently leaves a job on
+      # a stale pin. test/test-runner-args.sh pins the whole policy for all EIGHT
+      # carrying jobs under `block` — this SHA, 'block', disable-sudo (where set),
+      # no step if:, harden-runner FIRST, the job shape, the exact `with:` key set
+      # and the exact allowlist — plus which jobs carry the step at all and the
```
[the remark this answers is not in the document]
santiago-praetorian: **CONFIRMED** (Codex P2, lockstep note at L815). After AC6 the two image jobs are `audit`; this duplicated note still claims all eight are `block` with full-policy pins. That is a false assurance. Will update all eight copies to six-block / two-audit and describe the actual `EXPECTED_AUDIT_JOBS` assertions.

santiago-praetorian:
Adjudication of Gemini / Codex / Claude review bodies on `8f0500c` (re-review after AC6):

**Stale “all eight under block” / empty `EXPECTED_AUDIT_JOBS` comments** (Gemini L127 + L5333 + L2831; Codex L127 + L644; Claude L2827–2831 + L5273) — **CONFIRMED**. This PR authored those lines; after AC6 they disagree with the tree (`EXPECTED_HR_JOBS` is 6, `EXPECTED_AUDIT_JOBS` has the two image jobs, `hr_pad` has 7 entries). Will fix the comments in-PR.

**Image-job comment blocks still lead with a block-mode allowlist** (Claude ~825 / ~1078) — **CONFIRMED**. Dead “first block-mode CI run” prose on jobs that ship `audit`. Will lead with AC6.

**Unused `DEVCONTAINER_JOB_IF`** (Gemini L2762) — **CONFIRMED**. Defined in this PR, absent on `main`, never referenced. Will use it as the audit-loop `jobif=` pin (same gap Gemini asked to close) rather than delete it.

**Audit-loop missing job-level `if:`** (Gemini L3173) — **CONFIRMED** as a gap in the loop this PR added. Both image jobs carry `if: needs.devcontainer-changes.outputs.should-run == 'true'`. Will fold `jobif=` into the existing `aud_want` string (no new counted outcome).

**PR body still says all three jobs flip to `block`** (Claude) — **CONFIRMED**. Will reconcile.

No merge. Fixes next, then another collect.

.github/workflows/live-tests.yml
```diff
@@ -107,10 +121,10 @@ permissions:
...
 # figure is given here on purpose; writing one is how the previous two got stale.)
 #
-# The census belongs in EXPECTED_ALL_JOBS / EXPECTED_HR_JOBS /
-# EXPECTED_AUDIT_JOBS in test/test-runner-args.sh, which ARE asserted against the
-# workflow on every run and fail the build naming the job. Name the arrays here, not
-# the numbers.
+# The census belongs in EXPECTED_ALL_JOBS / EXPECTED_HR_JOBS in
+# test/test-runner-args.sh, which ARE asserted against the workflow on every run
+# and fail the build naming the job. Name the arrays here, not the numbers.
+# EXPECTED_AUDIT_JOBS lists the two image jobs that stay on audit (LAB-6222 AC6).
```
santiago-praetorian, line 122: @gemini review — re-review ec0371733013cc9d497c1c87e88499cc33b42ec8 (stale AC6 comments + audit jobif pin).

santiago-praetorian, line 122: @claude review — re-review ec0371733013cc9d497c1c87e88499cc33b42ec8 (stale AC6 comments + audit jobif pin).

santiago-praetorian, line 122: @codex review — re-review ec0371733013cc9d497c1c87e88499cc33b42ec8 (stale AC6 comments + audit jobif pin).

santiago-praetorian:
@coderabbitai review — re-review ec0371733013cc9d497c1c87e88499cc33b42ec8 (AC6 comment/jobif follow-up; prior walkthrough still described all three jobs as block).

test/test-runner-args.sh
```diff
@@ -3170,50 +3168,27 @@ else
...
-    # took 25 commits in six months. But `container:` (the action does not support
-    # container jobs), `services:` (service containers start before steps[0]) and
-    # harden-runner not being steps[0] each make the step a no-op while `egress-policy`
-    # still reads `audit` — and what is lost then is the very telemetry AC6's follow-up
-    # needs to derive these three allowlists, while the pin still reports them healthy.
     aud_got=$(yq_query "\"policy=\" + ([.jobs.\"${aud_job}\".steps[] | select((.uses // \"\") | test(\"step-security/harden-runner\")) | (.with.\"egress-policy\" // \"<unset>\")] | join(\",\"))
       + \" first=\" + ((.jobs.\"${aud_job}\".steps[0].uses // \"\") | test(\"step-security/harden-runner\") | tostring)
       + \" container=\" + ((.jobs.\"${aud_job}\" | has(\"container\")) | tostring)
       + \" services=\" + ((.jobs.\"${aud_job}\" | has(\"services\")) | tostring)
+      + \" jobif=\" + ((.jobs.\"${aud_job}\".\"if\" // \"false:\") | tostring)
```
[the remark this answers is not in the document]
santiago-praetorian: **CONFIRMED**. The audit loop this PR added pins `jobif=` but not `needs:`. Dropping `needs: devcontainer-changes` leaves the `if:` string identical and the job skipped forever. Will fold `needs=` into `aud_got`/`aud_want` the same way `hr_policy` does.

.github/workflows/live-tests.yml
```diff
@@ -15,23 +15,37 @@ concurrency:
...
+#
+# LAB-6222 source runs that executed all three remaining jobs:
+#   34280577854 (PR #226, head 9561b27) — jobs 102244066291 / 102244143615 / 102244143724
+#   34235998695 (same PR, earlier green)
+#   33943059457 (push to main)
+# Those logs were not readable from the implementing host (`gh run view <id> --log`
+# and `gh api .../jobs/<id>/logs` redirect to *.blob.core.windows.net, which 403s).
+# The three lists are therefore SOURCE-VERIFIED from committed URLs in this tree;
+# first block-mode CI is allowed to name missing endpoints (❌ Blocked) and those
+# are added to THAT job only, then re-pinned. Three guesses recorded on LAB-4732/LAB-6015 did not survive that check:
```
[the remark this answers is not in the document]
santiago-praetorian: **CONFIRMED**. Header still says “the three lists” and “first block-mode CI” as if both image jobs had allowlists. After AC6 only `devcontainer-changes` has a list. Will distinguish the one block list from the two audit jobs.

.github/workflows/live-tests.yml
```diff
@@ -15,23 +15,35 @@ concurrency:
...
 #
 #   32500986374 (393bf02) 32494976039 (b575e67) 32398237813 (c0bb167)
 #   32395398188 (d2a4e90) 32382787790 (f6c0624) 32297452612 (ff299ba)
 #
 # read back with `gh run view <id> --log`. `endpoint called` is a connect that
 # actually happened; `domain resolved` is only a DNS lookup, and the two differ
-# here. Three guesses recorded on LAB-4732/LAB-6015 did not survive that check:
+# here.
+#
+# LAB-6222: only `devcontainer-changes` got a new allowlist. Source-run logs
```
santiago-praetorian: @gemini review — re-review 0440c3bf93140a6ec22f975834cf83fe55b26cea (needs= pin + three-list prose fix).

santiago-praetorian: @claude review — re-review 0440c3bf93140a6ec22f975834cf83fe55b26cea (needs= pin + three-list prose fix).

santiago-praetorian: @codex review — re-review 0440c3bf93140a6ec22f975834cf83fe55b26cea (needs= pin + three-list prose fix).

santiago-praetorian:
@coderabbitai review — re-review 0440c3bf93140a6ec22f975834cf83fe55b26cea (needs= pin + three-list prose).

santiago-praetorian:
Adjudication of Gemini / Codex / Claude / CodeRabbit on `0440c3b`:

**Gemini / Codex workflow bodies** — no findings. Recorded.

**Codex connector** — no major issues. Recorded.

**Claude `+3` tally at ~5331** — **REFUTED**. That `+3` is the LAB-6015 delta (three audit pins added then). LAB-6222 moved one of those into the `+6` HR line; the same paragraph already says the two image jobs stay in `EXPECTED_AUDIT_JOBS` and net counted outcomes are unchanged. Relabeling it `+2` would misstate the LAB-6015 add. `EXPECTED_ASSERTIONS=254` is the live guard.

**CodeRabbit: PR title/body still say all three jobs flip to `block`** — **CONFIRMED**. `gh pr edit` body from the prior round did not persist. Title and body updated just now to six-block / two-audit / AC6.

No merge.

santiago-praetorian:
## Adjudication complete @ 0440c3b

**Frame skip:** no new layer — CI policy flip of existing harden-runner jobs plus pin/comment repairs. Size-or-behavior skip recorded.

**Coverage @ 0440c3b**

| Reviewer | Status |
| --- | --- |
| gemini-code | present (review 20:38:17Z) |
| claude-code | present (review 20:40:34Z) |
| codex-code | present (workflow 20:38:23Z + connector 20:38:32Z) |
| coderabbitai[bot] | present (issue comments through 20:42:01Z) |

reviewer items collected through: 2026-09-09T20:42:01Z

**Verdicts this round**

| Finding | Verdict |
| --- | --- |
| Gemini / Codex: no issues | recorded |
| Claude `+3` tally ~5331 | REFUTED (LAB-6015 historical delta; net unchanged) |
| CodeRabbit PR title/body still say all three `block` | CONFIRMED; title+body patched via REST (`ci(security): block devcontainer-changes; keep image jobs on audit (LAB-6222)`) |
| Codex P2 needs= / three-list prose on ec03717 | CONFIRMED; fixed in 0440c3b |

CI green on `0440c3b` including both image jobs under `audit`. LAB-6222 AC6 recorded on the ticket. No merge.

logan-bayless (changes requested):

**PR:** https://github.com/praetorian-inc/vespasian/pull/246
**Linear:** https://linear.app/praetorianlabs/issue/LAB-6222/vespasian-flip-the-three-devcontainer-live-tests-jobs-from-egress
**Head SHA:** 0440c3bf93140a6ec22f975834cf83fe55b26cea
**Reviewed at:** 2026-09-10T13:46:12+00:00

|---|---|---|---|---|---|
| Count | 0 | 1 | 8 | 12 | 2 |
| Blockers | 0 | 0 | 7 | 0 | 0 |

|---|---|---|
| capability-reviewer | 12/12 | 1 |
| test-lead | 9/9 | 0 |

**Vs previous run:** first run for this PR

| Origin | Findings | Blockers |
|---|---|---|
| Introduced by this PR | 18 | 7 |
| Accurate on `main`, made stale by this PR | 3 | 0 |
| Work this PR declined (AC1, AC4) | 2 | 0 |

_Three findings from the first pass of this review were pre-existing defects on `main` and have been removed: they were not caused by this PR. They are retained for the record in `.review/cut-pre-existing.json`. Every finding below either originates in this PR's changes or was made wrong by them, and every blocker is in the first row._

---

<details>
<summary><strong>Blockers (7)</strong></summary>

<details>
<summary><code>QUAL-001</code> · medium · QUAL · <code>.github/workflows/live-tests.yml:20</code></summary>

**Issue:** Lines 20-22 now state a census — which jobs are in EXPECTED_AUDIT_JOBS and that devcontainer-changes is on `block` — and lines 23-25, in the same paragraph, say the block 'deliberately does NOT restate which jobs or how many, because every hand-maintained count in this file has gone stale'. The block contradicts its own rule three lines after stating it, and it mixes authoritative membership data into narrative that no assertion reads. Nothing guards it: the only prose pin over this file greps for the literal string 'once telemetry confirms the set' (test/test-runner-args.sh:3163), and every structural pin reads the YAML, not the comments. So the next ticket that flips devcontainer-image to `block`, or that adds a third audit job, leaves L20-22 asserting the wrong membership with the whole suite green — the identical failure mode L133-138 of this same file documents having already occurred three times.

**Recommendation:** Cut the membership claim back to the pointer the block's own rule prescribes: keep 'the harden-runner steps in EXPECTED_HR_JOBS run block, the ones in EXPECTED_AUDIT_JOBS run audit, both arrays live in test/test-runner-args.sh and are the authority' and drop the enumeration of which jobs are in which array. The 54.185.253.63 rationale belongs at the two image jobs (L788-793, L1018-1021 already carry it), not in the shared header.

```
# EXPECTED_AUDIT_JOBS is the two image jobs (LAB-6222 AC6): docker's daemon
# still dials 54.185.253.63:443 after hostname allowlists. `devcontainer-changes`
# is on `block`.
# The array names live in test/test-runner-args.sh and are the authority — this
# block deliberately does NOT restate which jobs or how many, because every
# hand-maintained count in this file has gone stale (see the note below). The
```

</details>

<details>
<summary><code>QUAL-005</code> · medium · QUAL · <code>.github/workflows/live-tests.yml:652</code></summary>

**Issue:** devcontainer-changes flips to `egress-policy: block` with a structural pin only. The pin compares the YAML (mode, SHA, allowlist set, job shape) and nothing observes that the policy actually refuses an unlisted host in this job, so a green run under `block` is indistinguishable from the same job under `audit` — which is exactly the state this PR is trying to leave. Two other block jobs already close that gap with a runtime probe: preflight-selftest (L329) and validator-regression (L406) each run `./test/assert-egress-enforced.sh` as their last step, and test/assert-egress-enforced.sh:74-79 records which jobs the probe can drop into (it needs github.com allowlisted as the reachable control and proxy.golang.org absent from the job's allowlist as the refused target). devcontainer-changes now satisfies both conditions exactly as docs-check and preflight-selftest do, so the probe is a drop-in here. The consequence of leaving it out is not theoretical: this job is `needs:` for both image jobs and runs on every PR, so it is the highest-traffic block job in the file and the only one whose enforcement is asserted purely from text. LAB-6222 AC3 asks for the refusal 'demonstrated, not assumed', and the ticket's design constraints say 'Verify the failure path for these jobs specifically; a block policy that never blocked is indistinguishable from audit'. AC3 is still unchecked on the ticket and nothing in the diff supplies the demonstration.

**Recommendation:** Append the existing probe as the last step of devcontainer-changes — `- name: Assert egress policy enforces (AC3)` / `run: ./test/assert-egress-enforced.sh` — placed last for the reason L320-327 already documents, and extend the step-position pin in test/test-runner-args.sh to cover the third copy. If the step is deliberately omitted, record why at the job and say on the ticket that AC3 is met by the two existing copies rather than leaving it unstated.

```
          egress-policy: block
          disable-sudo: true
          allowed-endpoints: >
            github.com:443
            results-receiver.actions.githubusercontent.com:443
```

</details>

<details>
<summary><code>TEST-001</code> · medium · TEST · <code>.github/workflows/live-tests.yml:176</code></summary>

**Issue:** This PR introduces two new hand-maintained number-words into the duplicated harden-runner comment: 'pins the whole policy for the SIX jobs in EXPECTED_HR_JOBS' and 'pins the TWO image jobs in EXPECTED_AUDIT_JOBS'. Measured: `grep -c SIX .github/workflows/live-tests.yml` returns 8 - the claim is duplicated once per harden-runner job - and grep across test/test-runner-args.sh finds no assertion referencing either word or comparing anything to ${#EXPECTED_HR_JOBS[@]} / ${#EXPECTED_AUDIT_JOBS[@]}. The repo already has the exact guard pattern for this defect one comment over: test/test-runner-args.sh:4082-4092 extracts `bump ALL [A-Z]+ copies` from the workflow, maps the word to an integer, and fails if it disagrees with the counted harden-runner steps, with a comment recording the measured failure it was written for ('Measured on PR #228: deleting one harden-runner step left six comments claiming seven beside six actual steps, with every check green'). The new counts get none of that. This also contradicts the provenance header the same PR rewrote at .github/workflows/live-tests.yml:24-26, which says the block 'deliberately does NOT restate which jobs or how many, because every hand-maintained count in this file has gone stale'. Moving one job between the two arrays now silently leaves eight comment copies lying, and the next reader trying to work out what is pinned reads the comment, not the array.

**Recommendation:** Extend the existing lockstep-word block at test/test-runner-args.sh:4082 with two more comparisons in the same shape: extract `for the ([A-Z]+) jobs in EXPECTED_HR_JOBS` and `the ([A-Z]+) image jobs in EXPECTED_AUDIT_JOBS`, map through the existing hr_numbers associative array, and compare against ${#EXPECTED_HR_JOBS[@]} and ${#EXPECTED_AUDIT_JOBS[@]}. Add a matching hr_pad entry so the workflow-absent arm stays balanced, and bump EXPECTED_ASSERTIONS. Alternatively, drop both number-words from the comment and name the arrays only, which is what the provenance header at line 24 already prescribes and needs no new assertion.

```
      # bump ALL EIGHT copies in lockstep — a missed copy silently leaves a job on
      # a stale pin. test/test-runner-args.sh pins the whole policy for the SIX
      # jobs in EXPECTED_HR_JOBS under `block` (this SHA, 'block', disable-sudo
      # `with:` key set and the exact allowlist) and pins the TWO image jobs in
      # EXPECTED_AUDIT_JOBS as `audit` first-step (LAB-6222 AC6) — plus which
```

</details>

<details>
<summary><code>TEST-002</code> · medium · TEST · <code>.github/workflows/live-tests.yml:186</code></summary>

**Issue:** devcontainer-changes flips to `egress-policy: block` with a two-entry allowlist, and nothing in the repo proves that policy actually blocks anything in THIS job. The only runtime enforcement assertion is test/assert-egress-enforced.sh, invoked at .github/workflows/live-tests.yml:329 (preflight-selftest) and :406 (validator-regression); a repo-wide grep for `assert-egress-enforced` returns no third invocation. Every assertion this PR adds for devcontainer-changes reads the YAML that DESCRIBES the policy, not the policy's effect. The test file states the distinction itself at test/test-runner-args.sh:2997: 'This is the only runtime check that block mode actually ENFORCES; the policy pin above only checks what the YAML says.' Linear AC3 for this ticket requires 'a deliberate request to an unlisted host fails one of these jobs - demonstrated, not assumed', and AC3 is unchecked on the ticket. The gap has teeth here specifically because this job's allowlist is NOT telemetry-derived: the job's own header comment at lines 168-172 records that `gh run view` returned empty for all three source runs (Actions log blobs 403) and that the list is 'SOURCE-VERIFIED from committed URLs' instead. A block policy assembled by reading source and never observed refusing anything is indistinguishable from audit until the first real PR breaks on it, and the header's fallback plan ('First block-mode CI run is allowed to name missing endpoints') puts that discovery in a contributor's PR rather than in this one.

**Recommendation:** Append the AC3 step to devcontainer-changes the way preflight-selftest and validator-regression carry it, and extend the ac3_want pin at test/test-runner-args.sh:2973 to three jobs. test/assert-egress-enforced.sh's SCOPE paragraph already notes that the unlisted host must be absent from the job's own allowlist; devcontainer-changes allowlists only github.com:443 and results-receiver.actions.githubusercontent.com:443, so the script's existing proxy.golang.org probe and github.com control both work unchanged. If the step is deliberately not added (this job's 5-minute timeout, or not wanting a network-dependent assertion in a path filter), record that decision as an assertion rather than prose, and say plainly on the ticket that AC3 is unmet for devcontainer-changes rather than satisfied by the preflight-selftest proof.

```
          egress-policy: block
          disable-sudo: true
          allowed-endpoints: >
            github.com:443
            results-receiver.actions.githubusercontent.com:443
      # Egress allowlist SOURCE-VERIFIED from committed URLs (LAB-6222). Telemetry
      # of `endpoint called` was not readable from this host: `gh run view` for
      # First block-mode CI run is allowed to name missing endpoints (❌ Blocked);
```

</details>

<details>
<summary><code>TEST-003</code> · medium · TEST · <code>docs/agents/ci.md:36</code></summary>

**Issue:** The PR rewrites the ci.md sentence that states the block/audit split, replacing 'FIVE run egress-policy: block ... The three devcontainer jobs' with 'Six carrying jobs run `egress-policy: block` ... The two image jobs'. Neither new number is asserted. The immediately preceding sentence in the same file ('Eight of the nine non-container jobs ... open with a step-security/harden-runner step') IS asserted, in both halves, by test/test-runner-args.sh:4113-4151, and the comment introducing that assertion at :4104-4108 says exactly why: 'prose is what went stale twice: it said "the six non-container jobs" when seven were non-container, was corrected to "eight", and was wrong again the moment a tenth job landed. Derive both numbers and compare, so the third recurrence is a failing assertion rather than a review finding.' This PR just wrote a third and fourth hand-maintained number into the same paragraph of the same file and left them as a review finding rather than an assertion. Moving a job between EXPECTED_HR_JOBS and EXPECTED_AUDIT_JOBS - which is precisely the follow-up work AC6 defers - will leave this sentence wrong with the whole suite green.

**Recommendation:** Extend the existing ci.md prose block at test/test-runner-args.sh:4109-4152 with a third comparison: grep `(One|Two|...|Ten) carrying jobs run` and `The (one|two|...|ten) image jobs` from docs/agents/ci.md, map through the existing nc_numbers array, and compare against ${#EXPECTED_HR_JOBS[@]} and ${#EXPECTED_AUDIT_JOBS[@]}. Mirror the existing skip-credit arms so a yq-less host and an absent ci.md stay balanced, add the matching hr_pad entries, and bump EXPECTED_ASSERTIONS.

```
  Six carrying jobs run `egress-policy: block` with a per-job `allowed-endpoints` list (LAB-6015's five plus `devcontainer-changes`). The two image jobs (`devcontainer-image`, `devcontainer-image-arm64`) stay on `audit` under LAB-6222 AC6
```

</details>

<details>
<summary><code>REQ-001</code> · medium · REQ · <code>.github/workflows/live-tests.yml:633</code></summary>

**Issue:** AC2 requires each allowlist to be 'derived from harden-runner audit telemetry of real runs, and the source runs are named on this ticket'. Both halves fail for the one list this PR adds. The workflow's own comment records that the telemetry could not be read: `gh run view` for source runs 34280577854, 34235998695 and 33943059457 returned empty because the Actions log blobs 403'd, so the two endpoints were instead inferred from the job's committed steps ('SOURCE-VERIFIED from committed URLs'). Those three run IDs appear nowhere in the LAB-6222 description or its 2026-09-09 comment; the only run IDs recorded on the ticket are the five FAILED block-mode image-job runs. Unlike AC1 and AC4, the ticket's own tally of what the AC6 exception leaves unmet does not mention AC2, so nothing tracks it. The derivation may well be correct on inspection - the job runs checkout plus `git fetch --no-tags origin "$BASE_REF"` and nothing else, and github.com:443 plus results-receiver is the same pair docs-check and preflight-selftest run green under block - but 'correct on inspection' is the folklore-allowlist path AC2 was written to forbid, and the ticket's Proposed-work item 1 says so in as many words.

**Recommendation:** Either read the telemetry from a host that can reach the Actions log blobs and record the union of `endpoint called` lines for devcontainer-changes on the ticket, or amend AC2 on LAB-6222 to accept source verification for a job whose egress is one `git fetch`, and record that amendment as an explicit exception the way AC6 was recorded. Name the three source runs on the ticket either way.

```
      # Egress allowlist SOURCE-VERIFIED from committed URLs (LAB-6222). Telemetry
      # of `endpoint called` was not readable from this host: `gh run view` for
      # source runs 34280577854 (job 102244066291), 34235998695, and 33943059457
      # returned empty (Actions log blobs 403). This job is checkout + git fetch
```

</details>

<details>
<summary><code>REQ-002</code> · medium · REQ · <code>.github/workflows/live-tests.yml:652</code></summary>

**Issue:** AC3 requires 'a deliberate request to an unlisted host fails one of these jobs - demonstrated, not assumed - and the demonstration is recorded here'. No devcontainer job carries the probe. `test/assert-egress-enforced.sh` is invoked from exactly two jobs at head, preflight-selftest (live-tests.yml:329) and validator-regression (live-tests.yml:406), and is pinned to those two by name in test/test-runner-args.sh. devcontainer-changes is therefore flipped to block with a structural YAML pin and no runtime evidence, which is the condition the ticket's own Proposed-work item 4 names: 'A block policy that has never blocked anything is indistinguishable from audit.' The nearest real evidence points the other way: the five failed block-mode runs the AC6 exception cites are enforcement working on the two IMAGE jobs, both of which are back on audit. AC3 is not mentioned in the ticket's tally of what the exception leaves unmet, so unlike AC1 and AC4 it is not tracked anywhere.

**Recommendation:** Add an `./test/assert-egress-enforced.sh` step to devcontainer-changes and pin it the way the preflight-selftest and validator-regression steps are pinned. The script drops in unchanged: it probes UNLISTED_URL=https://proxy.golang.org/ and CONTROL_URL=https://github.com/, and devcontainer-changes allowlists github.com:443 while omitting proxy.golang.org, which is exactly the precondition its own SCOPE paragraph states. Then record the demonstration on LAB-6222.

```
          egress-policy: block
          disable-sudo: true
          allowed-endpoints: >
            github.com:443
            results-receiver.actions.githubusercontent.com:443
```

</details>

</details>

<details>
<summary><strong>Non-blockers (16)</strong></summary>

<details>
<summary>Ticket alignment (2)</summary>

<details>
<summary><code>REQ-003</code> · high · <code>.github/workflows/live-tests.yml:809</code></summary>

**Origin:** work this PR declined rather than a change it made; the underlying lines are unchanged from `main`. The image jobs' `egress-policy: audit` lines are unchanged from origin/main. AC1 is unmet because the PR chose not to flip them, recorded under AC6, not because of anything it altered.

**Issue:** AC1 requires all three devcontainer jobs to set `egress-policy: block` with an explicit scoped allowlist. At head only devcontainer-changes does. devcontainer-image (live-tests.yml:809) and devcontainer-image-arm64 (live-tests.yml:1041) still read `egress-policy: audit` with no `allowed-endpoints` key at all, so the two jobs that run `docker buildx`, QEMU and the devcontainer CLI - and that COPY and RUN test/install-chrome.sh as root, reaching dl.google.com and github.com - keep unrestricted outbound access. The deferral is declined and recorded rather than silent: it appears at both jobs, in EXPECTED_AUDIT_JOBS' justification, in docs/agents/ci.md:36, in the PR body, and on the ticket, and is corroborated by real failing runs (34394518703 and 34395398451 are confirmed `failure` in the branch run list). AC6, the ticket's own escape hatch for a job that cannot be flipped, is the one box checked.

**Recommendation:** No change requested in this PR. The remaining work is tracked on LAB-6222, which stays In Progress with AC1 unchecked. Worth resolving there rather than closing the ticket: harden-runner's hostname policy cannot see the docker daemon's own hop to 54.185.253.63:443, so the next step is a mechanism that can (a registry mirror or proxy the daemon is configured against, or an IP-level allowance), not a wider hostname list.

```
          egress-policy: audit
```

</details>

<details>
<summary><code>REQ-004</code> · medium · <code>test/test-runner-args.sh:2733</code></summary>

**Origin:** work this PR declined rather than a change it made; the underlying lines are unchanged from `main`. AC4 is unreachable while AC1 stays open. Not a defect in any line this PR wrote.

**Issue:** AC4 requires a normal live-tests run on main to be green under block across all eight harden-runner jobs with no allowlist gaps. Six of eight run block at head (EXPECTED_HR_JOBS at test/test-runner-args.sh:2733) and two run audit (EXPECTED_AUDIT_JOBS at :2741), so the precondition fails structurally regardless of any run's outcome. Declined and recorded: the ticket states 'AC1 (all three block) and AC4 (all eight block green on main) remain unmet by this exception'. What IS verified is narrower and worth stating: run 34401238792 on head 0440c3bf is green across every live-tests job, with devcontainer-changes passing under block in 15s.

**Recommendation:** No change requested in this PR. AC4 becomes reachable only once AC1 is satisfied; both stay open on LAB-6222.

```
EXPECTED_HR_JOBS=(preflight-selftest validator-regression docs-check devcontainer-changes integration-tests test)
EXPECTED_AUDIT_JOBS=(devcontainer-image devcontainer-image-arm64)
```

</details>

</details>

<details>
<summary>Quality (7)</summary>

<details>
<summary><code>QUAL-002</code> · nit · <code>.github/workflows/live-tests.yml:46</code></summary>

**Issue:** L46 is 88 characters where every other prose line in this comment block is at most 83 (measured with awk over the file's `^#` lines; the only longer hits are the two box-drawing rules at L18 and L139, whose length is a multibyte artefact). It is the single wrap-width outlier in the block and it was produced by appending a sentence to a paragraph rather than reflowing it, which is also why the preceding line ends mid-clause at 'while they stay on'.

**Recommendation:** Reflow the paragraph to the block's ~80-column wrap after moving the 'Three guesses' sentence back to the paragraph it belongs to.

```
# `audit` (AC6). Three guesses recorded on LAB-4732/LAB-6015 did not survive that check:
```

</details>

<details>
<summary><code>QUAL-003</code> · low · <code>.github/workflows/live-tests.yml:46</code></summary>

**Issue:** 'Three guesses recorded on LAB-4732/LAB-6015 did not survive that check' was appended to the end of the new LAB-6222 paragraph. Its antecedent is the LAB-6015 telemetry check described at L35-37 ('read back with `gh run view <id> --log`. `endpoint called` is a connect that actually happened; `domain resolved` is only a DNS lookup, and the two differ here'), but the nearest preceding 'check' is now LAB-6222's source-verification-from-committed-URLs at L42-45. A reader therefore takes objects.githubusercontent.com, nodejs.org and storage.googleapis.com to have been refuted by inspecting committed URLs rather than by reading run telemetry, which inverts the evidence standard this whole block exists to record — and the three entries were in fact refuted against the six named runs, not against the source tree.

**Recommendation:** Move the sentence back onto the paragraph that ends at L37 ('...and the two differ here.'), so it reads immediately after the check it names, and let the LAB-6222 paragraph end at 'while they stay on `audit` (AC6).'

```
# re-pin. The two image jobs have no `allowed-endpoints` while they stay on
# `audit` (AC6). Three guesses recorded on LAB-4732/LAB-6015 did not survive that check:
# objects.githubusercontent.com and nodejs.org are never contacted (setup-go and
```

</details>

<details>
<summary><code>QUAL-004</code> · low · <code>.github/workflows/live-tests.yml:638</code></summary>

**Issue:** 'First block-mode CI run is allowed to name missing endpoints (❌ Blocked); add those to THIS job only, then re-pin' (L638-639, and the same allowance at L43-44 in the shared header) is stale at head and weakening as it stands. Stale: the author records on the PR that CI is green on 0440c3b, which is this head, so the first block-mode run has already happened and did not name a missing endpoint — the sentence describes a pending event as future work. Weakening: left in place it reads as a permanent exception to the rule stated four lines earlier at L26-27, licensing the next maintainer to widen this allowlist from a failed run rather than from telemetry. The same PR also relaxed the header's rule from 'DERIVED FROM TELEMETRY, not guessed' to 'DERIVED FROM TELEMETRY where the logs were readable', so the guard and its exception now both live in prose with nothing pinning either.

**Recommendation:** Replace both copies with what was measured: the two endpoints, the green run id under `block`, and the standing rule that a future ❌ Blocked entry goes on THAT job's list with a note plus a re-pin — which L136-139 already states once for the whole file and does not need restating as a per-job licence.

```
      # against github.com — the same two endpoints as docs-check / preflight-selftest.
      # First block-mode CI run is allowed to name missing endpoints (❌ Blocked);
      # add those to THIS job only, then re-pin.
```

</details>

<details>
<summary><code>QUAL-006</code> · low · <code>.github/workflows/live-tests.yml:793</code></summary>

**Issue:** 'disable-sudo is NOT set: `docker build` needs the docker group' (L793, and again at L1020-1021 on the arm64 leg) gives a reason that refutes its own conclusion. harden-runner's `disable-sudo` removes the runner user's sudo access; reaching the daemon *through the docker group* is precisely the path that does not need sudo, so the clause as written is an argument for setting disable-sudo, not against it. The text this PR replaced made the real argument and kept both halves: 'docker build on a GitHub runner reaches the daemon through the docker group, BUT the base image's own layers and the browser install run as root inside the build sandbox, and pinning that down is what the build is for.' Condensing it dropped the 'but' clause that carried the reasoning. The cost is concrete rather than stylistic: the next maintainer who considers hardening these two jobs reads a false premise ('adding disable-sudo would break docker build') and drops a real hardening opportunity, on the two jobs that already have the weakest policy in the file.

**Recommendation:** State the reason that actually holds, in both copies: disable-sudo is omitted because the build runs the base image's layers and install-chrome.sh as root inside the build sandbox, which is what the job exists to exercise — not because `docker build` needs sudo. Alternatively, if disable-sudo has been measured to break the build, cite that run instead.

```
      # The daemon's Hub hop is outside harden-runner's hostname policy.
      # disable-sudo is NOT set: `docker build` needs the docker group.
```

</details>

<details>
<summary><code>QUAL-007</code> · low · <code>.github/workflows/live-tests.yml:1018</code></summary>

**Issue:** The arm64 comment says the job stays on audit for the 'same reason as `devcontainer-image`: docker/setup-buildx image-config pull keeps dialling 54.185.253.63:443'. devcontainer-image does not use that action. `yq -r '.jobs["devcontainer-image"].steps[] | (.uses // "run")'` returns only step-security/harden-runner and actions/checkout followed by nine `run:` steps; docker/setup-qemu-action and docker/setup-buildx-action appear solely in the arm64 job. The amd64 copy of the comment attributes the same dial to the daemon ('docker pulling image config', L790-792), which is compatible with the runs cited — the setup-buildx attribution is not, so the two comments now name different causes for one shared symptom and one of them names an action the job never runs. In a file whose stated convention is that every comment carries measured provenance, that folklore is what later widens the wrong allowlist.

**Recommendation:** Use the amd64 wording in both copies — the docker daemon's Hub image-config hop, which is outside harden-runner's hostname policy — and drop the setup-buildx attribution unless a run log shows the dial originating there.

```
      # THIS JOB STAYS ON `audit` (LAB-6222 AC6), same reason as
      # `devcontainer-image`: docker/setup-buildx image-config pull keeps
      # dialling 54.185.253.63:443 after hostname allowlists. disable-sudo is
      # NOT set: `docker build` needs the docker group.
```

</details>

<details>
<summary><code>QUAL-008</code> · nit · <code>docs/agents/ci.md:36</code></summary>

**Issue:** The rewrite replaced an explicit enumeration with a ticket-keyed referent: the previous sentence listed the five jobs by name ('`preflight-selftest`, `validator-regression`, `docs-check`, `integration-tests`, `test`'), and the new one says 'LAB-6015's five plus `devcontainer-changes`'. After this change no in-repo artefact states which six jobs run under `block` — the workflow header deliberately refuses to (L23-25), and this doc now defers to a Linear ticket. A reader of docs/agents/ci.md has to open Linear to answer 'which jobs enforce an allowlist', which is the primary question the paragraph exists to answer. The preceding paragraph enumerates its eight jobs inline, so the convention in this file is to name them.

**Recommendation:** List the six job names inline as the previous revision did, and keep the LAB-6015/LAB-6222 references as provenance for the two flips rather than as the identifier for the set.

```
  Six carrying jobs run `egress-policy: block` with a per-job `allowed-endpoints` list (LAB-6015's five plus `devcontainer-changes`).
```

</details>

<details>
<summary><code>QUAL-009</code> · low · <code>docs/agents/ci.md:36</code></summary>

**Issue:** The new sentence opens with a hand-maintained count, 'Six carrying jobs run `egress-policy: block`', that no assertion reads. The suite derives and compares only the two number-words in the preceding sentence: test/test-runner-args.sh:4113 and :4138 both grep for the fixed pattern '(One|...|Ten) of the (one|...|ten) non-container jobs' and compare each half against the workflow, and the comment at :4103-4108 states why ('prose is what went stale twice'). 'Six' matches neither pattern, so flipping either image job to `block` — the explicit next step of this AC6 exception — leaves this line claiming six with the whole suite green, which is the exact recurrence those two pins were written to stop.

**Recommendation:** Either phrase the count so the existing pin catches it, or add a third derived comparison against `[.jobs[] | select(.steps[]?.with."egress-policy" == "block")] | length` for this sentence. If neither is wanted, drop the numeral and name the arrays, the way the workflow header's own rule prescribes.

```
  Six carrying jobs run `egress-policy: block` with a per-job `allowed-endpoints` list (LAB-6015's five plus `devcontainer-changes`). The two image jobs (`devcontainer-image`, `devcontainer-image-arm64`) stay on `audit` under LAB-6222 AC6: docker's daemon still dials 54.185.253.63:443 after hostname allowlists. `test/test-runner-args.sh` pins the six in `EXPECTED_HR_JOBS` and the two in `EXPECTED_AUDIT_JOBS`.
```

</details>

</details>

<details>
<summary>Tests (4)</summary>

<details>
<summary><code>TEST-004</code> · low · <code>test/test-runner-args.sh:2752</code></summary>

**Origin:** accurate on `main`, made stale by this PR — the line itself is untouched. CONTROL: origin/main carries "five jobs by the un-gated-job guards", accurate against a 5-job array. The defective text at 2752 is untouched; the finding's span only reaches a touched region at 2762-2764.

**Issue:** Stale prose was one of the two defects this PR set out to fix, and three instances survive in the block it edited. Line 2758, 'The two gated jobs legitimately carry a job-level `if:`; the three un-gated guard jobs must not', is now wrong: four jobs in EXPECTED_HR_JOBS carry `jobif='false:'` - preflight-selftest, validator-regression, docs-check and the devcontainer-changes case this PR added at :2771. That sentence sits on the line directly above GATED_JOB_IF and two lines above the `DEVCONTAINER_JOB_IF=` line the PR introduced, so it is the comment the reader of the new line reads. Lines 2752-2755, 'continue-on-error: true ... was caught for three of the five jobs ... folding it in here makes the coverage uniform across all five', now describe six block jobs. Line 2764, added by this PR, reads 'Parameterize sudo/withkeys rather than hardcoding the five-job shape' - defensible as a reference to the shape being replaced, but next to two genuinely stale 'five's it reads as a fourth current-state claim. None of these is caught by anything: the AC5 guard at :3167 greps only for the literal 'once telemetry confirms the set', and the number-word assertions at :4082 and :4113 cover the workflow's lockstep comment and ci.md, not this file's own comments.

**Recommendation:** Say 'the four un-gated guard jobs' at :2758 and 'all six' at :2755, or re-word both to name the arrays instead of a count, which is what the workflow's own provenance header prescribes for the same reason. Reword :2764 to make the historical reference explicit ('rather than the single hardcoded shape the five-job version used').

```
# `continue-on-error: true` on the harden-runner step was caught for three of the
# five jobs by the un-gated-job guards further up, but NOT for docs-check or
# coverage uniform across all five rather than incidental.
# The two gated jobs legitimately carry a job-level `if:`; the three un-gated guard jobs
# must not. So `jobif` is pinned per job by VALUE rather than as a blanket absence — the
# jobs). Parameterize sudo/withkeys rather than hardcoding the five-job shape.
```

</details>

<details>
<summary><code>TEST-005</code> · low · <code>test/test-runner-args.sh:3174</code></summary>

**Issue:** The PR deletes roughly fifteen lines of comment that were the only record of WHY the audit loop pins each of its fields, and puts nothing in their place - the loop at head opens with `for aud_job in ...` and has no comment at all. Two measurements went with them. The first justified the job-shape half: 'container: (the action does not support container jobs), services: (service containers start before steps[0]) and harden-runner not being steps[0] each make the step a no-op while egress-policy still reads audit - and what is lost then is the very telemetry AC6's follow-up needs.' The second justified the step-level half by measurement: 'Measured: WITHOUT the step-level half of this string, `if: false` on an audit job's harden-runner step and a dropped `disable-sudo` both survived at 253/0 - the step is skipped, the job records NO telemetry, and the pin still reported "runs harden-runner in audit, explicitly".' The assertions themselves survive, so this is not a coverage regression in the counted sense. It is a durability regression, and it lands in the one file whose stated convention is that these annotations ARE the audit trail: :2843-2848 says 'this file, whose audit trail IS these MEASURED annotations' and :2864-2865 says 'Re-measure both arms when you touch either.' The next maintainer who finds a ten-field string on two audit jobs and wants to trim it now has nothing on disk telling them each field was a measured survivor.

**Recommendation:** Put a short comment back at test/test-runner-args.sh:3174, above the loop: what the loop pins, that the step-level half was measured (if: false and a dropped disable-sudo both survived without it), and that telemetry is the only documented route off audit under AC2 so a no-op'd step makes the AC6 follow-up unsatisfiable while reporting healthy. It does not need to be as long as what was removed, but it needs the two measurements.

```
            for aud_job in "${EXPECTED_AUDIT_JOBS[@]}"; do
    aud_got=$(yq_query "\"policy=\" + ([.jobs.\"${aud_job}\".steps[] | select((.uses // \"\") | test(\"step-security/harden-runner\")) | (.with.\"egress-policy\" // \"<unset>\")] | join(\",\"))
```

</details>

<details>
<summary><code>TEST-006</code> · low · <code>test/test-runner-args.sh:3191</code></summary>

**Issue:** The pass and fail messages are the test names here, and neither moved when the pinned string gained `jobif=` and `needs=`. The pass message still reads 'still runs harden-runner in audit as its FIRST step, with no container: or services:, explicitly' while the assertion now also pins the job's `if:` by exact value, its `needs:`, disable-sudo, the step's `if:`, the step's continue-on-error and the with: key set - so the green line understates its own coverage by six fields. The fail message is the sharper problem: it says "${aud_job}'s harden-runner egress-policy is '${aud_got}', expected 'audit'", which mis-names the defect for eight of the ten fields it can now trip on, and it prints no `want:`. A maintainer who deletes `needs: devcontainer-changes` from devcontainer-image gets told the egress-policy is wrong, is handed a ten-field got string, and is given no expected string to diff it against - the actual difference is one empty field in the middle. The hr_expected counterpart at :2894 does this correctly: it enumerates every mutation class that lands there ('or - the job-shape half - a changed runs-on, a container: or services: key, a job-level continue-on-error or if:, a changed needs:, or an action added to or removed from the job'), and other pins in this file print `want:`/`got:` pairs (:2961-2962, :3143-3144).

**Recommendation:** Name the new fields in the pass message (job-level if: by value, needs:, disable-sudo, the step's if:/continue-on-error and the with: key set), reword the fail message to lead with 'harden-runner policy does not match the pin' rather than 'egress-policy is', enumerate the mutation classes that land there the way :2894 does, and print `want: ${aud_want}` alongside the got string.

```
                                  pass "${aud_job} still runs harden-runner in audit as its FIRST step, with no container: or services:, explicitly (LAB-6222 AC6: docker daemon IP outside hostname allowlists)" ;;
                    *)            fail "${aud_job}'s harden-runner egress-policy is '${aud_got}', expected 'audit'. If this job is being flipped to block, hostname allowlists have already failed (54.185.253.63:443) — record why, or move it to EXPECTED_HR_JOBS with an hr_expected entry." ;;
```

</details>

<details>
<summary><code>TEST-007</code> · low · <code>test/test-runner-args.sh:5398</code></summary>

**Issue:** The file's mutation ledger at test/test-runner-args.sh:5398 opens 'Mutation-proven. Every mutation below was applied, PROVEN applied, then judged by exit code and reverted' and then enumerates every mutation the section's pins catch. This PR adds two fields to the audit loop and adds nothing to that ledger. The only `needs=` entry in the whole file's annotations is at :5294, and it is about a different assertion - 'dropping check-label from a GATED job's needs:' - i.e. hr_policy's needs= for integration-tests and test, not the audit loop's. So the repository records no proof for either new field, and it simultaneously deleted the audit loop's own inline measurement (see the finding above). I verified both fields independently with interleaved controls against the head workflow, so they do work: control MATCH -> `del(.jobs."devcontainer-image".needs)` CAUGHT, `del(..."if")` CAUGHT, `.if = false` CAUGHT, needs widened to two entries CAUGHT -> control MATCH. That verification lives in this review, not in the repo, which is the gap: the `needs:` defect the review threads recorded was exactly a pin that looked complete and was not, and the guard against its recurrence is the ledger entry.

**Recommendation:** Add to the ledger's 'Caught by the newer pins' sentence: `needs: devcontainer-changes` deleted from an image job, the image job's job-level `if:` deleted, and `if: false` substituted for it - all three measured CAUGHT by the audit loop's new jobif=/needs= fields. Note in the same breath which mutations still SURVIVE that loop (runs-on, job-level continue-on-error, an added action), since the ledger's value is that it distinguishes the two.

```
# Mutation-proven. Every mutation below was applied, PROVEN applied, then judged by exit
# code and reverted. Caught by the per-job pin: policy -> audit; allowlist emptied; a
# newer pins: a job added without harden-runner; install-chrome-e2e losing `container:`;
```

</details>

</details>

<details>
<summary>Documentation (3)</summary>

<details>
<summary><code>DOC-001</code> · low · <code>docs/agents/ci.md:13</code></summary>

**Origin:** accurate on `main`, made stale by this PR — the line itself is untouched. CONTROL: origin/main docs/agents/ci.md carries "three of the five policy jobs" and origin/main EXPECTED_HR_JOBS holds exactly 5 jobs, so the claim was accurate before. PR made it six. Line 13 is untouched.

**Issue:** This PR changed the block-mode job set from five to six and rewrote the count in line 36, but line 13 of the same file still reads 'the text-grep neutering guards cover only three of the five policy jobs'. Verified at head: six jobs set egress-policy: block (preflight-selftest, validator-regression, docs-check, devcontainer-changes, integration-tests, test). Line 13 is not in any hunk of this PR's diff for this file; only line 36 changed.

**Recommendation:** Change 'three of the five policy jobs' to six, or drop the figure. The adjacent sentence's number-words ARE asserted (test/test-runner-args.sh:4113), so an unasserted count in the same file is the drift this repo has already had three times.

```
  or if any harden-runner step drifts from its pinned policy — the action SHA, `block`, `disable-sudo`, ... the text-grep neutering guards cover only three of the five policy jobs
```

</details>

<details>
<summary><code>DOC-002</code> · low · <code>test/test-runner-args.sh:5288</code></summary>

**Origin:** accurate on `main`, made stale by this PR — the line itself is untouched. CONTROL: same counts at origin/main:5312 and :5346, accurate against a 5-job array. Lines untouched by this PR.

**Issue:** The PR updates the ledger's '+5 one per job in EXPECTED_HR_JOBS' to '+6' (line ~5296) but leaves two counts in the same comment block stale: line 5288 'the text-grep neutering guards cover only three of the five policy jobs' and line 5322 'continue-on-error: true on the step was caught for three of the five jobs'. Neither line falls in a diff hunk (hunks in this region are at 5276 and 5334), so both were rendered wrong by this PR rather than being pre-existing. This is the failure mode this file's own comments document repeatedly ('DO NOT WRITE A JOB COUNT HERE. This sentence has now been wrong three times').

**Recommendation:** Update both counts to six. They sit in the same comment block whose '+5' the PR already corrected to '+6', so this is finishing an edit rather than starting one.

```
#       text-grep neutering guards cover only three of the five policy jobs), and any FOURTH
#       on the step was caught for three of the five jobs by the un-gated-job guards
```

</details>

<details>
<summary><code>DOC-003</code> · low · <code>test/test-runner-args.sh:5334</code></summary>

**Issue:** The ledger header at :5273 reads 'This branch adds SEVENTEEN counted outcomes on top of main's 236'. Measured with a base/head control: on origin/main the '+N' entries in that block sum to exactly 17; at head they sum to 18. The PR raised the first entry from '+5' to '+6' (correct, EXPECTED_HR_JOBS went 5 to 6) without lowering the '+3' audit entry (EXPECTED_AUDIT_JOBS went 3 to 2), and rewrote that entry to read 'LAB-6015: one per job in the then-EXPECTED_AUDIT_JOBS' while the loop now emits two. The net actual change is zero, which EXPECTED_ASSERTIONS=254 unchanged at base and head confirms, so the totals are right and only the itemization is off by one. This was raised in review and REFUTED by the author on the ground that the '+3' is a historical LAB-6015 delta; the refutation does not survive the arithmetic, because the header asserts a sum the list no longer reaches. The sibling balance comment at :2827 WAS updated correctly to 'six ... two', which isolates the '+3' entry as the outlier rather than a deliberate convention.

**Recommendation:** Either change the '+3' entry to '+2' and keep the LAB-6015 history in its prose, or raise the header to EIGHTEEN and say which outcome the extra one is. The base revision summing to exactly 17 is the evidence that the list is meant to reconcile against the header.

```
# This branch adds SEVENTEEN counted outcomes on top of main's 236 (main's own base moved
#   +3  LAB-6015: one per job in the then-EXPECTED_AUDIT_JOBS — the three
```

</details>

</details>

</details>

---

_All 23 findings are listed above; none were dismissed as already addressed. Three blockers describe one defect from three lanes and collapse to a single action — `QUAL-005`, `TEST-002` and `REQ-002` all say `devcontainer-changes` was flipped to `block` with no runtime proof that it blocks, and all three are closed by adding an `./test/assert-egress-enforced.sh` step to that job. Likewise `TEST-003` and `QUAL-009` are the same unasserted count in `docs/agents/ci.md:36`. The remaining blockers — `QUAL-001`, `TEST-001`, `REQ-001` — are independent._

_Two of the four Phase 4 reviewers did not run: neither security agent's trigger fires on a changeset of `.yml`, `.md` and `.sh` paths, so this CI egress-hardening PR received no dedicated security-agent pass. Its security content was covered by the quality and test reviewers and by the orchestrator's own passes, but that is a coverage gap in the review tool for this PR shape, not a clean security result._

santiago-praetorian:
**REQ-001 (AC2 telemetry)** — Confirmed: source-run logs are unreadable. Re-measured 2026-09-11: `gh run view 34280577854 --log` is 0 bytes.

Amended AC2 on LAB-6222 to accept SOURCE-VERIFIED lists for a job whose only egress is a single `git fetch` when audit logs 403. Exception recorded the same way as AC6. Source runs named on the ticket: 34280577854, 34235998695, 33943059457.

The `devcontainer-changes` list stays `github.com:443` + `results-receiver.actions.githubusercontent.com:443` (checkout + `git fetch`; same pair as `docs-check` / `preflight-selftest`).

https://linear.app/praetorianlabs/issue/LAB-6222

santiago-praetorian:
**REQ-001 (AC3 demonstration)** — Confirmed. Live Tests run 34621018792 on `d82f84c`: `devcontainer-changes` ran `./test/assert-egress-enforced.sh` last; unlisted host unreachable (curl exit 7); allowlisted github.com reachable. Same two verdicts on `preflight-selftest` and `validator-regression`. Recorded and AC3 checked on LAB-6222.

**REQ-002 (AC1) / REQ-003 (AC4)** — Confirmed as AC6, not a defect in this PR. Image jobs stay on `audit` (daemon hop 54.185.253.63:443). No code change. AC1 and AC4 remain open on LAB-6222.

https://linear.app/praetorianlabs/issue/LAB-6222
https://github.com/praetorian-inc/vespasian/actions/runs/34621018792

logan-bayless (approved):
Re-reviewed at `d82f84c`. All 23 findings from the 2026-09-10 review verified fixed or tracked; no blockers remain. Approving.

